Issue #5, opened 2026-08-25
Accept a session name in every lifecycle command


pmoust:
### Problem

`brig ls` prints a name that no other command accepts. The list prints the
sandbox name, for example `brig-claude-review`. The lifecycle verbs parse a
profile name instead, so the printed name fails with
`unknown profile "brig-claude-review"`.

To act on a session today you repeat the profile and the flag that created
the session:

```
brig run   claude --name review
brig exec  claude --name review -- git status
brig stop  claude --name review
brig rm    claude --name review
```

A user who copies what the list prints gets an error. A name that is shown
must work as an argument.

### Proposed solution

Make the session the operand of the lifecycle verbs.

1. Print a `SESSION` column in `brig ls`, and make that value the identifier.
2. Accept the session name as the first argument of `exec`, `shell`, `stop`,
   `rm` and `env`.
3. Keep a bare profile name working. It means the unnamed session of that
   profile, which is what it means today.
4. Accept a unique prefix of a session name. If the prefix matches more than
   one session, exit with the list of matches.
5. Keep `--name` on `run` and `create`. That is where a session is named.

```
brig run  claude --name review
brig exec review -- git status
brig stop review
brig rm   review
```

### Alternatives considered

Keep the current grammar and print the profile and the session name as two
columns. It is cheaper, but a copied identifier still does not work as an
argument, which is the fault this ticket is about.

### Additional context

`deprecated()` already carries the old spellings of `agents` and `template`.
The same mechanism can carry `--name` on the
lifecycle verbs for one release.

asapranidis:
**Rescope proposed, from the #47 planning pass.** #47 is being split into sub-issues on a
one-issue-one-PR rule, and this becomes the *lifecycle verbs* half of the grammar switch. The body
is left as filed; this comment is the delta.

Suggested title: **Switch the lifecycle verbs to refs**

### Conflicts with #47 as filed

**1. The operand is `<agent>@<label>`, not a flat session name.** This issue proposes
`brig exec review`. #47 rejects a flat name explicitly: it forbids `claude@refactor` and
`codex@refactor` coexisting, and it needs a new authoritative record to resolve. The ref is
derived, never stored.

**2. §4, unique-prefix matching, is dropped.** Beyond #47 not proposing it, it is ambiguous under
refs: `claude` is simultaneously a *complete* ref — that agent's default session — and a prefix of
`claude@refactor`. There is no resolution rule that does not surprise someone.

[...]

PR #67, opened 2026-08-25
feat(verify): verify the digest that boots, not the tag

### LINKED ISSUE #14: Verify the image that boots, not the tag it came from
### Problem

Under the default `missing` pull policy, cosign verifies the tag in the
registry. The sandbox boots the copy in the local store, and the two are not
required to be the same object. The security doc records this under "Two
limitations", with `BRIG_PULL=always` as the workaround.

Provenance is the strongest supply-chain claim brig makes: not "was this
signed" but "was this built by that workflow, in that repo". The claim is
about the bytes that run. Today the check is about a name in a registry.

### Proposed solution

1. Resolve the reference to a digest before the check.
2. Verify that digest.
3. Boot that digest, by passing the digest to the runtime rather than the tag.
4. Report the digest that booted, in the envelope from #10 and in `--json`.
5. When the local store holds a different digest than the one verified, treat
   it as the "signature fails" row of the table in the security doc: stop for
   an image under `ghcr.io/brig-sh/`, warn for anyone else's.

Delete the first of the two limitations when this lands.

### Alternatives considered

- Change the default pull policy to `always`. It closes the gap and it makes
  every run wait on the registry, including runs on a plane.
- Verify the local store instead of the registry. It is the right object, and
  it leaves the first pull of an image unverified, which is the run that
  matters most.

None as such. It narrows one: the guest that boots is the guest that was
checked.

pmoust:
## Summary

Provenance is the strongest supply-chain claim brig makes: not that an image was
signed, but that it was built by that workflow in that repo. The claim is about
the bytes that run. The check was about a name in a registry, and under the
default `missing` pull policy those need not be the same object, so brig could
print "signature verified" about an image it had not checked.

brig now resolves the reference to a digest, verifies that digest, and boots it,
on both platforms. Whether a runtime can boot a digest from its store is a fact
about that runtime, so the seam asks it rather than assuming.

## Related issues

Closes #14
Refs brig-sh/hull#5, brig-sh/hull#14

## Changes

- `verify.Policy.Verify(ref, localDigest)` resolves the reference through
  `cosign triangulate`, verifies that digest, and compares it against what the
  store holds under the tag. A local copy that differs stops the boot as a failed
  signature would, and a yes boots the verified digest rather than the copy on
  disk.
- An image brig did not publish makes no cosign call: `NotOurs` before anything
  is resolved, and it boots by tag as it always has. Resolving it anyway cost a
  registry round trip per boot and a stall off the network, for a pin brig
  cannot vouch for.
- A registry that cannot be reached stops a first-party boot and asks, and
  refuses under `require`. That is what the same outage did before this PR, when
  it failed inside `cosign verify`; an earlier revision of this PR booted the
  cached tag unchecked in that case, which review caught.
- Every cosign call is bounded by a 30 second timeout, so an outage fails in
  seconds rather than hanging the boot.
- `PinsDigest()` and `LocalDigest()` on the runtime seam. containerd always
  pins. hull asks its binary: `hull --version` at 0.1.0-rc23 or newer pins,
  because that is where hull's store lookup started answering a digest
  reference (brig-sh/hull#5, shipped in rc23). Older prints one line saying the
  tag is verified and booted instead, and why. A version brig cannot read is a
  build from source and pins, since a wrong guess that way costs a re-pull and
  never a weaker check.
- The success line names the digest, and is never printed on a boot that did not
  pin one.
- `docs/security.md` and the README state which runtime gives which guarantee.
- Smoke now drives the pinned path end to end against the stub hull: the digest
  reaches `hull run`, and an rc21 stub boots the tag and is told why.

## What stays narrower on macOS, and why

**The first pinned boot after upgrading hull pulls once.** An image pulled under
an older hull has no index digest on record, and a multi-arch tag resolves to its
index digest, so the store misses and pulls. From then on it answers. Under
`BRIG_PULL=never` that first boot fails until the image is pulled again. Verified
against the store on this machine: the one image in it, pulled under rc21, is
recorded by manifest digest `14a04b7a…` while cosign resolves its tag to index
digest `ccbb91d6…`.

**The mismatch report is Linux-only for now.** `hull images` prints a truncated
manifest digest and no index digest, so there is nothing brig can compare
against the digest cosign resolved. Answering the manifest digest would make
every multi-arch image look like a mismatch and stop a first-party boot with a
prompt for nothing, so `LocalDigest` answers "cannot say" on hull and the boot
is pinned without that report. Tracked as brig-sh/hull#14.

**Users cannot get rc23 yet.** The tag pins hull's cask to rc21 and the rc23
release is not cut yet. Until both move, a Homebrew
user is on the tag path with the one-line notice. Nothing here breaks on rc21.

## Checklist

No run against a real runtime. Everything was exercised against the stub runtime
and stub cosign. This changes what brig passes to the runtime on the boot path,
which is exactly the class the checklist warns can pass every test here and still
be wrong, so it wants a real `brig run` on both platforms before merge: a Linux
boot to confirm the digest resolves against the containerd store, and a macOS
boot on hull rc23 once a release binary exists, to confirm the pinned boot finds
its bytes after the one re-pull.

This affects the first promise by way of what the guest is: the image is the code
that runs inside the boundary, and verifying a different object than the one that
boots is a false assurance about it.

The assertions that matter are negative: a local digest that differs from the
verified one stops the boot for a first-party image, warns for a third-party one,
and the digest-named success line does not appear on a path that did not pin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,10 +346,18 @@ third-party images included. `BRIG_VERIFY=off` skips the check.
 [docs/security.md](docs/security.md) explains the reasoning, and what brig does
 not protect you from.
 
-Two limitations worth knowing: under the default `missing` pull policy cosign
-verifies the tag in the registry, not necessarily the copy already in your
-local store; and `claude-desktop` and `ubuntu` still point at
-`ghcr.io/nofireai/` images, so they warn on every boot until those move.
+For an image under `ghcr.io/brig-sh/`, brig resolves the tag to a digest,
+verifies that digest, and boots it, so the image it checked is the image that
+runs and the line saying so names the digest. If the registry cannot be reached
+the boot stops and asks, as it does for a bad signature. Images brig did not
+publish boot by tag with no registry round trip, as before. Pinning needs a
+runtime whose store answers a digest: containerd on Linux, and hull from
+0.1.0-rc23 on macOS. An older hull cannot, so brig verifies and boots the
+tag there and says so; under the default `missing` pull policy that check is of
+the tag in the registry, not necessarily the copy already in your store, and
+`BRIG_PULL=always` is the workaround until you upgrade. `claude-desktop` and
+`ubuntu` still point at `ghcr.io/nofireai/` images, so they warn on every boot
+until those move.
 
 ### Image tags
 
--- a/docs/security.md
+++ b/docs/security.md
@@ -399,11 +399,46 @@ setting falls back to `warn` rather than silently disabling it.
 Point `BRIG_VERIFY_REGISTRY`, `BRIG_VERIFY_IDENTITY` and `BRIG_VERIFY_ISSUER`
 at your own registry and workflow if you publish signed images yourself.
 
-### Two limitations
+### The digest, not the tag
 
-Under the default `missing` pull policy, cosign verifies the tag in the
-registry, not necessarily the copy already in your local store. Use
-`BRIG_PULL=always` when that distinction matters to you.
+A tag is a name, and a name can resolve to different bytes in the registry and
+in your local store. The provenance claim is about the bytes that run, so brig
+resolves the reference to the digest the registry serves before the check,
+verifies that digest, and boots it. The object cosign checked is the object that
+runs, and the success line names the digest rather than the tag it came from. A
+local store holding a different digest under the tag is treated as the
+signature-failure row above: it stops, and a yes boots the verified digest
+rather than the copy on disk. A registry that cannot be reached stops in the
+same way, because one of our images could not be checked, and a network
+failure must not be what turns the default mode into "unchecked";
+`BRIG_VERIFY=require` refuses outright. Every cosign call is bounded, so an
+outage fails in seconds rather than hanging the boot.
+
+All of this is for images under `ghcr.io/brig-sh/`. An image brig did not
+publish carries no signature of ours to check, so brig makes no cosign call for
+it at all: it boots by tag, as it always has, with the one line saying whose
+image it is. Resolving a digest for it would cost a registry round trip on every
+boot and buy a pin brig cannot vouch for.
+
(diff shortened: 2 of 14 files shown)


pmoust:
The macOS exception now has a tracked reason to be removed: brig-sh/hull#5.

hull's store lookup compares the reference string against the stored ref and the stored digest, and a `repo@sha256:...` reference matches neither, so it misses even when those exact bytes are unpacked on disk. Once that resolves against the store, the exception here can go and macOS gets the same guarantee as Linux.

pmoust:
hull 0.1.0-rc23 will ship the store fix from brig-sh/hull#5, so the macOS exception is gone: the adapter now asks `hull --version` and pins from rc23, and an older hull boots the tag and says why. Two things stay narrower on macOS and are in the body: the first pinned boot after upgrading pulls once, and the local-versus-registry mismatch report waits on brig-sh/hull#14.

pmoust:
Revised for the two review findings, in one commit.

**A registry outage no longer boots a first-party tag unchecked.** `Unresolved` now stops and asks under the default and refuses under `require`, which is what the same outage did before this PR when it failed inside `cosign verify`. Pinned by `TestVerifyDigestUnreachableRegistryRefusesWithoutATerminal`, which fails on the previous revision.

**Third-party images make no cosign call.** `NotOurs` is decided before cosign is looked up, so a BYO or local-only image boots by tag with no registry round trip, as it always did. Pinned by `TestVerifyThirdPartyMakesNoCosignCall` and `TestVerifyDigestThirdPartyBootsTheTagWithoutCosign`, which count invocations. The cost is that third-party images are no longer pinned or mismatch-checked, which is stated in the docs.

**Every cosign call is bounded** by a 30 second timeout via `exec.CommandContext`, pinned by `TestRunGivesUpOnAHangingCosign`. A hang reads as the same outage.

PR #69, opened 2026-08-25
feat(policy): the policy document, and the commands to manage it

marrgko:
## Summary

Adds the policy document, a named YAML or JSON document that declares an egress allowlist (a default of `allow` or `deny`, plus `host`/`cidr` exceptions on either side), and the `brig policy` commands that manage it.

The document lives one per file under `$XDG_CONFIG_HOME/brig/policies` (or `BRIG_POLICY_DIR`). The `name:` inside the file is what it's known by, not the filename, and a directory can hold any number of policies independently.

`brig policy create <name>` writes a starter document and opens it in your editor. `edit` re-opens an existing one, validating a save in a scratch copy before it replaces the real file, so a broken save never reaches it. `show [--json]` prints the parsed document. `rm` deletes one. `ls` (or `brig policies`) lists every policy that parses, by name and description. Parsing is strict throughout: a field the format doesn't recognise is refused rather than silently dropped.

## Changes

- `internal/policy`: the `Policy`/`Egress`/`Rule` types, strict YAML/JSON parsing, and validation (`apiVersion`, `egress.default`, exactly one of `host`/`cidr` per rule, name character rule via the exported `CheckName`)
- `internal/policy`: `Dir()` and `LoadAll()`, directory resolution, name-in-file-wins, one bad file doesn't block the rest, duplicate-name reporting
- `cmd/brig`: `brig policy ls/create/edit/show/rm` and `brig policies`
- `README.md`: lists the new commands in the command table

## Checklist

- [ ] ~~I have run `go test ./... -race`, if the change touches concurrency, subprocesses or the daemon~~

## LLM usage

Authored with assistance from Claude Code; I have reviewed every line and am accountable for the change.


cmd/brig/policy.go
```diff
@@ -0,0 +1,302 @@
...
+		return fmt.Errorf("%s no longer parses, and has been left as you saved it:\n  %w",
+			path, err)
+	}
+	fmt.Printf("%s created\n", path)
+	return nil
+}
+
+// policyStarter is a minimal, valid policy document to start editing from.
+func policyStarter(name string) string {
+	return fmt.Sprintf(`apiVersion: %s
```
pmoust, line 270: The name is put into the template without quotes, so YAML reads some names as something else.

CheckName allows no, yes, on, off, true, false, and names made only of digits. YAML 1.1 turns those into a boolean or a number.

Running brig policy create no prints that no.yaml was created and exits 0, but the policy is named false. brig policies lists it as false, and brig policy show no and brig policy rm no both answer unknown policy no. The user is left with a policy they cannot reach by the name they typed. brig policy create 0x10 gives a policy named 16.

Two ways to close it: put quotes around the name in the template, or have CheckName refuse names that YAML reads as something else. Hand written files hit the same coercion, but the template is the part this change owns.

cmd/brig/policy.go
```diff
@@ -0,0 +1,302 @@
...
+
+	dir := policy.Dir()
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return err
+	}
+	path := filepath.Join(dir, name+".yaml")
+	// A starter document is generated bytes with nothing worth keeping, but
+	// the file it would replace might not be, so create refuses to
+	// overwrite one without --force.
+	if !force {
```
pmoust: This only asks whether a file of that name exists. The policy name lives inside the file, and the name field wins over the file name, so this can create a second policy with a name that is already taken.

With a custom.yaml that declares name no-net, brig policy create no-net still succeeds. After that every policy command prints a duplicate name warning saying custom.yaml and no-net.yaml both declare no-net, and the user cannot clear it without finding the other file by hand.

Ask LoadAll whether the name is taken, rather than asking the file system whether the path is taken.

internal/policy/policy.go
```diff
@@ -0,0 +1,117 @@
...
+			if err := r.validate(); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// validate checks that a Rule names exactly one host, in exactly one form.
+func (r Rule) validate() error {
```
pmoust, line 89: This checks that a rule carries either a host or a cidr, but never checks that the cidr is a real cidr, or that the host pattern is well formed.

A rule with cidr 10.0.0/8, a typo with an octet missing, passes validation today, and brig policy show prints it back as though it were fine. The mistake only shows up when the rule is enforced, and a deny rule that cannot be parsed fails in the permissive direction, which is the wrong way round for a deny rule.

This change says the format refuses bad input instead of quietly accepting it, so it is worth closing. net.ParseCIDR here is one line.

cmd/brig/policy.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+	}
+	if err := os.WriteFile(path, []byte(policyStarter(name)), 0o644); err != nil {
+		return err
+	}
+
+	editor := editorCommand()
+	cmd := exec.Command(editor[0], append(editor[1:], path)...)
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+	if err := cmd.Run(); err != nil {
```
pmoust, line 248: When the editor exits with an error this returns the error, but the starter file written just above is left on disk. So create fails and half succeeds at the same time.

The command reports the editor error, brig policies then lists the policy anyway, and running brig policy create again refuses because the file is already there.

editPolicy removes its scratch copy when the editor fails, so this reads as an oversight rather than a decision. Remove the starter when the editor fails, unless force replaced a file that was already there.

cmd/brig/policy.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	names := make([]string, 0, len(entries))
+	for name := range entries {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	for _, name := range names {
+		fmt.Printf("%-15s %s\n", name, entries[name].Policy.Desc)
+	}
+	if len(names) == 0 {
```
pmoust, line 75: LoadAll returns an error for a directory it cannot read. This prints that error to stderr, then still says there are no policies, and exits 0.

With the policy directory set to mode 000, stdout says no policies yet while stderr says permission denied, and the exit status is 0. Anything reading stdout concludes the user has no policies, which is a different thing from not being able to look.

Print the empty message only when there was no error, and return the error otherwise.

cmd/brig/policy.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// lookupPolicy finds one policy by name, or says which command lists what
+// exists.
+func lookupPolicy(name string) (policy.Entry, error) {
+	entries, err := policy.LoadAll(policy.Dir())
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "brig: "+err.Error())
+	}
+	e, ok := entries[name]
```
pmoust, line 89: Same as the listing above. When the directory cannot be read the user is told unknown policy, which points them at brig policies, which will also show nothing.

This affects show, edit and rm. Return the LoadAll error instead of falling through to the not found message when the directory itself could not be read.

cmd/brig/policy.go
```diff
@@ -0,0 +1,302 @@
...
+		// The scratch copy is left in place, named in the error, so the
+		// edit itself is not lost -- only not yet valid.
+		return fmt.Errorf("not saved, %s is unchanged: %w\nyour edit is still at %s",
+			entry.Path, err, scratchPath)
+	}
+	edited, err := os.ReadFile(scratchPath)
+	if err != nil {
+		return err
+	}
+	if err := os.WriteFile(entry.Path, edited, 0o644); err != nil {
```
pmoust: The scratch copy exists so the real file is left alone until the new content is known to be good, but this last step writes over the real file in place.

If the disk fills up, or the process is killed between the truncate and the write, the real policy file is left empty or half written, which is the thing the scratch copy was there to prevent.

Write to a temp file in the same directory and rename it over the old one.

README.md
```diff
@@ -115,6 +115,8 @@ separate Linux re-implementation of it.
 | `brig env <agent>` | what would be forwarded, **by name only**, and whether the guest will be authenticated |
 | `brig profiles` | the profiles, their images, and what each one refuses to forward |
 | `brig profile ls\|export\|import\|edit\|rm` | manage profiles (`brig export` is the short form of `brig profile export`). `export --json` for JSON instead of YAML |
+| `brig policies` | every policy that parses, by name and description |
+| `brig policy ls\|create\|edit\|show\|rm` | manage policies. `show --json` for JSON instead of YAML |
```
pmoust, line 128: These two rows name the commands, but nothing here or anywhere else in the change shows what a policy document looks like.

Someone reading the docs cannot learn what goes in the file without running brig policy create and reading the starter it writes. Please add an example of a complete policy: apiVersion, a name, a description, one host rule and one cidr rule, and a sentence saying what happens to traffic that no rule matches.

pmoust (commented):
very point below was reproduced by running it against a build of this branch. go vet and go test ./... both pass here, so none of these are caught by the tests as they stand.

Seven things, one of them worth fixing before this merges and two more that fail in the permissive direction. The one that matters most is the starter template. It puts the policy name into the file without quotes, and YAML reads several perfectly legal names as something else, so the user gets a policy they cannot address by the name they typed.

The other two worth closing are that a bad cidr is accepted without complaint, and that creating a policy only checks the file name rather than the policy name, so it can make a duplicate.

One more thing that is not a bug. The documentation does not show what a policy actually looks like. The README gains two table rows naming the commands, and there is no page and no worked example anywhere in the change. Someone who reads this has no way to learn what goes in the file without running create and reading the starter. Please add an example of a complete policy document, with a host rule and a cidr rule, and a sentence on what happens to traffic that no rule matches.

cmd/brig/policy.go
marrgko: @/tmp/claude-1000/-home-mgouth-brig/fa80e6a3-a66b-4f1d-9fc8-5efb7a1ff4c8/scratchpad/replies/c7.txt

cmd/brig/policy.go
marrgko: Fixed in `7e3dc13`. `CheckName` now round-trips the candidate through the same YAML decoder `Parse` uses, and refuses one that doesn't come back as itself. Covers the create-time template, and since `Validate` calls `CheckName` too, a hand-written file with the same problem. Tests cover the exact repro plus the wider class (`y`/`n`, hex/octal/binary, digit separators).

cmd/brig/policy.go
marrgko: Fixed in `1e31553`. `create` now checks the name against `LoadAll`'s result instead of the target path, and refuses regardless of `--force`, since forcing would just leave two files declaring the same name.

internal/policy/policy.go
marrgko: Fixed in `fd951a2`. `cidr:` is now checked with `net.ParseCIDR`. `host:` gets a narrower fix: whitespace and control characters are refused, but wildcard shape isn't judged, since this package doesn't know how a rule gets enforced and there's no pinned glob grammar to validate against yet. Ready to discuss if you'd rather see a stricter host check.

cmd/brig/policy.go
marrgko: Fixed in `8d02df4`. `create` removes the starter it wrote if the editor exits with an error, except when `--force` replaced a file that already existed.

cmd/brig/policy.go
marrgko: Fixed in `738a796` (shared with the show/edit/rm case below). `listPolicies` now returns `LoadAll`'s error when the directory itself couldn't be read (`entries == nil`), rather than falling through to the empty message.

cmd/brig/policy.go
marrgko: Same commit as the listing case above (`738a796`). `lookupPolicy` has the identical fix, covering `show`, `edit` and `rm`.

cmd/brig/policy.go
marrgko: Fixed in `90da4b4`: the write-back now goes through a temp file in the same directory and a rename. Follow-up in `2fdfaeb`, since the first version hardcoded the new file to `0644`, which silently reset a file someone had `chmod`ed down. It now carries over the real file's existing mode.

README.md
marrgko: Added `docs/policies.md` in `472a681`, linked from the README's Documentation index, with a complete example, a field table, the five verbs, a worked example and an errors table. Every console example in it is real output, captured by actually running the built binary.

ananos:
Thanks @marrgko! I went through the round: every one of @pmoust's findings checks out as fixed, and I verified the branch on our mac testbed -- suite green, and the create/edit failure paths behave as promised (no starter debris, the real file untouched on an aborted edit).

One ask before this lands: let's squash the `fix(policy)` commits into the semantic commits they amend, so the history does not carry a bug and its fix as separate steps of the same PR. The fixup commits were exactly right for the review round -- reviewers could see what changed between rounds -- but merged history should read as if the code was written correctly the first time. For instance, the cidr/name validation fixes fold naturally into the document-format commit, and the LoadAll error handling into the CLI commands commit. Group them as you see fit; the docs commits can stay their own.

Once that's done (and @pmoust acks his round), this is good to go from my side.

ananos (approved):
thanks @marrgko! LGTM!

Issue #7, opened 2026-08-25
Add `--json` to ls, env, profiles and secret ls


pmoust:
### Problem

Every list command prints aligned text. `brig ls` prints three padded columns,
and `brig env` prints prose lines. A script that reads either one parses column
positions, and any change to the text breaks it.

`brig export --json` already exists, so the flag name
and the behaviour are set.

### Proposed solution

Add `--json` to `ls`, `env`, `profiles`, `profile ls` and `secret ls`. Print
one object per command, with a version field:

```json
{
  "apiVersion": "brig.sh/v1alpha1",
  "kind": "SandboxList",
  "items": [
    {
      "session": "review",
      "sandbox": "brig-claude-review",
      "profile": "claude-code",
      "state": "running",
      "workspace": "/Users/panos/src/brig",
      "runtime": "hull",
      "image": "ghcr.io/brig-sh/claude-code@sha256:..."
    }
  ]
}
```

`env` reports the same facts that `Status()` prints today: workspace, sandbox,
runtime, image, pull policy, credentials by name and source, git over HTTPS,
and the guest login state.

Cover each shape with a golden test, and write down the compatibility rule:
fields are added, not renamed or removed, inside one `apiVersion`.

### Effect on the sandbox boundary

The JSON reports credential names and never credential values, which is the
rule `brig env` follows now. A secret value in any field, an error message
included, is a defect.

asapranidis:
**Rescope note, from the #47 planning pass.** Body left as filed.

The commands rename under #47, so the title becomes **Add `--json` to `ls`, `info`, `agent ls` and
`secret ls`** — `env` is `info`, and `profiles`/`profile ls` are `agent ls`.

The example object needs its field names settled *with* the grammar rather than after it:

```json
{
  "apiVersion": "brig.sh/v1alpha1",
  "kind": "SandboxList",
  "items": [
    {
      "ref": "claude@review",
      "sandbox": "brig-claude-review",
      "agent": "claude-code",
      "state": "running",
      "home": "/Users/panos/brig/claude-review",
      "project": "/Users/panos/src/brig",
      "runtime": "hull",
      "image": "ghcr.io/brig-sh/claude-code@sha256:..."
    }
  ]
}
```

- `session` → `ref`, and its value is the full `<agent>@<label>` that every verb accepts, not a
  bare label.
- `profile` → `agent`, matching the CLI noun.
- `workspace` splits into `home` and `project`. Under #47 plus #6 those are two different mounts
[...]

PR #70, opened 2026-08-25
fix(wrap): open the workspace by descending where a guest can reach

### LINKED ISSUE #68: Workspace path is re-resolved mid-walk, so a guest can redirect a host write
## Problem

A guest that can write to a parent directory of its workspace can make brig write the guest git
credential helper outside the workspace, at a path the attacker chooses. That helper is mode 0755 and git
later runs it on the host, as the user, outside any sandbox.

`TestWorkspaceWritesNeverLandOutsideUnderARace` already catches this. It fails on `main` today.

## Measured

At `06cfd1e` with no other changes, on Linux, 8 cores, running the test alone:

| Condition | Runs | Failures | Rate |
| --- | ---: | ---: | ---: |
| plain | 100 | 4 | 4% |
| `-race` | 100 | 2 | 2% |
| `GOMAXPROCS=1` | 100 | 0 | 0% |
| six copies concurrently | 180 | 1 | 0.56% |

`GOMAXPROCS=1` hides it: the window needs the attacker running genuinely in parallel rather than
interleaved. The `-race` failures are the test's own assertion firing, not a data race report. The
detector has nothing to say because the contended state is the filesystem, not Go memory.

## The window

Not where the existing guard looks.

`openWorkspace` captures the workspace identity from `checkWorkspacePath`, opens the root, then calls
`rootIs` to compare the opened root against that identity. That closes the swap between the check and the
open, and it works: an instrumented reproducer recorded sixteen escapes and none of them went through
that gap.

The window is one level tighter, inside the walk that produces the identity in the first place.
`checkWorkspacePath` iterates the path components and calls `os.Lstat` on each. `os.Lstat` does not follow
the final component, but it does follow the intermediate ones, so the iteration that reaches the leaf
re-resolves the entire prefix from scratch. An attacker who flips a parent component into a symlink
between the parent iteration and the leaf iteration gets the leaf resolved through that symlink, and the
identity captured is the attacker's directory.

From there everything downstream agrees with itself. `os.OpenRoot` resolves the same path the same way and
lands on the same attacker directory. `rootIs` compares that directory against an identity that is already
that directory, so it passes. `os.Root` then confines every write faithfully, to a root that was wrong
before it was ever opened.

The value the whole scheme trusts is derived by a non-atomic re-resolution of the path it is meant to be
validating.

## Exploitability

The attacker is the guest, which is the threat model the surrounding comments already name. It needs:

1. Write access to a parent component of the workspace, so it can rename that component aside and drop a
   symlink in its place. A workspace nested under a directory another sandbox can write gives exactly
   this.
2. To win a sub-millisecond gap. That is retriable without limit: spin the flip continuously and trigger
   runs, at 2 to 4% each.

The payoff is not a stray file. It is host code execution as the user, through a credential helper git
executes.

Where the workspace's parent is under the user's own home and nothing hostile can write those components,
this is not reachable. The exposure is the nested case.

## Proposed solution

Stop resolving path strings, rather than adding a third guard to the same walk.

Open the first component with `os.OpenRoot`, then descend one component at a time with `Root.OpenRoot`.
Each step refuses a symlink escape by construction, the handle at the end is the workspace, and no path is
ever re-resolved from a string. The identity capture and the `rootIs` comparison then stop being
load-bearing and can go.

This guard has now been defeated twice by variations of one trick: first by making a component vanish
during the walk, then by flipping one between two syscalls of the same walk. Both fixes closed an
instance. The class stays open for as long as the workspace is reached by resolving a string more than
once.

## Acceptance criteria

- `TestWorkspaceWritesNeverLandOutsideUnderARace` passes at least 500 consecutive runs on Linux with real
  parallelism, where it currently fails around 4%.
- No path string is resolved more than once on the way to the workspace handle.
- A symlinked component is refused whether it appears before the walk, during it, or between the walk and
  the open, each covered by a test through the existing seam.
- The existing refusal messages keep naming the component that caused the refusal, since that is what
  makes the failure diagnosable.
- The test that drives the swap deterministically through the seam stays, and gains a case for the
  parent-to-leaf window specifically, so this instance cannot regress silently.

## Note on disclosure

Raising this in public because the repository has no private reporting channel yet, the exploit needs a
non-default layout, and there is no meaningful installed base. #34 is the issue that adds that channel,
and a finding like this one is the reason it should land.

pmoust:
## What this fixes

A guest that can write a parent directory of its workspace could make brig
write outside the workspace, at a path the guest chooses. brig prepares the
workspace on the host before boot, as the user brig runs as, and one of the
files it writes is the guest git credential helper: mode 0755, run by git on
the host, outside any sandbox. Closes #68.

The path was validated by resolving the path string more than once. Every
resolution follows symlinks from the top, so a guest that flips a parent
component into a symlink between two resolutions redirects the write, and the
comparison afterwards agrees with it, because both resolutions followed the
same planted link.

## How it works

The user brig runs as can swap a directory entry only inside a directory it can
write. The path splits at the first such directory:

- The prefix above it is opened by name in one call. The kernel resolves it and
  follows whatever links the system or an administrator put there; brig needs
  only search permission on the way.
- From there down, brig descends one component at a time against the directory
  already open above it, refusing every symlink. Nothing above a component is
  resolved again.

`trustedPrefix` finds the split from ownership and `access(2)`, so an ACL, or a
group beyond `NGROUPS_MAX`, counts the way the kernel counts it. As root, brig
trusts only what nobody but root could have written (root-owned, no group or
other write), which keeps the system's own links usable.

After the descent, `verifyStillOurs` confirms the handle still resolves to the
directory it opened, because the workspace reaches the runtime as a string that
another process resolves later.

## Behaviour change

A root-owned symlink inside a directory the user can write is now refused
rather than followed, with a message naming it. That is a layout an
administrator would have had to create inside user-writable territory.

## Limits

`checkWorkspacePathBeforeCreate`, on the `MkdirAll` path, still walks by string.
Worst case is one empty directory created through a planted link, in a directory
the attacker can already write, before the run is refused. The ACL question is
tracked in #71.

Not yet exercised against a real `brig run` on each platform; this is the path
that prepares the workspace for every run, so it wants one before merge.

## Tests

Each window is driven through a seam and fails against the previous code: a
parent swapped during the descent, a symlink in the tail, a trusted absolute
link on the prefix, and a search-only `0711` ancestor. `go test -race ./...`
and `script/smoke.sh` pass.

Closes #68. Refs #71, brig-sh/hull#7

### DIFF
--- a/internal/wrap/rootio.go
+++ b/internal/wrap/rootio.go
@@ -12,15 +12,34 @@ import (
 	"syscall"
 )
 
-// errPlantedSymlink marks a refusal caused by a symlink inside the workspace
-// that leads out of it, so a caller that otherwise shrugs off an I/O error --
-// trustGuestCwd treats an unreadable state file as nothing to do -- can tell
-// this one apart and fail the run instead of quietly carrying on.
+// errPlantedSymlink marks a refusal because the workspace, or the path to it,
+// cannot be trusted: a symlink inside it that leads out, one on the way to it
+// that the guest could have planted, or a component that moved while brig was
+// looking. A caller that otherwise shrugs off an I/O error -- trustGuestCwd
+// treats an unreadable state file as nothing to do -- can tell this one apart
+// and fail the run instead of quietly carrying on. A plain configuration
+// error, such as a regular file where a directory was named, does not carry
+// it.
 var errPlantedSymlink = errors.New("a symlink in the workspace leads out of it")
 
-// afterWorkspaceCheck runs between the path check and the open. It does
-// nothing outside tests; see openWorkspace.
+// afterWorkspaceCheck runs after the workspace has been opened and before the
+// root is handed back. It does nothing outside tests; see openWorkspace.
 var afterWorkspaceCheck = func() {}
+
+// duringWorkspaceWalk runs at the start of the last step of the descent,
+// before the workspace itself is looked at. It does nothing outside tests.
+//
+// This is the gap an attacker gets against a walk that resolves strings: a
+// parent flipped here poisons everything a by-name walk does next, and then
+// everything downstream agrees with itself about the wrong directory. Against
+// the descent it is meant to do nothing, and the test that drives it says so.
+var duringWorkspaceWalk = func() {}
+
+// betweenLstatAndOpen runs inside one step of the descent, after the entry has
+// been looked at and before it is opened, with the entry's name. It does
+// nothing outside tests. This is the one window a step has, and the identity
+// check after the open is what closes it.
+var betweenLstatAndOpen = func(name string) {}
 
 // workspaceRoot is the workspace opened once as an [os.Root]. Every host-side
 // read and write brig makes inside the workspace goes through it.
@@ -42,25 +61,16 @@ var afterWorkspaceCheck = func() {}
 type workspaceRoot struct {
 	root *os.Root
 	dir  string
-	// ident is the directory the path check ended on, captured while the path
-	// was known to be symlink-free. Everything that asks "is this still our
-	// workspace?" compares against this, never against a fresh resolution of
-	// the path.
+	// ident is the directory the descent ended on, read from the handle it
+	// holds. Before the path is handed to the runtime, verifyStillOurs resolves
+	// it afresh and compares against this: a resolution can only disagree with
+	// the handle, and disagreement refuses.
 	ident os.FileInfo
 }
 
 // openWorkspace creates the workspace if it is not there yet and opens it as a
 // root.
 func (c *Config) openWorkspace() (*workspaceRoot, error) {
-	// The walk below inspects a path *string*; MkdirAll and OpenRoot then
-	// resolve that same string again, from scratch. Two resolutions, and the
-	// guest owns a component in between them whenever a workspace is nested
-	// under a directory a sandbox can write -- which is the case the walk's own
-	// doc comment names as the reachable one. Flipping that component between
-	// the check and the open redirected everything brig writes next, and one of
-	// those things is a mode-0755 credential helper, so the race was worth
-	// running: it landed outside the workspace in 43 attempts.
-	//
 	// Cleaning first because the walk cleans and the callers did not, so
 	// `<tmp>/link/../real` was checked as `<tmp>/real` -- skipping `link`
 	// entirely -- and then opened as written, through the link.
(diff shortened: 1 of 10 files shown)


internal/wrap/rootio.go
```diff
@@ -80,63 +90,253 @@ func (c *Config) openWorkspace() (*workspaceRoot, error) {
...
+			_ = root.Close()
+			return nil, nil, fmt.Errorf("refusing to use %s as the workspace: %s could not be "+
+				"read while brig was checking the path (%v), so something is moving it: %w",
+				abs, p, err, errPlantedSymlink)
+		}
+		if info.Mode()&os.ModeSymlink != 0 {
+			target, _ := root.Readlink(name)
+			_ = root.Close()
+			return nil, nil, plantedSymlinkErr(abs, p, target)
+		}
```
cmainas, line 184: I noticed that in other parts of this PR a file is checked not only as a symlink but also as a directory with `isDir`.  I am not sure if this is intentional, but I think is necessary (see generic review comment).

cmainas (commented):
Overall the PR looks good and resolves the issue. However, it opens a tiny window  which can lead to hangs. This cna only happen in nested workspaces from `openWorkspaceHandle` when traversing a path. While going through all files, the loop checks against links but it does not perform any other check against other types of files. Therefore, in a very small window, if a file changes and becomes a non-regular, non-directory and non-link file (e.g. fifo) opening it will lead to hangs.

internal/wrap/rootio.go
pmoust: Confirmed, and the hang is real. Reproduced it directly:

    os.Root.OpenRoot on a fifo -> HUNG (no return after 4s)

Opening a fifo waits for a writer and nothing times it out, so a guest that can swap a parent component -- the capability this whole change assumes it has -- can stop the boot rather than redirect it.

One correction on the fix, because it cost me a wrong first attempt. `IsDir` alone does not close it. The swap happens *after* the look, so the check passes on a directory and the open still lands on the fifo. With only the `IsDir` check added, the new test still hangs.

What closes it is opening with `O_DIRECTORY` first, where the kernel refuses a non-directory at open time:

    OpenRoot                 HUNG
    OpenFile O_DIRECTORY     errored: openat pipe: not a directory
    control: real dir        <nil>

So the step now does both: `IsDir` to refuse a component that is already a non-directory, which also makes the descent consistent with `resolveTrusted` and `checkWorkspacePathBeforeCreate` as you noted, and an `O_DIRECTORY` guard immediately before `OpenRoot` so no swap can block us whatever it puts there.

One limit worth stating: `os.Root` has no constructor that takes a descriptor, so `OpenRoot` still resolves the name a second time after the guard. A swap landing between those two opens is caught by the existing `SameFile` check, which is now reachable because the guard has established the entry is a directory. Closing that last re-resolution would mean walking with `/dev/fd` or `/proc/self/fd`, which is platform-specific -- happy to do it if you want it airtight rather than merely non-blocking.

Test added in `f563a58`, driving the swap through the `betweenLstatAndOpen` seam. It runs the open in a goroutine with a bound, because a regression here does not fail the suite, it stops it. Verified it bites by removing the guard.

cmainas: Indeed,, `isDir` is not enough by itself. However, closing the re-resolution sounds straightforward and cheap. Do not really see the reason not to do so.

pmoust: Done in `6890c74`. You were right that it is cheap.

The kernel names an open descriptor as a path, so the root is built from the descriptor the guard already opened: `/proc/self/fd` on Linux, `/dev/fd` elsewhere. There is no second lookup left to race.

Checked on both platforms before writing it. The decisive check is the last line: rename the component away, and a root built this way still reads the directory it was opened on.

    OpenRoot(fdpath) ok, name = /proc/self/fd/4
    sees inner through the fd root: yes
    after rename, fd root still correct: yes (no re-resolution)

One open per component now, where there were two.

scape verification re-run on Linux after both changes, since they touch the same loop: 200 runs, 0 escapes. Control on unfixed main is 4 in 60.

cmainas (approved):
Thank you @pmoust for the updates.

Issue #22, opened 2026-08-25
Test each public security claim in CI


pmoust:
### Problem

The security doc makes specific, testable claims: the guest sees one
directory, a symlinked workspace is refused, an image under our registry with
a bad signature stops the boot, credentials are named and not ambient. Nothing
in CI proves any of them, so they hold until a refactor quietly ends one.

### Proposed solution

Add a claims suite that runs on every merge. One test per public claim:

- a fixture file outside the workspace is not readable from the guest
- a symlinked `--workspace` is refused
- an image under `ghcr.io/brig-sh/` with a broken signature refuses to boot
- only the declared credential names reach the guest environment
- a denied billing key does not reach the guest
- the digest that booted is the digest that was verified, after #14

Link each test from the sentence it defends in the security doc, so a claim
without a test is visible in review.

### Alternatives considered

Cover this with manual tests. That is where some of it lives today, and a
manual test does not run on the pull request that breaks it.

PR #73, opened 2026-08-25
docs(security): scope the no-host-read claim to credentials

### LINKED ISSUE #35: Scope the claim that a run reads nothing outside the workspace
### Problem

`docs/security.md` states that nothing on the `run`, `exec` or `shell` path reaches a file outside the
workspace, or a host command. The git plumbing does both on every invocation: it shells out to `git config`
in the directory brig was invoked from, to resolve the commit identity, and it reads gh's hosts file to
find a login to pair with the forwarded token. That file normally holds an OAuth token.

Neither read is a credential read for brig's own purposes, and the scoped claim in the surrounding
paragraph, that a run on a shipped profile reads no host credential source, survives. The broader sentence
does not.

### Proposed solution

Narrow the sentence to what the code does, and name the two reads with what each is for. Keeping the
distinction visible is worth more than the broader wording, since the point being made is about credential
sources.

### Acceptance criteria

- The sentence names the git config read and the gh hosts file, or is scoped to credential sources
  explicitly.
- A test in #22 asserts no other host path is opened on the run path.

pmoust:
## Summary

`docs/security.md` said that nothing on the `run`, `exec` or `shell` path
reaches a file outside the workspace or a host command. The git plumbing does
both on every invocation that loads a profile, which includes `stop`, `rm` and
`env` as well: it runs `git config --get` in the invoking directory for the
commit identity it forwards, and when `github.user` is unset it reads the login
from gh's `hosts.yml`. Neither is a credential read, and the paragraph's real
promise, that a run on a shipped profile reads no host credential source,
survives. The sentence as written did not.

The paragraph now says exactly what is read, for what, and that no setting turns
either read off: `BRIG_GIT_IDENTITY`, `BRIG_GIT_CONFIG` and `BRIG_GIT_USER`
change what brig does with the answers, not whether it asks. That last part was
checked against the built binary with a logging `git` shim.

## Related issues

Closes #35

## Changes

- One paragraph in `docs/security.md`, rewritten. Nothing else in the file.

## Checklist

- [ ] ~~I have added or updated tests covering the change~~
- [ ] ~~I have run `go test ./... -race`, if the change touches concurrency, subprocesses or the daemon~~

The claim as rewritten is one the claims suite can defend: the assertion is that
on a run, the only host paths opened outside the workspace are the invoking
directory's git config and gh's `hosts.yml`, and that the `oauth_token:` line
in the latter is never read. That test belongs with #22.

No behaviour changes. This narrows a published claim about the second promise to
what the code does. One thing came out of verifying it: a FIFO planted at
`hosts.yml` blocks brig on open. That file sits outside the workspace, so the
guest cannot plant it, and it is filed separately rather than fixed here.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -32,11 +32,20 @@ them for itself.
 
 ## Credentials
 
-**A run on a shipped profile reads no host source.** Nothing on the `brig run`,
-`exec` or `shell` path reaches a keychain item brig did not write, a file
-outside the workspace, or a host command. Your host login enters brig's own
-store once, when you type `brig secret import <profile>`, and every run
-afterwards reads only that store:
+**A run on a shipped profile reads no host credential source.** Nothing on the
+`brig run`, `exec` or `shell` path reaches a keychain item brig did not write,
+a credential file outside the workspace, or a host command that produces one.
+Two host reads do happen, and no setting turns either off. brig runs `git
+config --get` in the directory you invoked it from, for `user.name`,
+`user.email` and `github.user`, to resolve the commit identity forwarded into
+the guest. If `github.user` comes back empty it then reads the `user:` line
+from the stanza for your git host in gh's `hosts.yml`, under `$GH_CONFIG_DIR`
+or `~/.config/gh`, to find the login that pairs with the forwarded token. That
+file usually carries gh's own OAuth token as well; brig takes the login and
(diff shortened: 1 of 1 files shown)


ananos (approved):
Thanks @pmoust! Scoping the claim to credentials and naming the two reads that do happen makes the page honest -- and it pairs nicely with the hosts.yml hardening in #102. LGTM.

Issue #23, opened 2026-08-25
Remove BRIG_CREDENTIALS_CMD


pmoust:
### Problem

`BRIG_CREDENTIALS_CMD` is deprecated in the help text, and it warns at runtime
that it "goes in the next release". The replacement exists:
`brig secret import <profile> <name> --from-command` does the same job once,
at import, instead of on every run.

The promise has a deadline in it. Keeping the variable past that release makes
the next deprecation notice worth less.

### Proposed solution

Remove the variable and the host-read path that exists only to serve it. When
it is set, exit with the one-line replacement rather than ignoring it
silently. Note the removal in the release notes, under breaking changes.

### Alternatives considered

Keep it as an undocumented alias. It costs one branch, and it keeps a live
credential command running on every boot, which is the reason it was
deprecated.

Issue #20, opened 2026-08-25
Publish a support matrix, and settle the macOS floor


pmoust:
### Problem

The README says macOS needs `hull` on PATH and Linux needs its runtime. It
names no minimum OS version, no tested runtime versions and no architectures.

#4 is what that costs: a macOS 14.5 host, a sandbox that never becomes ready,
and `dyld: missing symbol called` from the VMM. The requirement was real and
the user met it as a crash, not as a message.

### Proposed solution

1. Publish a table: host OS and minimum version, architecture, runtime and the
   versions tested, per platform.
2. Answer the question behind it in the ticket: is the macOS floor a
   Virtualization.framework API that brig or hull needs, or is it the version
   the work happened on? The answer decides whether the floor can come down.
3. Check the floor at runtime and say so plainly when the host is below it,
   rather than letting the VMM fail. `brig doctor` (#9) is where that check
   belongs.
4. Say which versions CI covers, so the table is a claim with evidence.

### Alternatives considered

Keep the requirements informal and fix reports as they come. Every one of
those reports costs the reporter a broken first run, which is the run that
decides whether they keep the tool.

PR #75, opened 2026-08-25
docs: add a compatibility policy and name removal releases

### LINKED ISSUE #60: State the compatibility policy for 0.x
### Problem

There is no statement anywhere about what a version number promises. Which surfaces are covered, how long
a deprecated spelling survives, and what may change in a patch release are all currently answered by
whatever the last commit did.

That question is live right now. Deprecation notices in the tool say a setting goes "next release" without
naming which. Work in flight assumes a two-release window for old command spellings, and that window is
written in a proposal rather than in a policy.

### Proposed solution

A short page: which surfaces are covered by the promise, which are explicitly not, what 0.x means as
opposed to 1.0, how long a deprecated spelling lives once its replacement ships, and where breaking
changes are recorded. Keep it short enough that it can be kept true.

Cover at least: the CLI grammar, exit codes, the profile schema, the settings, the filesystem layout, the
JSON output, and the daemon protocol. Some of those should be declared unstable on purpose, which is a
useful thing to say out loud.

### Acceptance criteria

- Every public surface appears in the policy as covered or not covered.
- Deprecation messages name the release that removes the thing, rather than "next release".

pmoust:
## Summary

Nothing said what a version number promised. Which surfaces were covered, how
long a deprecated spelling survived, and what a patch release could change were
all answered by whatever the last commit did, and three deprecation messages in
the tool said a thing "goes in the next release" without naming one.

`docs/compatibility.md` states it, short enough to stay true. Every public
surface is classified from the source as covered or explicitly not covered, the
deprecation window is two releases with removal in the third, and breaking
changes are recorded where the changelog is generated from. The `brigd` line
protocol, the JSON output and the human-readable text are declared unstable on
purpose, which is the more useful half of the page.

The three "next release" messages now name `v0.1.0-rc17`, with tests asserting
the release name so a later edit cannot slide back to "next".

## Related issues

Closes #60
Refs #23, #20

## Changes

- `docs/compatibility.md`, linked once from the README's Documentation list.
- `BRIG_CREDENTIALS_CMD` and `hostCredential:` warnings, and the help-text line,
  name `v0.1.0-rc17` instead of "next release". `cmd/brig/deprecated_test.go`
  and `internal/wrap/config_test.go` assert it.

## Two things to know before merging

`BRIG_CREDENTIALS_CMD` is being removed in #23, in flight now. If that lands in
rc17 the message is exactly right; if it slips, the policy's own rule is what
decides the number, not this text.

The policy says the support matrix lives in the README's platform sections. #20
is putting it in `docs/support.md`; whichever merges second fixes the pointer.

## Checklist

Message text only. No behaviour changes on either promise.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -1,0 +1,106 @@
+# What a version number promises
+
+This page says which parts of brig you can build on and which you cannot, how
+long a name survives after it is deprecated, and where a break is written down.
+It is meant to stay short enough to stay true, so it names surfaces rather than
+restating what each one does; the README and the other docs are where the
+detail lives.
+
+## What 0.x means
+
+brig has no stable release yet. There are twelve `v0.1.0-rc` tags and nothing
+without the `-rc`, so every version to date is a release candidate. Until a
+`1.0` ships, treat the whole tool as prerelease: a covered surface can still
+change between releases, but it changes under the deprecation rule below and
+the change is recorded where [breaking changes](#where-breaking-changes-are-recorded)
+go. `1.0` is the point at which the covered surfaces stop moving without a
+major version bump. The unstable surfaces below carry no such promise even
+then, and they say so on purpose.
+
+## What the promise covers
+
+These are the surfaces a script or a profile can depend on. Within `0.x` they
+change only through a deprecation, and across a `1.0` they will not change
+without a major bump.
+
+- **The CLI verbs.** `run`, `create`, `exec`, `shell`, `stop`, `rm`, `ls`,
+  `reset`, `env`, `version`, `help`, the profile verbs (`profiles`, and
+  `profile ls|export|import|edit|rm`, plus the top-level `export` and `import`
+  aliases), and the secret verbs (`secret create|read|update|delete|ls|import`).
+  The list is `cmd/brig/main.go`.
+- **The run-line flags.** `-n`/`--name`, `-t`/`--image`, `-w`/`--workspace`,
+  `-m`/`--memory`, `--cpus`, `-d`/`--detach`, and `--skills`. Everything else
+  on a run line belongs to the agent, by design, so the set brig owns is the
+  set here and no more.
+- **The settings documented in the README.** Every `BRIG_*` variable in the
+  README's [environment variables](../README.md#environment-variables) tables,
+  read under the `BRIG_<AGENT>_<KEY>` then `BRIG_<KEY>` order stated there. A
+  variable that is not in those tables is not part of this promise; see below.
+- **The profile schema.** The fields `brig profile export` writes and
+  documents in its header: `name`, `desc`, `kind`, `image`, `guestHome`,
+  `binary`, `secrets`, `env`, `files`, `volumes`, `deny`, `mem`, `cpus`,
+  `hypervisor`, `runtimeBin`, `rootfsType`, `genericBoot` and `reserved`. The
+  types are `internal/profile`. The parser refuses an unknown field rather than
+  ignoring it, so adding one is a change you will see, not a silent no-op.
+- **The exit status, coarsely.** `0` is success and non-zero is failure. On
+  `run`, `exec` and `shell` brig replaces itself with the runtime, so the
+  status you get back is the agent's own. The split between success and failure
+  is covered; a specific non-zero number is not a contract yet.
+- **The filesystem layout.** Workspaces under `~/brig/<agent>` (a named session
+  appends `-<slug>`); brig's own runtime sockets under `~/.brig`; and custom
+  profiles under the config directory, `$XDG_CONFIG_HOME/brig` or
+  `~/.config/brig` when that is unset. `brig secret` stores each item in the
+  macOS login keychain under the service `sh.brig.secret`.
+
+## What is explicitly not covered
+
+Depending on any of these is depending on something we intend to move.
+
+- **The `brigd` line protocol.** The JSON operations in
+  [docs/brigd.md](brigd.md) are unstable on purpose. `brigd` is a small,
+  optional daemon that most people never run, and its protocol is free to
+  change between releases without a deprecation. Drive brig through the CLI if
+  you want the promise above.
+- **The JSON output.** There is little of it today, `brig profile export
+  --json` being the main case, and its shape is not fixed. Do not parse it in
+  anything you need to keep working across releases.
+- **The human-readable output.** The wording of messages on stdout and stderr,
+  the warnings, and the column layout of `ls` and `profiles` are for people to
+  read, not for `grep` to match. They get reworded whenever a clearer sentence
+  turns up.
+- **Undocumented and test-only settings.** Any `BRIG_*` variable not in the
+  README tables, including the `BRIG_TEST_*` ones, exists for brig's own use
+  and can change or vanish without notice.
+- **Deprecated spellings.** Anything named in the next section is already on its
+  way out and covered only for its remaining window.
+
+## How long a deprecated spelling lives
+
+When a replacement ships, the old spelling keeps working for two releases and
+is removed in the third. While it works, brig prints a notice that names the
+release which removes it, so the deadline is a version number you can read and
+not a vague "soon".
+
+The spellings deprecated before this policy are already scheduled: `brig
+agents`, `brig template`, the profile keys `hostCredential:`, `forward:` and
+`statePaths:`, and the setting `BRIG_TEMPLATE_DIR`. Their notices name
+`v0.1.0-rc17` as the release that removes them.
+
+`BRIG_CREDENTIALS_CMD` is gone already. A run that still sets it fails and
+names the replacement, rather than warning and carrying on.
+
+## Where breaking changes are recorded
+
+A change that breaks a covered surface carries a `BREAKING CHANGE:` footer on
+its commit, per the Conventional Commits rule in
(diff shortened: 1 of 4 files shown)


docs/compatibility.md
```diff
@@ -0,0 +1,115 @@
...
+
+This page says which parts of brig you can build on and which you cannot, how
+long a name survives after it is deprecated, and where a break is written down.
+It is meant to stay short enough to stay true, so it names surfaces rather than
+restating what each one does; the README and the other docs are where the
+detail lives.
+
+## What 0.x means
+
+brig has no stable release yet. There are twelve `v0.1.0-rc` tags and nothing
```
ananos, line 11: `v0.1.0-rc17` shipped on 28 Aug, so the count is out of date. Perhaps say "every version to date is a release candidate" without the number -- then it cannot go stale again between drafting and merging.

docs/compatibility.md
```diff
@@ -0,0 +1,115 @@
...
+you can read rather than a vague "soon".
+
+The spellings deprecated before this policy are scheduled unevenly, and this is
+what each actually does today:
+
+- `BRIG_CREDENTIALS_CMD` is gone. It was deprecated in `v0.1.0-rc16` and
+  removed; a run that still sets it fails and names the replacement, rather
+  than warning and carrying on.
+- The `hostCredential:` profile key prints a runtime notice naming
+  `v0.1.0-rc17` as the release that removes it. It was deprecated in
```
ananos, line 91: This names `v0.1.0-rc17` as the release that removes `hostCredential:`, and rc17 has already shipped -- the field is still in `internal/profile` (it has left the shipped profiles, which is the other half). So the page would land stating a deadline that has passed.

Either remove the field and keep the promise, or move the removal to the next release and say so here. The first is truer to the policy this page is introducing; the second is fine too, as long as the notice in `main.go` moves with it. The test pins the string, so it will keep passing either way -- which is why it is worth catching by eye now.

ananos (commented):
Worth having -- naming the removal release in the notice, and making that named release the authority, is the part that turns a deprecation into a promise someone can plan against.

Two bits have gone stale while it sat in draft, inline.

docs/compatibility.md
ananos, line 11: There are **13** `v0.1.0-rc` tags now, not twelve. This is the second time this number has gone stale on this branch -- it was eleven when I first read it.

So I would drop the count rather than correct it: "every version to date is a release candidate, and nothing without the `-rc` has shipped" says the same thing and cannot rot. The number adds no information a reader needs, and it is the one sentence on the page guaranteed to be wrong again by the next release.

docs/compatibility.md
```diff
@@ -0,0 +1,115 @@
...
+major version bump. The unstable surfaces below carry no such promise even
+then, and they say so on purpose.
+
+## What the promise covers
+
+These are the surfaces a script or a profile can depend on. Within `0.x` they
+change only through a deprecation, and across a `1.0` they will not change
+without a major bump.
+
+- **The CLI verbs.** `run`, `create`, `exec`, `shell`, `stop`, `rm`, `ls`,
```
ananos, line 26: This list and brig's actual verbs have diverged a long way. Asked the binary built from current main:

```
agent completion doctor info ls policy rm run secret sh stop telemetry version
```

So the page covers `create`, `exec`, `shell`, `reset`, `env`, `help` and `profiles`, and does not mention `agent`, `completion`, `doctor`, `info`, `policy`, `telemetry` or `sh`. Some of that is renaming (`shell` to `sh`, `env` deprecated toward `info`), but `doctor`, `policy`, `telemetry` and `completion` are simply newer than the branch.

That matters more here than on an ordinary page: a promise that names the covered verbs, and omits half of them, tells a script author their `brig policy` usage is uncovered when it is not.

ananos (commented):
Still worth having -- naming the removal release and making it the authority is the part that turns a deprecation into something someone can plan against.

The difficulty is that this branch is 30 commits behind main and has been open since 4 September, and this is the one page where being behind is worse than being absent. Every other doc that goes stale is merely wrong; a compatibility page that goes stale promises coverage for surfaces that are not there and omits the ones that are, and a reader has no way to tell. Most of what follows is not a flaw in the writing -- it is main having moved under it.

I checked each of these against the current tree rather than reading, and inlined them. Nothing here needs re-thinking, but the page wants a pass over its lists before it comes out of draft.

The underlying tension is worth a thought too: four of the six are hand-kept lists that duplicate something the binary already knows, which is the same drift #145 is about for verbs and flags. Where a list can be derived -- the verbs and the run-line flags both come off `brigFlags` and the dispatch table -- deriving it in a test would keep this page honest the way `6ed914b` just did for the completion test.

docs/compatibility.md
```diff
@@ -0,0 +1,115 @@
...
+  `reset`, `env`, `version`, `help`, the profile verbs (`profiles`, and
+  `profile ls|export|import|edit|rm`, plus the top-level `export` and `import`
+  aliases), and the secret verbs (`secret create|read|update|delete|ls|import`).
+  The list is `cmd/brig/main.go`.
+- **The run-line flags.** `-n`/`--name`, `-t`/`--image`, `-w`/`--workspace`,
```
ananos, line 31: Same drift, and this one closes the set explicitly -- "the set brig owns is the set here and no more" -- which makes an omission a wrong promise rather than an incomplete one.

What the run line actually offers today:

```
--cpus --detach --home --image --json --mem --network --no-project --offline --skills
```

Against the seven here: `--name`, `--workspace` and `--memory` are not among them, and `--home`, `--json`, `--network`, `--no-project`, `--offline` and `--mem` are missing from the page. `--network` is the one I would least want left out, since it is the flag that selects an isolation posture.

This is the list I would most like to see derived rather than typed: the rows are already in `brigFlags` with a position each, so a test could assert the page names exactly the `posRun` set.

docs/compatibility.md
```diff
@@ -0,0 +1,115 @@
...
+  README's [environment variables](../README.md#environment-variables) tables,
+  read under the `BRIG_<AGENT>_<KEY>` then `BRIG_<KEY>` order stated there. A
+  variable that is not in those tables is not part of this promise; see below.
+- **The profile schema.** The fields `brig profile export` writes and
+  documents in its header: `name`, `desc`, `kind`, `image`, `guestHome`,
+  `binary`, `secrets`, `env`, `files`, `volumes`, `deny`, `mem`, `cpus`,
+  `hypervisor`, `runtimeBin`, `rootfsType`, `genericBoot` and `reserved`. The
+  types are `internal/profile`. The parser refuses an unknown field rather than
+  ignoring it, so adding one is a change you will see, not a silent no-op.
+- **The exit status, coarsely.** `0` is success and non-zero is failure. On
```
ananos, line 45: This says a specific non-zero number is not a contract. README now says the opposite, under "Exit codes" at line 222:

> brig returns a small, stable set of exit codes so a script can tell what went [wrong]

followed by a table giving each of `0`-`6` its own meaning. And #154 is building on that: `internal/exitcode` produces the same six for `brigd`, so brigd's clients will branch on them too.

Of the two, README's is the promise that is actually being kept, so I would bring this into line with it rather than the other way round -- otherwise the compatibility page is the more conservative of two documents that disagree, which is the confusing direction.

The sentence above it needs the same exception `docs/security.md` just gained on main: brig no longer always replaces itself with the runtime -- under `--json` it stays as the agent's parent to print one status line. The exit status is still the agent's own, so the promise holds; the mechanism described does not.

docs/compatibility.md
```diff
@@ -0,0 +1,115 @@
...
+  `run`, `exec` and `shell` brig replaces itself with the runtime, so the
+  status you get back is the agent's own. The split between success and failure
+  is covered; a specific non-zero number is not a contract yet.
+- **The filesystem layout.** Workspaces under `~/brig/<agent>` (a named session
+  appends `-<slug>`); brig's own runtime sockets under `~/.brig`; and custom
+  profiles under the config directory, `$XDG_CONFIG_HOME/brig` or
+  `~/.config/brig` when that is unset. `brig secret` stores each item in the
+  macOS login keychain under the service `sh.brig.secret`.
```
ananos, line 53: No longer macOS-only: #153 merged this morning, so `brig secret` also stores into a freedesktop Secret Service keyring on Linux (`internal/secret/secretservice_linux.go`), one item per secret in the default collection under the same `sh.brig.secret` service attribute.

Since this bullet is about the filesystem layout a user can depend on, it is worth saying both, and that the service name is the same on each.

ananos (commented):
Thanks @pmoust, a compatibility page is a good idea. A few things before it can go in:

- The branch is 90 commits behind main and has conflicts, so it needs a rebase first. Main now has `docs/stability.md`, which covers much of the same ground and disagrees with this page in places (exit codes, the `--json` output, `~/.brig`). It may be simpler to fold what is missing into that page.
- The `hostCredential:` warning now says the key goes in `v0.1.0-rc17`, but rc17 was tagged before this branch was cut, and main has since removed the key.
- The verb and flag lists include spellings that are deprecated in the same tree (`create`, `exec`, `shell`, `-n`, `-w`, ...) and leave out ones we teach (`sh`, `agent`, `--home`, `--network`).
- "A specific non-zero number is not a contract yet" contradicts the README's exit-code section.
- Small ones: `../README.md#linux` has no target, and the "twelve rc tags" count is out of date.

PR #76, opened 2026-08-25
ci: add a claims suite tying security.md promises to tests

### LINKED ISSUE #22: Test each public security claim in CI

pmoust:
## Summary

`docs/security.md` makes specific, testable promises, and nothing in CI proved
any of them, so each held until a refactor quietly ended one. Several came
close recently. This ties every testable sentence on that page to the test that
defends it, and fails CI when a row names a test that no longer exists, so a
claim cannot be left silently undefended after a rename.

## Related issues

Closes #22

## Changes

- `docs/claims.md`: fourteen rows, each the exact sentence, where it appears,
  and a typed token naming its defence: `go:TestName`, `smoke:<check text>`,
  `vm`, or `pending`.
- `script/check-claims.sh`: resolves every token, `go:` against
  `go test -list ./...` and `smoke:` against the `ok` lines in `script/*.sh`,
  and fails on any that resolves to nothing. `--self-test` first runs it against
  a row that names a missing test and confirms it fails, so the guard is proven
  live before it is trusted.
- CI step `claims table`: self-test, then the check. `make claims` runs the
  same thing locally.
- Four new smoke checks against the stub runtime, which already logs the
  environment it receives: only the declared credential names reach the guest,
  a denied billing key does not, a `scheme://` value is refused, and no
  credential is written into the workspace.
- `script/claims-vm.sh` and `make claims-vm` for the two rows only a real VM can
  prove (the guest sees one directory; the keychain, SSH agent and other host
  directories are not reachable). It boots a sandbox where a runtime is present
  and skips cleanly where none is.
- Two rows marked `pending`, for the fail-open switch fix in #63 and the digest
  pin in #67; each flips to a real token when its PR merges.
- One link added near the top of `docs/security.md`; nothing else there.

## Checklist

`claims-vm.sh` was seen to skip cleanly on a box with no runtime. Its boot half,
the two VM rows, has not been run against a real sandbox yet, and it should be
once before this merges, since those two rows are the ones the page leads with.

Verified here beyond the gates: the checker refuses a row naming a missing Go
test and a row naming a missing smoke check, and its self-test passes.

No behaviour changes. This is the mechanism that keeps the claims about both
promises honest after the fact: every sentence the page makes is now either
defended by a named test, marked as needing a VM, or marked pending a specific
PR, and CI refuses a fourth state.


docs/claims.md
```diff
@@ -0,0 +1,48 @@
...
+| "The `deny` list keeps a metered key from being forwarded by accident" (a denied billing key does not reach the guest) | security.md, Things brig does not claim | `smoke:the metered key is refused, and says why` + `smoke:a denied billing key never reaches the guest env line` + `smoke:the denied key's value never reaches argv` + `go:TestDenyAppliesToRefdValues` | existing + new |
+| "A `scheme://` value read from the environment is refused as an unresolved secret-manager reference." | security.md, Credentials (the three rules) | `smoke:a secret-manager reference is not forwarded` + `go:TestUnresolvedReferencesAreRejectedButOrdinaryURLsAreNot` | existing |
+| "Nothing is written into the workspace." (no credential lands in the workspace) | security.md, Credentials | `smoke:no credential is written into the workspace` | existing |
+| "`BRIG_ALLOW_DENIED=false` does not forward the denied variable" (the fail-open switch) | security.md, Credentials | `go:TestOffSpellingsDoNotForwardADeniedCredential` | existing |
+| "the digest that booted is the digest that was verified" | security.md, Guest images | `go:TestVerifyResolvesVerifiesAndPinsAMatchingDigest` | existing |
```
ananos, line 48: The table has no rows for the network claims, and #130 added two to `security.md` yesterday:

- a policy bound to a sandbox is enforced at the gateway, and a run that cannot enforce it is refused rather than booted unconstrained;
- `--network isolated` gives the sandbox a network of its own, and is refused where brig owns no network.

Both have tests already, so they would come in as `existing` rather than `pending`: the refusals are covered in `script/smoke.sh` and in `internal/runtime` (`TestNerdctlRefusesAPolicyItCannotEnforce`, `TestIsolatedIsRefusedWhereBrigOwnsNoNetwork`), and the enforcement itself is measured in `docs/manual-tests/egress-policy.md`, which is `vm` territory.

Worth rebasing on main and adding them -- a claims table that misses the newest claims is the exact failure it exists to prevent. Happy to write the rows if that helps.

ananos (commented):
I like this a lot -- a claim with no test being a visible row rather than a silent gap is the right mechanism, and the `vm` / `pending` states keep it honest about what CI can actually prove.

One gap to close before it comes out of draft, inline.

docs/claims.md
```diff
@@ -0,0 +1,48 @@
...
+  "unknown": where a pending row also records that the current behaviour is
+  wrong, that is called out in the row itself, so pending is never mistaken for
+  a passing claim.
+
+## The table
+
+| Claim (exact sentence from the docs) | Where it appears | Defended by | Status |
+| --- | --- | --- | --- |
+| "the agent sees one directory and the credentials you chose to give it, and nothing else on the host." | security.md, intro | `vm` | needs a VM |
+| "It does not have your keychain, your SSH agent, your secret manager, or any other directory on the host." | security.md, The boundary | `vm` | needs a VM |
```
ananos, line 36: The table still has no rows for the network and policy claims, and #130 merged three days ago, so `docs/security.md` on main now carries two promises this suite does not track:

- a policy bound to a sandbox is enforced at the gateway, and a run that cannot enforce it is refused rather than booted unconstrained;
- `--network isolated` gives the sandbox a network of its own, and is refused where brig owns no network.

Both already have tests, so they come in as `existing` rather than needing anything written. The refusals are covered by `go:TestNerdctlRefusesAPolicyItCannotEnforce` and `go:TestIsolatedIsRefusedWhereBrigOwnsNoNetwork` -- and note the first of those got stronger this morning: #149 moved the GUI refusal into `nerdctl.CanRun` so the join path is covered too, which is the same shape. The enforcement itself is measured in `docs/manual-tests/egress-policy.md`, which is `vm` territory by this table's own rule.

I am flagging it again rather than letting it go because it is precisely the failure this page exists to prevent: the newest security claims are the ones with no row, and a reader would take the table's silence for "nothing more is promised". Happy to write the two rows if that is quicker than you doing it.

ananos (commented):
I still like this a lot. A claim with no test being a visible row rather than a silent gap is the right mechanism, and retiring the `pending` rows in `059be4d` was the right call -- `pending` was the one status that could quietly mean "this claim is not defended" while looking tracked.

The typed-token scheme has aged well too: `go:` checked against `go test -list`, `smoke:` matched on its `ok` line, `vm` for what only a real boot can show. That a stub cannot prove isolation, and that nothing here pretends otherwise, is the honest part.

One gap, unchanged since I first read this, and now the more serious for it. Plus a note on where the branch sits.

ananos:
Separately from the table itself: this branch is 85 commits behind main and has been open since 25 August. The suite reads `docs/security.md` and matches `ok` lines in `script/smoke.sh`, and both have moved a long way since -- `smoke.sh` in particular gained cases from #130, #137, #146 and #158. So the checker may well fail on the rebase for reasons that are not defects, simply because a token no longer resolves.

Worth rebasing before spending effort on the rows, so the run you are debugging is the real one.

ananos (commented):
Thanks @pmoust, tying the security claims to tests is a great idea. Some points, mostly on `claims-vm.sh`:

- The branch is 145 commits behind main and has conflicts, so it needs a rebase. After the rebase, one smoke check has a new name on main, so `check-claims.sh` will report it missing.
- `claims-vm.sh` uses the real brig state and the usual sandbox name. Running it restarts an open `brig run claude` sandbox on a temp dir and stops it at the end. Pointing `BRIG_STATE_DIR` at a temp dir, the way `smoke.sh` does, would avoid that.
- The workspace check matches `*inside*`, which also matches cat's own "No such file" error, so it passes when the workspace is not mounted. The outside check reads the host path from inside the guest, where that path never exists, so it cannot fail either.
- The script runs whichever `brig` is on PATH, not the checkout.
- `check-claims.sh` checks that the named tests exist, but not that the quoted sentence is still in `security.md`.
- `docs/security.md:141`: `BRIG_ALLOW_DENIED` also accepts `true`, `yes` and `on`, so "only that" is not quite right.

PR #101, opened 2026-08-27
fix(runtime): tell a stopped sandbox from one that cannot be asked

### LINKED ISSUE #49: Distinguish a stopped sandbox from a runtime that failed to answer
### Problem

Both runtime adapters implement `Running` by shelling out to the runtime's `ps` and returning false when
the command returns an error. Any failure to ask the question is recorded as a definite "no".

So a runtime binary that is missing, a daemon that is not up, a socket that is not readable, a timeout, or
an output format brig does not expect all look identical to a sandbox that is stopped. The caller then
does what it does for a stopped sandbox and boots another one, or reports a sandbox as gone while it is
still running.

The distinction matters most in the case where it is hardest to notice: a partly working runtime.

### Proposed solution

Separate "not running" from "could not tell". Return an error from the probe and let callers decide.
Booting is the wrong response to an unanswered question, and reporting state to a user is the wrong place
to guess.

### Acceptance criteria

- A runtime that errors produces a message naming the failure, not a silent "not running".
- No verb boots a second sandbox because the first could not be probed.
- Tests cover a missing binary, a non-zero exit, and unparseable output.

pmoust:
Both adapters returned `false` from `Running()` on any exec error, so a missing binary, a permission error, or a broken runtime all read as "nothing is running" -- and `EnsureRunning` then booted a second sandbox over the first.

There are three answers, not two: running, not running, and could not tell. `Running` returns `(bool, error)` now, and a caller that cannot tell refuses rather than boots:

    cannot tell whether the sandbox brig-claude-code is already running, so
    refusing to start a second one over it: ...

Failing to answer is not the same as answering no, and booting on it is the dangerous direction: it is how two sandboxes end up on one workspace.

Every implementation, caller and test double updated, including brigd.

Closes #49

### DIFF
--- a/cmd/brigd/daemon_test.go
+++ b/cmd/brigd/daemon_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"net"
 	"os"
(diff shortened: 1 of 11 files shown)


ananos (approved):
Thanks @pmoust! Verified on our mac testbed beyond the unit suite: with a present-but-failing hull stub, rc16 proceeded toward booting a second sandbox over the held workspace, while this branch refuses and surfaces the runtime's own stderr ("cannot tell whether the sandbox ... is already running, so refusing to start a second one over it"). Exactly the right direction to fail in. LGTM.

PR #102, opened 2026-08-27
fix: refuse a bad hosts.yml and an over-long socket path

### LINKED ISSUE #74: Refuse a hosts.yml that is not a regular file instead of blocking on it
### Problem

When `github.user` is not set in git config, brig reads the login from gh's
`hosts.yml` under `$GH_CONFIG_DIR` or `~/.config/gh`, on every verb that loads a
profile. The read is a plain open. A FIFO at that path blocks it, and brig hangs
with no message.

The workspace reader already refuses a FIFO for exactly this reason, because
the guest can plant one there. `hosts.yml` sits outside the workspace, so the
guest cannot, and this is a robustness gap rather than a boundary one: anything
that replaces that file with something that does not return on open stops
every brig command until it is killed.

### Proposed solution

Open the file the way the workspace reader opens things: check the type before
reading, refuse anything that is not a regular file, and say what was found. The
same helper can serve both.

### Acceptance criteria

- A FIFO, a device or a directory at the `hosts.yml` path produces a one-line
  refusal naming the path and the type, not a hang.
- A test covers the FIFO case, which is the one found.

### LINKED ISSUE #80: Refuse a unix socket path over the platform limit instead of failing with invalid argument
### Problem

On the `hvi` hypervisor brig starts a user-mode network gateway and listens on
`$BRIG_GATEWAY_SOCK`, which defaults to `~/.brig/gateway.sock`. A unix socket
path on macOS is limited to 104 bytes, and on Linux to 108. When `HOME` is
long, which is ordinary for a temporary or per-test home, the bind fails with:

    error: failed to listen on /.../.brig/gateway.sock: listen unix ...: bind: invalid argument

and brig reports only that the gateway did not come up within ten seconds and
points at the log. `invalid argument` names neither the limit nor the path
length, so the user is left to guess.

Observed with a home directory under `/private/tmp/claude-501/...`, 120 bytes
into the socket path.

### Proposed solution

Check the length before listening, and refuse with a message that names the
path, its length and the limit, and suggests `BRIG_GATEWAY_SOCK` pointing at a
shorter path. The check belongs where the gateway socket path is chosen, so it
fires before the gateway is spawned and before the ten-second wait. The same
limit applies to `brigd`'s socket and to the `.qemu` twin of the gateway socket.

### Acceptance criteria

- A socket path over the platform limit is refused before anything is started,
  with a message naming the path, its length and the limit.
- A test covers a path one byte over the limit.

pmoust:
Two of the same shape: refuse a bad path with a sentence instead of hanging or failing with a kernel error.

**#74** -- the gh `hosts.yml` was opened with a plain `os.Open`, on every verb that loads a profile. A fifo there blocks brig forever. Now:

    brig: refusing to read .../hosts.yml: it is a named pipe (FIFO), not a regular
    file, and brig would block on it or fail later with a kernel error. Point
    GH_CONFIG_DIR at a directory holding a real hosts.yml, or remove what is there

**#80** -- a unix socket path over the platform limit failed with `bind: invalid argument`, which names nothing. It is refused before binding now, naming the limit and the length given. This one bites us today: `cmd/brigd/daemon_test.go` carries a `shortDir` helper purely to work around it, and I hit it by hand running brigd with a long `--socket`.

Verified both: the fifo case returns promptly rather than hanging, and the long socket path is refused before bind.

Closes #74
Closes #80

### DIFF
--- a/cmd/brigd/daemon_test.go
+++ b/cmd/brigd/daemon_test.go
@@ -90,7 +90,9 @@ func ask(t *testing.T, socket, request string) Response {
 // limit. t.TempDir puts the test's own name in the path, and on macOS that
 // lands the socket past 104 bytes, where bind fails with "invalid argument"
 // and the daemon looks broken for a reason that has nothing to do with it.
-// See #80 for the daemon-side check that should name this limit.
+//
+// chooseSocket now refuses such a path by name, but refusing it is not binding
+// it: a test that wants a listener still needs one that fits.
 func shortDir(t *testing.T) string {
 	t.Helper()
 	dir, err := os.MkdirTemp("", "brigd")
--- a/cmd/brigd/main.go
+++ b/cmd/brigd/main.go
@@ -111,8 +111,15 @@ type Session struct {
 }
 
 func main() {
-	socket := flag.String("socket", defaultSocket(), "unix socket to listen on")
+	flagSocket := flag.String("socket", "",
+		"unix socket to listen on (default $XDG_RUNTIME_DIR/brigd.sock, or ~/.brig/brigd.sock)")
 	flag.Parse()
+
+	socket, err := chooseSocket(*flagSocket)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "brigd: "+err.Error())
+		os.Exit(1)
+	}
 
 	// The daemon builds the registry from the same sources the CLI does. It
 	// looks profiles up by name, so without this it would know only the names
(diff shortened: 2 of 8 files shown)


ananos (approved):
Thanks @pmoust! Both legs verified live on the mac testbed: the fifo hosts.yml genuinely hung rc16 (killed it after 8s) while this branch refuses instantly naming the path and the type, and the socket check reports the platform limit (103 here), the actual length and which flag supplied it, where rc16 said only "bind: invalid argument". Nice that the missing-file fallback stays intact. LGTM.

Issue #30, opened 2026-08-25
Report the isolation boundary, and let a profile require a microVM


pmoust:
### Problem

Someone installs brig for a VM boundary. Several settings hand them something weaker, and no brig command
says so.

`BRIG_CONTAINERD_RUNTIME` replaces the urunc shim with whatever it names. Setting it to `runc` produces a
container sharing the host kernel, with exit 0 and no warning. The name of the setting does not appear
anywhere in the CLI package, so no help text, error or status line mentions it.

The nerdctl adapter falls back to `docker` when nerdctl is absent, with no message, while its `Kind` method
keeps returning `nerdctl`. `brig env` then reports the runtime as nerdctl and the binary as docker.

`brig env` names no containerd runtime, no hypervisor, no rootfs type, no verify mode and no network mode.
The settings that decide what boundary a user got are the ones brig does not report.

### Proposed solution

Two halves, and both are needed.

1. An unconditional status line, in `brig env` and in the pre-boot report from #10, naming the effective
   containerd runtime, the hypervisor, the rootfs type, the verify mode and the network mode. Positive and
   always printed. A warning fired only when the runtime is not urunc would be worse than nothing: it is a
   string comparison that certifies a boundary brig cannot observe, and it stays silent in the case nobody
   chose, where the shim is absent, shadowed or not registered.

2. An explicit contract: `--isolation microvm | container | auto`. `microvm` is the default and fails when
   it cannot be delivered. `container` is an opt-in to a weaker boundary. `auto` prints what it chose. A
   profile may require `microvm`, and no environment variable may weaken a profile's requirement.

### Acceptance criteria

- `brig env` names all five settings on every invocation, on both platforms.
- `BRIG_CONTAINERD_RUNTIME` naming a non-urunc runtime, under the default `--isolation microvm`, fails with
  a message naming the setting.
- `Kind` reports the binary actually being driven.
- A test asserts a profile requiring `microvm` cannot be downgraded by any setting.

### Additional context

#10 owns the report block. This owns the fields in it and the contract behind them.

PR #103, opened 2026-08-27
feat(envelope): report the isolation boundary of the sandbox

pmoust:
`Kind()` returned "nerdctl" while `isDocker()` in the same file already knew better, and `BRIG_CONTAINERD_RUNTIME` can point the container at an ordinary runtime -- a shared kernel rather than a microVM -- with nothing said. A reader could not tell from any output whether they got a VM or a container.

The envelope gains the row its own comment has been reserving:

    ISOLATION    microVM (hull, hvi backend)

It reports what is true rather than what was configured, covers hull, nerdctl, docker and a replaced containerd runtime, and says it cannot tell rather than guessing.

**Deliberately not in scope:** the `--isolation` contract that lets a profile *require* a microVM. That is the larger half of #30 and a separate PR. This is the reporting half only.

Refs #30

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,7 @@ your own.
 $ brig info claude
 PROFILE      claude-code
 SANDBOX      brig-claude-code (hull)
+ISOLATION    microVM (hull, vz backend)
 WORKSPACE    /Users/you/brig/claude-code (read-write)
 IMAGE        ghcr.io/brig-sh/claude-code-stock:latest (pull missing)
 CREDENTIALS  GH_TOKEN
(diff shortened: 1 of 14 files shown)


ananos (approved):
Thanks @pmoust! Verified on the mac testbed: `brig info claude` resolves and reports `ISOLATION microVM (hull, hvi backend)` -- the resolved-not-configured behavior is exactly right, and the runc warning case is the honest line the docs needed. Rebased over #101 (union of Isolation + the three-answer Running in the interface), full suite green. LGTM.

Issue #109, opened 2026-08-31
Key the session index by ref, and record the project beside the home


asapranidis:
Part of #47. Follows the ref-and-parser issue.

### Problem

`internal/wrap/index.go` already is the session index #47 asks for. It is `~/.brig/workspaces.json`,
it maps a sandbox name to the host directory that sandbox was started with, it writes through a
temporary file and renames, it treats a missing or corrupt file as an empty map rather than an
error, and `ForgetWorkspace` prunes an entry when the sandbox is removed. Its header comment
already reasons carefully about why it is an index and not a source of truth.

Two things about it are wrong under #47, and they are the key and the value.

**The key is ambiguous.** It is the sandbox name, and #47 rejects sandbox names as identifiers for
exactly this reason: a user agent called `claude-refactor` and a `refactor` session of `claude`
produce the same string, and a dash cannot express which reading is meant.

**The value cannot hold what is now needed.** It is a single path string. After #6 a session has
two directories with two different lifetimes -- a static guest home that *is* its identity, and a
per-run project mount that is not -- and one string cannot carry both.

### Proposed solution

`~/.brig/sessions.json`, keyed by the ref, with a value that carries the home, the project and the
sandbox name.

Everything else in `index.go` is reused as it stands: the atomic write, the tolerant read, the
0700 directory, pruning on removal, and the rule that the runtime stays authoritative about what
is actually mounted while this file is only a hint.

**No migration from `workspaces.json`.** Deriving a ref from an existing `brig-claude-refactor` key
is precisely the ambiguity above, so any migration would have to guess. The old file is deleted on
sight instead, and each existing session takes one restart -- which `index.go`'s own header already
names as the accepted cost of an absent entry, and which is safe because everything persistent
lives in the workspace on the host.

Pruning runs on every `brig ls`, dropping entries whose sandbox is gone.

### Acceptance criteria

- An entry survives a stop and is dropped by a remove.
- `brig ls` prunes an entry whose sandbox no longer exists.
- An explicit `--home` or `BRIG_WORKSPACE` still wins over the recorded value -- asking for a
  different directory is something a user is entitled to do, restart and all.
- A corrupt or unreadable file costs one restart and no command failure, as today.
- The old `workspaces.json` is removed, and its absence is not an error.

### Effect on the sandbox boundary

None. The file names host directories rather than reaching them, which is why `index.go` already
creates its directory 0700; that stays. It grants the guest nothing it did not have.

### Additional context

[...]

Issue #108, opened 2026-08-31
Parse a session ref, and give brig's flags three positions


asapranidis:
Part of #47. This is the spine of the CLI vNext work: no command line changes behaviour for a
user, and everything else under #47 consumes what this lands.

### Problem

Two defects with one cause.

**brig's flags have no position.** `split` (`cmd/brig/main.go:419`) decides the boundary from a
single piece of state -- have I seen the profile name yet -- so brig's own flags are legal on both
sides of it and the agent's are legal on one. The consequence is that `brig run claude --json`
hands `--json` to Claude Code, which does not have that flag. brig never sees it, cannot explain
the failure, and the exit status belongs to the agent. The same holds for `-q`, `--no-color`, `-y`,
`--mem`, `--cpus` and `-d`. #41 closed the half of this that mattered most -- an unknown flag left
of the profile is now a usage error rather than a swallowed operand -- but the one-dimensional
position model underneath it is unchanged.

**There is no session identifier.** A session is named with `--name` and reduced to a directory
slug by `session.Slug` (`internal/session/session.go`), which sanitises and then *truncates* at
`maxSlug = 10`. Truncation is why two different names can slug onto one sandbox, why
`EnsureRunning` needs `claimSlug` to refuse the collision, and why `main.go` has to print
`session %q uses %s (sandbox %s)` whenever the two differ. None of that is needed once the
identifier is a first-class thing that is refused when it does not fit.

### Proposed solution

**1. A ref type, in `internal/session`.** That package already owns `Slug`, `Resolve` and the
reserved-name refusal, so this belongs there rather than in a new package.

```go
type Ref struct{ Agent, Label string }   // Label == "" is the agent's default session
func ParseRef(string) (Ref, error)
func (Ref) String() string               // "claude" | "claude@refactor"
```

The rule that makes the rest cheap: **a label must already be slug-clean.** `ParseRef` refuses
anything `Slug` would have to rewrite instead of quietly rewriting it. That is #47's "refuse an
over-long label rather than truncating it", and it removes the collision class entirely -- so the
`session %q uses %s` warning survives only on the deprecated `--name` path, where truncation still
happens.

Exactly one `@`. `claude@` is an error, not a default. The reserved-name check
(`profile.Reserved`) moves onto the label. Label length caps at 20: `maxSlug` is 10 today only
because truncation had to avoid collisions, and refusing removes that pressure.

**2. Three positions, two closed sets.** `split` becomes a state machine keyed on *have I passed
[...]

Issue #13, opened 2026-08-25
Settle one CLI grammar for 0.2, and show it in the help text


pmoust:
### Problem

brig has three grammars at once, and a user cannot tell from a command line
which one applies.

- A plural noun is a command of its own: `brig profiles`, and the deprecated
  `brig agents`. Both do what `brig profile ls` does.
- Two noun commands are also spelled at the top level: `brig import` and
  `brig export` are `brig profile import` and `brig profile export`.
- One verb means two things. `brig create` starts a sandbox and does not
  attach. `brig secret create` writes a secret.
- The same operation has several spellings. `ls` and `list`, `export` and
  `save`, `import` and `load`, `delete` and `rm`. None of the second spellings
  are in the help text, so they are found by accident and then written into
  scripts.
- The verb sets differ per noun. A profile is managed with
  `ls|export|import|edit|rm`. A secret is managed with
  `create|read|update|delete|ls|import`. The same job has a different name
  under each noun.

The help text lists the commands but shows no complete command line. A reader
gets the shape `brig exec <profile> -- cmd [args...]` and has to assemble the
first real invocation.

### Proposed solution

Settle one grammar for 0.2, and print examples of it.

**1. Two forms, and no third.**

```
brig <verb> [session] [flags]      the lifecycle of a sandbox
brig <noun> <verb> [name] [flags]  everything you manage
```

**2. No plural commands.** `brig profiles` becomes `brig profile ls`.
`brig agents` is already deprecated and goes with it.

**3. No top-level spelling of a noun command.** `brig import` and
`brig export` become `brig profile import` and `brig profile export`.

**4. One verb set under every noun.**

| Verb | Meaning | Replaces |
|---|---|---|
| `ls` | list them | `list` |
| `show` | print one | `secret read` |
| `set` | write one, new or not | `secret create`, `secret update` |
| `edit` | open one in `$EDITOR` | |
| `rm` | remove one | `secret delete` |
| `import` | take one from the host | `load` |
| `export` | write one out | `save` |

**5. One spelling per operation.** The undocumented aliases go: `list`,
`save`, `load`, `delete`.

**6. `create` keeps one meaning.** `brig create <profile>` becomes
`brig run <profile> --detach`, and `brig start <session>` starts a session
that exists and is stopped. A verb that means "make a sandbox" and "write a
secret" in the same CLI has to lose one of the two.

**7. Show the grammar in `brig help`,** with lines a reader can paste:

```
[...]

Issue #47, opened 2026-08-25
CLI vNext for 0.2: one grammar, one session address, one isolation flag


pmoust:
Follows on from #13, which asks for one grammar to be settled for 0.2. This is a concrete proposal to
argue with, not a decision already taken.

### Why now

Migration cost is near zero today and permanent after the first stable release. Nothing installs brig the
way the pipeline intends, every tag is a prerelease, the cask is pinned by hand, and there is no stability
statement anywhere in the repository. No scripts exist to break. This is the last release where changing
the grammar is free.

### The three decisions to settle

**1. The taught first command is `brig run claude`.**

`brig claude` still parses, where verbs always win, but it is never taught and appears in no help example.
Making the verbless form canonical makes the tool's surface data-dependent: no verb can be added without
checking it against the agent registry, and no agent named without checking it against the verb list. git,
docker and kubectl all declined this deliberately.

`brig run claude .` is rejected, which is where this disagrees with #6. A positional directory reads as
"mount this here", which is Docker and devcontainer semantics. The workspace is the guest's *home*, so
pointing it at a repository puts the agent's dotfiles in that repository. Directory selection stays an
explicit `--in PATH` with a warning attached. **This one needs a decision, since #6 proposes the
opposite.**

**2. A session is `<agent>@<label>`, derived and never stored.**

`claude` is the default session of the `claude` agent; `claude@refactor` is a second one. That string is
what `ls` prints, what every verb takes, what completion offers and what appears in JSON. This is what #5
asks for, with a separator that can carry the boundary.

Rejected, each for a reason: profile plus `--name` is the defect being fixed, and takes a flag Claude Code
owns. A flat session name forbids `claude@refactor` and `codex@refactor` coexisting and needs a new
authoritative record. The sandbox name cannot work, because a user agent named `claude-code-refactor` and
a `refactor` session of `claude-code` produce the same string, and a dash cannot express which reading is
meant. Workspace identity makes two sessions over one directory unaddressable. A generated ID hands the
user a second unusable string.

One index file, `~/.brig/sessions.json`, maps ref to workspace path so a session started with `--in` shows
the right directory. It is an index, not a source of truth: the runtime stays authoritative and every `ls`
prunes entries whose sandbox is gone. It can be deleted rather than migrated once hull grows an `inspect`.

**3. brig's own flags typed right of the ref reach the agent, and brig should say so.**

`brig run claude --json` hands `--json` to Claude Code, which does not have that flag. brig never sees it,
[...]

PR #113, opened 2026-09-01
refactor(brigd): rename the protocol's vm field to sandbox

### LINKED ISSUE #112: Rename the daemon protocol's `vm` field to `sandbox`
Part of #47, and the one item on its checklist that depends on nothing else. Worth doing now.

### Problem

The daemon protocol calls a sandbox a VM. `cmd/brigd/main.go:101` is:

```go
VM        string `json:"vm"`
```

and `docs/brigd.md:56` and `:70` document it in the two example payloads. brig says "sandbox"
everywhere a user can see, and for a good reason -- it is a microVM on macOS and a container on
Linux, so "vm" is wrong on one of the two platforms.

The protocol has no client yet. That makes the rename free today and a compatibility problem after
0.2, which is the whole argument for doing it in this release.

### Proposed solution

Rename the field and its JSON tag to `sandbox`. Update `cmd/brigd` and its tests, and the two
payloads in `docs/brigd.md`.

While the file is open: the same payloads carry `"workspace"`, and #6 splits that concept into a
static guest home and a per-run project mount. Either rename it in this pass or record here why it
keeps the old name -- the point of doing this now is that no client exists to break, and that is
equally true of both fields.

### Acceptance criteria

- No `"vm"` in the protocol or in `docs/brigd.md`.
- `docs/brigd.md`'s examples match what the daemon actually emits.
- The `"workspace"` field is either renamed with it or explicitly left, with the reason written
  down.

None. A field name, on a protocol with no clients.

asapranidis:
## Summary

The brigd daemon protocol was the last place in brig still calling a sandbox a "vm". brig says
"sandbox" everywhere a user can see, because it is the only word true on both platforms -- a microVM
on macOS, a container on Linux -- so `"vm"` on the wire disagreed with the CLI, the docs and the
error messages about the very thing it was reporting.

The protocol has no client yet, which is the whole reason to do this now. Today it costs a
three-file diff; after 0.2 it is a compatibility break, and the field would keep the wrong name for
as long as anyone spoke the old protocol.

## Related issues

Closes #112. Part of #47, where this is the one sub-issue with no dependencies and so the first to
land.

## Changes

- `Session.VM` becomes `Session.Sandbox`, and its JSON tag `"vm"` becomes `"sandbox"`.
- The fill site in `daemon.remember` follows, as does `cmd/brigd/daemon_test.go` including its
  `byVM` map.
- Both example payloads in `docs/brigd.md` now match what the daemon actually emits.
- Two doc comments that named the *thing* rather than the technology now say "sandbox". Comments
  that genuinely mean the VM technology ("microVM") are left alone.
- `workspace` deliberately keeps its name. #6 will touch this struct to add a `project` field beside
  it, and renaming `workspace` is that change's business rather than a second break here -- the "no
  client yet" window does not close until 0.2 ships, so nothing is lost by waiting.
- `wrap.Config.VMName` is untouched: an internal Go identifier no client ever sees. That leaves
  `Sandbox: cfg.VMName` as a naming seam at the fill site, which is intentional and scoped out.

Two notes on the checklist above. **No new test was added** because the compiler enforces a field
rename; the existing daemon tests were updated instead, which is what the box is ticked for.
**`script/smoke.sh` does not pass, and does not pass at `8340bb4` either** -- the two failure sets
are identical (11 each, verified by diffing them), so nothing here regresses it. At least two of
those failures are the host environment rather than the tree: an imported credential that expired
11 days ago, and a `GH_TOKEN` holding an unresolved `op://` reference.

## Checklist

## LLM usage

Authored with assistance from Claude Code (Opus 5), which produced the rename, the commit message
and this description. Verification was re-run independently of the agent that made the change:
`make all` green, and the `script/smoke.sh` failure set diffed against the parent commit to confirm
it is unchanged. Accountability for every line remains the author's.

### DIFF
--- a/cmd/brigd/daemon_test.go
+++ b/cmd/brigd/daemon_test.go
@@ -512,24 +512,25 @@ func (l livenessRuntime) Running(string) (bool, error) { return l.running, l.err
 // The inventory re-reads liveness from the runtime on every report, so a
 // runtime that cannot be asked has to be reported as unanswered rather than as
 // a stopped sandbox. Folded into running:false it reads as a sandbox that
-// exited, and the operator goes looking for a VM that may well still be up.
+// exited, and the operator goes looking for a sandbox that may well still be
+// up.
 func TestStatusSaysWhenTheRuntimeCouldNotBeAsked(t *testing.T) {
 	d := newDaemon()
 	d.sessions["brig-broken"] = entry{
-		Session: Session{Agent: "claude-code", VM: "brig-broken"},
+		Session: Session{Agent: "claude-code", Sandbox: "brig-broken"},
 		rt:      livenessRuntime{err: errors.New("cannot connect to the daemon")},
 	}
 	d.sessions["brig-up"] = entry{
-		Session: Session{Agent: "claude-code", VM: "brig-up"},
+		Session: Session{Agent: "claude-code", Sandbox: "brig-up"},
 		rt:      livenessRuntime{running: true},
 	}
 
-	byVM := map[string]Session{}
+	bySandbox := map[string]Session{}
 	for _, s := range d.status() {
-		byVM[s.VM] = s
+		bySandbox[s.Sandbox] = s
 	}
 
-	broken := byVM["brig-broken"]
+	broken := bySandbox["brig-broken"]
 	if broken.RunningError == "" {
 		t.Error("a runtime that could not be asked was reported as a plain stopped sandbox")
 	}
@@ -541,7 +542,7 @@ func TestStatusSaysWhenTheRuntimeCouldNotBeAsked(t *testing.T) {
 	}
 
 	// The answers that were given are unchanged.
-	if up := byVM["brig-up"]; !up.Running || up.RunningError != "" {
+	if up := bySandbox["brig-up"]; !up.Running || up.RunningError != "" {
 		t.Errorf("a running sandbox was misreported: %+v", up)
 	}
 }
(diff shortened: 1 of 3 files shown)


asapranidis:
**#115 is stacked on this branch.** Its base is `refactor/112-brigd-sandbox-field` rather than
`main`, so its diff shows only its own five files, and GitHub retargets it to `main`
automatically when this merges.

Merge order is this one, then #115. Both are green.

Stack, in order:

1. **#113** (this) -- `refactor(brigd)`: rename the protocol's `vm` field to `sandbox`. Closes #112.
2. **#115** -- `feat(cli)`: give brig's flags a position and read a session ref. Closes #108.

Both are sub-issues of #47, taken in the order recorded there.

asapranidis:
**#117 is stacked on this branch**, so the stack is now three deep. Merge bottom-up.

Each PR's base is the branch below it, so every diff shows only its own change and GitHub retargets
each one as the one below merges. All three are green.

asapranidis:
green.

ananos (approved):
Went through this one in full and found nothing to change.

The rename is complete: no `"vm"` is left on the wire or in the docs, and what
does still say "vm" is test fixture names and `wrap.Config.VMName`, which is
what the description says it is. `go build`, `go vet` and the tests are green.

The argument for doing this now rather than after 0.2 is the right one. This is
also the bottom of the stack, so landing it takes the rebase pressure off the
other six.

Issue #114, opened 2026-09-01
Raise the session label cap to 20, and warn where it moves a guest home


asapranidis:
Follows #108, which made the session label a first-class thing that is refused rather than
truncated. Part of #47.

### Problem

A session label is capped at `maxSlug`, which is 10 characters
(`internal/session/session.go:19`). That is too tight for the thing it names. `experiment` fits
exactly; `refactoring`, `credentials`, `integration` and `benchmarking` do not.

The cap is 10 for a reason that no longer fully applies. Its doc comment says it "keeps directory
names short. It is a tight budget, so two long names can slug the same and share one sandbox" -- the
budget was set low because `Slug` *truncates*, and truncation is what makes two different names
collide onto one sandbox. #108 removed that hazard for refs by refusing an over-long label instead
of rewriting it, so for a ref the cap is no longer protecting against a collision. It is only
protecting the length of a directory name.

And that is the argument for raising it: **the slug becomes the directory name of brig's guest
home**, and the suffix of the sandbox name (`vmName += "-" + slug`, `internal/wrap/config.go:217`).
A guest home directory can comfortably be longer than ten characters. The budget should fit a label
a person would actually choose rather than a path-safety minimum.

### Proposed solution

`const maxSlug = 20`.

`ParseRef` needs no change -- it caps at `maxSlug` through the constant, so it picks the new value
up for free, and #108's tests drive off the constant rather than the literal.

Rewrite the constant's doc comment so it says two things rather than one, because the split is the
interesting part after #108:

- For a **ref**, the collide-by-truncation hazard is gone: `ParseRef` refuses a label over the cap
  instead of truncating it.
- For **`--name`**, it survives: `Slug` still truncates, so two long `--name` values can still slug
  the same, and `claimSlug` in `EnsureRunning` is still what catches it.

Do not write it as though the hazard were gone everywhere.

### The migration, which is the actual work

Raising the cap changes the derived slug for any existing session whose `--name` is 11 to 20
characters. That means a new sandbox name and a new guest home directory: the old ones are orphaned,
and the guest's in-memory state is lost on the next run. The work in the workspace survives, because
it is on the host -- an in-sandbox login does not.

So this needs a warning, and it needs to be precise enough that nobody unaffected sees it. Fire only
when **both** hold:

1. The slug the old cap would have produced -- the name truncated at 10, with `Slug`'s existing
   trailing `-`/`.` trim reapplied -- differs from the new slug, and
[...]

Issue #111, opened 2026-08-31
Rewrite the README and docs around the new nouns


asapranidis:
Part of #47. Lands last, once the grammar and the workspace model have settled.

### Problem

Every place brig teaches its own use will be wrong after #47: the `usage` const in
`cmd/brig/main.go:38`, `README.md`, and `docs/`.

The grammar PRs each fix `usage` for their own changes, because it is a `const` in the file they
edit and cannot not change. Nothing makes them fix the prose, and prose is where the harder problem
is: **the CLI-only rename means two words are correct at once.** `brig agent ls` is the command,
and `~/.brig/profiles/`, `BRIG_PROFILE_DIR` and the `profile` file format keep their names. Docs
that blur those two are worse than docs that are simply out of date, because a reader cannot tell
which word they are supposed to type. `docs/profiles.md` is the file where that lands hardest.

This also carries the clause dropped from #13's title -- showing the settled grammar, not just
listing the commands.

### Proposed solution

- README rewritten around the new nouns, with the taught first command being `brig run claude`. The
  verbless `brig claude` appears in no example, per #47: teaching it would make the surface
  data-dependent.
- Help text led by the boundary -- what the guest can see and what it holds -- rather than by the
  command list.
- Complete command lines, not shapes. #13's complaint is that a reader gets
  `brig exec <profile> -- cmd [args...]` and has to assemble the first real invocation themselves.
- An old-to-new table and the deprecation window written down where a user will find it: both
  grammars work in 0.2, the old spellings go in 0.3, `run` is never removed.
- `docs/quickstart.md`, `docs/profiles.md` and the rest reconciled to the split vocabulary, saying
  explicitly which word is the command and which is the file format.
- The stability statement #47 notes is missing from the repository entirely.

### Acceptance criteria

- No example anywhere in `README.md` or `docs/` uses a retired spelling.
- The deprecation window is documented once, in a place a user reaches from the README.
- The `agent` (command) and `profile` (file format, directory, env var) split is stated where a
  reader meets it, not left to be inferred.

### Additional context

Behind #5, #6 and #13. The retired-spelling check is worth automating rather than reviewing by eye,
so it does not rot the first time a doc is edited after 0.3.

asapranidis:
**Two README lines are stale as of #115, which is worth knowing when this is picked up.**

`README.md:144` and `:146` still document the short flag spellings:

| `-t, --image IMAGE` | guest image to boot |
| `-m, --memory MB` | guest memory |

[...]

Issue #110, opened 2026-08-31
Run the agent as a child under `--json` instead of replacing the process


asapranidis:
Part of #47, where this is one line on the checklist. It is filed separately because it is not one
line of work, and because it changes behaviour for runs that never pass `--json`.

### Problem

`brig run` ends by replacing brig's own process with the agent: `hull.Replace`
(`internal/runtime/hull.go:492`) execs the runtime over brig. That is the right default -- the agent
owns the terminal, signals reach it directly, and its exit status is the command's exit status with
nothing in the middle.

It also means brig can never report its own failure in a machine-readable way on a run line. Once
the process is replaced there is no brig left to print anything, so a script cannot tell "brig
refused to start this" from "the agent ran and exited 1". #47 asks for a real `--json` on the left
of the line for exactly this reason.

### Proposed solution

Under `--json`, run the agent as a child instead of replacing the process: forward signals to it,
propagate its exit status per the table in #8, and emit one object that distinguishes a brig
failure from an agent failure. Without `--json`, keep `Replace` and today's behaviour exactly.

### Why this is its own issue

Running the agent as a child changes TTY ownership and signal forwarding, and getting it wrong is
not a JSON bug -- it is an interactive agent that stops responding to Ctrl-C, or one that cannot
tell it has a terminal. The blast radius is every interactive run, not the flag. Two code paths for
the same thing is also a real cost, and the alternative of moving *every* run to a child process is
worth considering on its merits rather than as a side effect of adding a flag.

### Alternatives considered

- **Always run as a child, one path.** Simpler to maintain and to reason about, and the exit status
  and signal handling get one implementation instead of two. It changes the signal and TTY
  behaviour of every interactive run, which is the thing users would notice first if it were wrong.
- **Leave it.** Costs nothing, and leaves a scripted `brig run` unable to distinguish a refusal
  from a failure, which is the gap #8's exit codes were supposed to close and cannot on this path.

### Acceptance criteria

- `brig --json run claude` reports a brig-side refusal as a brig failure with brig's own exit code.
- The same line reports a non-zero agent exit as the agent's, distinguishably, with the agent's
  status propagated.
- Ctrl-C reaches the agent, and an agent that requires a TTY still gets one.
- Without `--json`, the process is still replaced -- verified, not assumed.

### Effect on the sandbox boundary

Worth stating plainly: on this path brig stays in the middle of a stream it used to hand over
entirely. Nothing new is mounted and no new variable is forwarded, but brig is now reading and
[...]

PR #115, opened 2026-09-01
feat(cli): give brig's flags a position and read a session ref

### LINKED ISSUE #108: Parse a session ref, and give brig's flags three positions

asapranidis:
## Summary

Two defects with one cause, and this is the spine the rest of #47 consumes.

**brig's flags had no position.** `split` decided where brig's arguments ended from a single
piece of state -- have I seen the profile name yet -- so brig's own flags were legal on both
sides of it, and there was nowhere a global flag could stand at all. `brig --json run claude`
reported a command that does not exist, and `brig run claude --json` handed `--json` to an agent
that does not have it.

**A session had no identifier.** `--name` goes through `Slug`, which sanitises the name and then
truncates it. Truncation is why two different names can slug onto one sandbox, why
`EnsureRunning` needs `claimSlug` to refuse the collision, and why a run has to print which
directory it actually got.

Nothing is renamed here and no command line that works today changes behaviour.

## Related issues

Fixes #108. Part of #47, where this is the spine eight other sub-issues sit behind.

**Stacked on #113** -- the base is `refactor/112-brigd-sandbox-field`, so GitHub retargets this
to `main` once #113 merges. Review #113 first.

Deliberately excluded: #114, the label cap raise. It was written, then cut from this branch and
preserved on `feat/114-label-cap-20`, because it re-derives the sandbox name and guest home of
every existing long-named session and needs a migration warning of its own rather than riding
along inside a parser change.

## Changes

- `session.Ref` and `ParseRef`, in the package that already owns `Slug`, `Resolve` and the
  reserved-name refusal. A label must already be what `Slug` would make of it: the length check
  runs first, and below the cap `Slug` cannot truncate, so a single `label != Slug(label)` is
  then an exact test for "not slug-clean" -- the character rules stay in `Slug` and are not
  restated, so the two cannot drift.
- The cap is `maxSlug`, referenced through the constant. A longer cap would let two labels
  differing only past that character slug the same and share one sandbox and one guest home,
  which is the collision the refusal exists to remove.
- `brigFlags` carries the position a flag is legal in, and `ours` answers "mine, and legal
  here?". The zero value is no position, so a new entry has to say which one it means -- both
  wrong defaults are silent.
- The global position, left of the verb, is established **closed and empty**. A flag written
  there is a usage error naming the token; #24, #11, #30 and #110 are what put flags in it.
- `split` is keyed on whether the ref has been passed and returns a `session.Ref`. The label
  lands on the same path `--name` lands on, so `brig run claude@refactor` is
  `brig run claude --name refactor`. Both spellings on one line ask for two sessions at once and
  are refused by name.
- Right of the ref brig still answers to its own flags, because `brig run claude -q` is a line
  that works today. Once an unknown token has ended brig's reading, a brig flag further along is
  the agent's -- and brig now says so **without taking the token**, so every line that works
  today keeps working. Nothing is said after an explicit `--`.
- `--mem` joins `--memory`, both accepted. `-t` and `-m` keep working and print the one-line
  notice naming `--image` and `--mem`, which is the window #47 sets: v0.2 ships both grammars,
  v0.3 removes the short forms. `-n` and `-w` are the other two and retire with #5 and #6.
- The `usage` text follows: `--image` and `--mem` replace the short spellings, and the `--name`
  entry names the `<profile>@<label>` form.

**README is knowingly left stale by this PR.** Lines 144 and 146 still document
`-t, --image` and `-m, --memory`, and the `@label` form appears nowhere in it. The full docs
rewrite is #111; flagging it here so a reviewer is not misled into thinking the table is current.

## Manual testing

Every parsing path can be exercised without booting a sandbox, because parsing happens before
brig looks for a runtime -- so `runtime unavailable: unknown BRIG_RUNTIME` is a pass:

```

brig run claude@                      # refuses: ends with "@" and names no session
brig run @refactor                    # refuses: names no agent
brig run claude@a@b                   # refuses: more than one "@"
brig run claude@Refactor              # refuses: would have to become "refactor"
brig run claude@my-really-long-label  # refuses: 20 characters, caps at 10
brig run claude@desktop               # refuses: the claude-desktop profile owns it
brig run claude@one --name two        # refuses: two spellings, one session

brig --json run claude                # refuses: unknown flag before the command
brig run --mem 2048 claude            # accepts; --memory still accepted too
brig run -t img claude                # warns: `-t` is now `--image`

brig run claude -q                    # silent: still brig's own flag
brig run claude -p hi --quiet         # warns --quiet is the agent's, forwards it anyway
brig run claude -- --quiet            # silent: the line already said the rest is the agent's
```

The check worth making deliberately: in every warned case the agent must still **receive** the
token. A warning that quietly captured it would break the lines this design exists to protect.

## Checklist

- [ ] `script/smoke.sh` passes
- [ ] I have updated the affected docs (README, `docs/security.md`, `docs/profiles.md`, `docs/brigd.md`)

Three of those need saying rather than ticking:

**Real runtime**: not booted. This does touch the run path, so the box applies and is honestly
unticked -- every check above was made at the parse level with the runtime stubbed out. A boot
pass under a ref (`brig run -d claude@shake`, `brig ls`, `brig stop`, `brig rm`) is still worth
doing before merge.

**Docs**: `usage` in `cmd/brig/main.go` is updated; README is not, as described above.

This narrows the boundary rather than widening it, and that is most of the point. `ParseRef`
refuses a label that `Slug` would have to rewrite, which removes the case where two different
labels shorten onto one sandbox name and one guest home -- two sessions sharing one directory.
`claimSlug` still catches it for `--name`, which keeps its lenient behaviour; for a ref the case
can no longer arise. No new mount, no new forwarded variable, no new host path the guest can
reach.

## LLM usage

messages and this description. Verification was re-run independently of the agent that wrote the
change, from a frozen tree: `gofmt`, `make all`, a forced uncached `-race` run over
`cmd/brig`, `internal/session` and `internal/wrap`, and the smoke comparison against `85dee40`.
Every expected output in the Manual testing section was captured from the built binary rather
than read off the source. Accountability for every line remains the author's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,15 +135,24 @@ separate Linux re-implementation of it.
 for one release, each printing a one-line note naming the new one; they no
 longer appear above. There is deliberately no `brig template edit`.
 
-Flags go before the agent's own arguments; `--` ends brig's parsing outright,
-so an agent flag spelled like one of brig's still reaches the agent.
+A brig line has three places a token can stand, and they are not the same kind
+of place. Left of the verb are brig's global flags, a closed set -- an unknown
+flag there is a usage error naming the token, rather than an operand quietly
+swallowed. Between the verb and the agent are the run-line flags below. Right of
+the agent the vocabulary is the agent's, and `--` ends brig's parsing outright,
+so an agent flag spelled like one of brig's still reaches it.
+
+brig does still read its own flags after the agent, because `brig run claude -q`
+is a line that works. But once an unknown token has ended brig's reading, a brig
+flag further along belongs to the agent -- and brig now says so, without taking
+the token, so a line that works today keeps working.
 
 | flag | what it does |
 | --- | --- |
-| `-n, --name NAME` | a session of its own: own workspace, own sandbox |
-| `-t, --image IMAGE` | guest image to boot |
+| `-n, --name NAME` | a session of its own: own workspace, own sandbox. Also written `<agent>@<label>` |
+| `--image IMAGE` | guest image to boot (`-t` still works, with a note) |
 | `-w, --workspace PATH` | host directory to mount as the guest home |
-| `-m, --memory MB` | guest memory |
+| `--mem MB` | guest memory (`--memory` also works; `-m` works with a note) |
 | `--cpus N` | guest vCPUs |
 | `-d, --detach` | with `run`: start the sandbox, print its name, exit |
 | `--skills` | seed your own `~/.claude` skills and plugins into the workspace |
@@ -182,7 +191,8 @@ brig shell codex 'ls -la ~'                # one command, in a login shell
 ### Named sessions
 
 ```bash
-brig run claude --name refactor
+brig run claude@refactor
+brig run claude --name refactor    # the same session, the older spelling
 ```
 
 A session of its own: its own workspace (`~/brig/claude-code-refactor`), its
@@ -193,15 +203,22 @@ underscore, ten characters -- so `my_project` and `my-project` stay separate,
 while `Foo` and `foo` do not (macOS filesystems ignore case, and two names
 sharing one directory but not one VM is a bug waiting to happen). If the slug
 differs from what you typed, brig says which directory it used.
-`brig info claude --name refactor` prints the one in use.
+`brig info claude@refactor` prints the one in use.
 
-**Every other verb needs the same `--name`.** A named session is addressed by
-its agent *plus* its name, not by the name alone:
+The `@` form is the stricter of the two: a label brig would have to rewrite --
+uppercase, a space, more than ten characters -- is refused rather than quietly
+shortened, and the message names what it would have become. That is what makes
+the label safe to use as an address: nothing is silently mapped onto a directory
+you did not ask for. `--name` keeps its older, lenient behaviour.
+
+**Every verb takes the ref.** A named session is addressed as
+`<agent>@<label>`, and the older agent-plus-`--name` spelling still works
+everywhere it did:
 
 ```bash
-brig exec claude --name refactor -- git status
-brig stop claude --name refactor
-brig rm   claude --name refactor      # removes the sandbox, keeps the workspace
+brig exec claude@refactor -- git status
+brig stop claude@refactor
+brig rm   claude@refactor      # removes the sandbox, keeps the workspace
 ```
 
 `brig ls` shows the session as `brig-claude-code-refactor`, but that string is
--- a/cmd/brig/deprecated_test.go
+++ b/cmd/brig/deprecated_test.go
@@ -94,3 +94,59 @@ func TestBuiltInProfilesDoNotWarnAboutHostCredential(t *testing.T) {
 		}
 	}
 }
+
+// -t and -m keep working for one release and say what replaces them. They are
+// two of the four short flags #47 removes in v0.3, and the notice is the same
+// one the retiring verbs print: a working command and a line about it, not a
+// failure.
+func TestDeprecatedShortFlagsStillWork(t *testing.T) {
+	for _, c := range []struct {
+		args        []string
+		replacement string
+		check       func(options) bool
+	}{
+		{[]string{"claude", "-t", "img:1"}, "--image", func(o options) bool { return o.load.Image == "img:1" }},
+		{[]string{"claude", "-m", "2048"}, "--mem", func(o options) bool { return o.load.Mem == 2048 }},
+		// The inline form is the same flag, so it hears the same notice.
+		{[]string{"claude", "-t=img:1"}, "--image", func(o options) bool { return o.load.Image == "img:1" }},
+	} {
+		var (
+			o   options
+			err error
+		)
+		warning := captureStderr(t, func() { o, _, _, err = parse(c.args) })
+		if err != nil {
+			t.Errorf("parse(%q): %v", c.args, err)
+			continue
+		}
+		if !c.check(o) {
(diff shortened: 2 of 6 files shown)


asapranidis:
**#117 is stacked on this branch** — base `feat/109-session-index-by-ref` → `feat/108-session-ref-flag-positions`. It closes #109 and will retarget to `main` as this one merges. Merge order: #113, then this, then #117.

asapranidis:
green.

ananos (approved):
Reviewed in full. The flag positions and the ref parser hold up: I ran every
case in the Manual testing section against the built binary and all thirteen
behave exactly as written, including the ones that matter most, where a warned
token still reaches the agent.

The `position` type with no usable zero value, plus `TestBrigFlagsDeclareAPosition`
to enforce it, is the right shape for a boundary this easy to get wrong
silently. Warning about brig's own flags in the tail without taking the token is
also the right call.

Three small things are filed as follow-ups rather than held against this PR,
none blocking: #123 (the tail warning fires on verbs that forward no tail),
#124 (the overlap test covers claude-code only, brig ships eight profiles), and
#125 (the reserved-label refusal is not scoped to the agent that could
collide, so `codex@desktop` is refused for no reason).

One correction to the description: it says the README is knowingly left stale,
and that lines 144 and 146 still document `-t, --image` and `-m, --memory`. The
head commit 4f2ed2f rewrites exactly those lines and teaches the `@label` form,
so that paragraph and the unticked docs box are both a commit out of date. Worth
fixing so the next reader is not sent looking for stale docs that are not there.

PR #117, opened 2026-09-01
feat(wrap): key the session index by ref

### LINKED ISSUE #109: Key the session index by ref, and record the project beside the home

asapranidis:
## Summary

The session index was keyed by sandbox name, and that key cannot be read back. A sandbox is
`brig-<profile>-<slug>`, so a profile named `claude-code-refactor` and a `refactor` session of
`claude-code` produce the same string, and no rule recovers which reading was meant. Its value was
a single path, which was enough when a session had one directory.

Both change here. The index becomes `~/.brig/sessions.json`, keyed by the ref `#108` introduced and
carrying a typed value.

The key is built from the **resolved** profile name and the slug, which is what makes it stable:
`--name` accepts a name `Slug` then sanitises, so the name as typed is not an identifier, and
`claude` and `claude-code` are one profile through an alias. Keying on the resolution means a
session created under one spelling is found under the other, rather than becoming two entries and
two homes that drift apart.

## Related issues

Fixes #109. Part of #47.

**Stacked on #115** — the base is `feat/108-session-ref-flag-positions`, not `main`, so the diff
shows only this change and GitHub retargets it once #115 lands. Review order is #113, #115, then
this.

Scope narrowed from the issue title: the `project` field moved to #6, which introduces the project
mount and so can pick the field's shape against the real thing instead of guessing. Nothing here
records a project.

## Changes

- `~/.brig/sessions.json` keyed by ref, valued `{home, sandbox}`, replacing the sandbox-keyed
  `workspaces.json`. Every reader moves with it: the remembered-workspace path in `config.go`, the
  two `rememberSession` calls and the removal in `run.go`, and `workspaceOf` plus reset in
  `cmd/brig`.
- `readIndex`/`writeIndex` are generic over the value type rather than duplicated per file. The
  atomic write and the tolerant read are the whole reason these files behave alike; two copies
  would be two chances for one of them to stop being atomic or tolerant.
- Two read directions, because the callers hold different things: a ref-keyed lookup for the run
  path, and a scan by sandbox name for `brig ls`. `ls` deliberately does not re-derive a ref from a
  sandbox name — that is the ambiguity this change removes.
- `PruneSessions` drops entries whose sandbox the runtime no longer lists, on every `brig ls`.
- **`workspaces.json` is deleted rather than migrated**, for the ambiguity above. One restart per
  session in it, which is what an absent entry has always cost, and safe because everything
  persistent lives in the workspace on the host.
- **The slug-claim index keeps its data across a filename change.** It held the name
  `sessions.json`; the session index has taken that, so claims move to `slug-claims.json`. Unlike
  the workspace index this migration is unambiguous — same key, same value, only the filename
  differs — so `migrateSlugClaims` moves them instead of dropping them. Dropping them would turn
  off the refusal that stops two names shortening onto one sandbox, until every session had run
  again.

Three details in that migration worth a reviewer's eye:

1. It proceeds only on `os.IsNotExist` for `slug-claims.json`. A stat that fails another way has
   not established absence, and migrating over a claim index that merely could not be read would be
   the one way to lose a claim here.
2. It writes `slug-claims.json` before removing `sessions.json`, so a failed write leaves the
   claims where they are for the next command to retry.
3. It can only fire on an old file. It reads through `readIndex[string]`, and a current
   `sessions.json` has object values, so it does not parse as a map of strings and the migration
   is skipped. That is the safety property, and it has its own test.

The read that triggers it sits **above** `claimSlug`'s early return for an unnamed run. An unnamed
run claims nothing, but its `rememberSession` is the writer that would replace the old-format file
and take the claims with it, so it has to carry them across too.

## Testing

`script/smoke.sh` gained 13 checks across this change, so most of what a reviewer would do by hand
is now automated:

- the session is filed under its ref; an unlabelled session under the bare agent name; a session
  created as `claude@label` under that ref
- the alias finds the workspace the session was created with
- `rm` and `reset` drop entries; a corrupt index costs a restart and not a failure
- the old sandbox-keyed index is deleted on sight
- the old claims are carried to `slug-claims.json`, and a claim written under the old file name
  still refuses a colliding name
- an unnamed run carries the old claims across, and claims nothing itself
- a current session index is not mistaken for claims, and no claims are invented out of session
  entries

The last two pairs were mutation-tested. Moving `readSlugClaimIndex()` back below the unnamed-run
early return — the plausible tidy-up that would reintroduce the bug — turns
`an unnamed run dropped the old claims` red while `an unnamed run claims nothing` stays green, so
the pair localises the break rather than both failing at once. Weakening the guard's strict read
turns the two guard checks red and names the file it read.

`script/smoke.sh` also has failures that predate this branch and depend on host state. The sets are
byte-identical on this branch and on its parent `930e948`, verified by diffing them, so nothing
here regressed it.

## Manual testing

For anyone who would rather see it than trust the checks. It needs no microVM, no registry and no
credentials: `script/smoke.sh` drives brig against a stub `hull`, and the same stub works by hand.
From a checkout of this branch:

```bash
make build        # builds ./brig

REPO=$(git rev-parse --show-toplevel)
LAB=$(mktemp -d)
a=$(grep -n 'hull" <<' "$REPO/script/smoke.sh" | head -1 | cut -d: -f1)
b=$(awk -v s="$a" 'NR>s && $0=="STUB"{print NR; exit}' "$REPO/script/smoke.sh")
sed -n "$((a+1)),$((b-1))p" "$REPO/script/smoke.sh" > "$LAB/hull"
chmod +x "$LAB/hull"
mkdir -p "$LAB/assets" "$LAB/state" "$LAB/profiles" "$LAB/ws"
for f in Image bzImage container-initrd; do printf 'stand-in\n' > "$LAB/assets/$f"; done
export STUB_LOG="$LAB/argv.log"       STUB_STATE="$LAB/instance"
export BRIG_RUNTIME=hull              BRIG_RUNTIME_BIN="$LAB/hull"
export BRIG_WORKSPACE="$LAB/ws"       BRIG_READY_TIMEOUT=5
export BRIG_STATE_DIR="$LAB/state"    BRIG_VERIFY=off
export BRIG_BOOT_ASSETS="$LAB/assets" BRIG_HYPERVISOR=vz
export BRIG_PROFILE_DIR="$LAB/profiles"
brig()  { "$REPO/brig" "$@"; }
state() { for f in "$BRIG_STATE_DIR"/*.json; do [ -e "$f" ] && \
          { echo "== $(basename "$f")"; cat "$f"; echo; }; done; }
```

Two things to know before reading any output:

- **`brig run` and `brig create` exit 1 in this lab even when they worked.** The stub cannot fake an
  ephemeral mount, so brig's own credential-safety check refuses. That is brig being right about a
  fake runtime. Judge these checks by the contents of the state files, not by the exit status —
  except the collision refusal, where the exit status is the point and is called out.
- Re-run the block above for a fresh lab between checks. Several of these want an empty state
  directory, and the stub tracks a single instance, so a second session in one lab makes the reuse
  check read as a failure that isn't one.

**The index, and the shape of a key**

```bash
brig create claude -d          ; state    # keyed "claude-code"          — bare agent, no "@"
brig create claude@refactor -d ; state    # keyed "claude-code@refactor"
brig create claude --name rc23 -d ; state # keyed "claude-code@rc23"     — same as the @ form
```

Note the key is `claude-code`, the resolved profile, not the `claude` you typed.

**The alias, in a fresh lab with one session**

```bash
brig create claude --name rc23 -d
: > "$STUB_LOG"
brig exec claude-code --name rc23 -- uname -a
grep -c '^argv: run ' "$STUB_LOG"    # must be 0: the sandbox was reused, not rebooted
: > "$STUB_LOG"
brig exec claude-code@rc23 -- uname -a
grep -c '^argv: run ' "$STUB_LOG"    # must be 0 as well
```

**Removal prunes**

```bash
brig rm claude@rc23 ; state          # sessions.json is {}
```

**Upgrading: the old workspace index goes**

```bash
printf '{"brig-claude-code-rc23":"/tmp/old"}' > "$BRIG_STATE_DIR/workspaces.json"
brig create claude --name rc23 -d
ls "$BRIG_STATE_DIR"                 # workspaces.json is gone
```

**Upgrading: the old claims are kept, and still refuse**

```bash
printf '{"brig-claude-code-refactorin":"refactoring-api"}' > "$BRIG_STATE_DIR/sessions.json"
brig create claude --name other -d
cat "$BRIG_STATE_DIR/slug-claims.json"   # holds refactoring-api's claim alongside other's

brig create claude --name refactoring-web -d ; echo "exit=$?"
```

That last command must exit 1 and say `refactoring-web` shortens onto the sandbox
`refactoring-api` already uses. The refusal arrives after the execution envelope has printed,
because the claim is checked as the sandbox comes up — so do not truncate the output looking for
it.

An unnamed run carries the claims across too, and claims nothing of its own: seed the same file,
run `brig run claude -d`, and `slug-claims.json` holds only the seeded claim. In a clean lab an
unnamed run produces no `slug-claims.json` at all.

**The guard: a current index must not be mistaken for claims**

```bash
brig create claude --name keep -d           # a real entry exists
rm -f "$BRIG_STATE_DIR/slug-claims.json"    # after the session exists, not before
brig rm claude --name keep
cat "$BRIG_STATE_DIR/sessions.json"         # unchanged
ls "$BRIG_STATE_DIR"                        # no slug-claims.json invented
```

Clear the claim file *after* creating the session. A named run legitimately claims its own slug, so
clearing first lets that claim reappear and the check reads as a failure that isn't one.

**An unusable index costs a restart, not a failure**

```bash
printf '{not json at all' > "$BRIG_STATE_DIR/sessions.json"
brig run claude -d
cat "$BRIG_STATE_DIR/sessions.json"         # rewritten, well-formed
```

## Checklist

Two of those need saying rather than ticking. **`script/smoke.sh`** has failures that predate the
branch and depend on host state; the sets are identical on this branch and on `930e948`, and this
branch adds 13 passing checks. **Real runtime**: not booted. This touches the run path, so the box
applies and is honestly unticked — everything above was verified against the stub runtime, which
exercises the whole code path except the VM. A boot pass under a ref is worth doing before merge.

Docs are ticked because no user-facing behaviour or vocabulary changes here: the file this touches
is internal bookkeeping under `~/.brig`, documented nowhere, and `brig ls` prints exactly what it
printed before.

This affects the second of brig's two promises — that a sandbox gets the credentials it was named
and no others — and it holds it steady rather than widening it.

The concrete risk in this area is two sessions landing on one guest home, where each was handed
different credentials and whichever booted last leaves its own behind. `claimSlug` is what refuses
that, and the claim index is what it reads. Renaming that file without carrying its contents would
have switched the refusal off until every session had run again, so the migration exists for that
reason and is tested from both directions: the carried claim still refuses, and the migration
cannot fire on a current index.

No new mount, no new forwarded variable, no new host path the guest can reach. The index records
host directory paths and does not grant access to them.

## LLM usage

message and this description. Verification was re-run independently of the agent that wrote the
change, from a frozen tree: `gofmt`, `make all`, an uncached `-race` pass over `cmd/brig`,
`internal/wrap` and `internal/session`, and the smoke failure sets diffed against `930e948`. The
mutation test described under Testing was reproduced independently in a throwaway worktree rather
than taken on trust, and every command in the Manual testing section was run before being written
down. Accountability for every line remains the author's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,10 +232,16 @@ included, leaving all workspaces alone.
 
 **`-w` is remembered.** A session created with `--workspace` keeps that
 directory for every later verb, so `brig exec claude --name refactor -- ls`
-finds it without repeating the flag. brig records the path a sandbox was
-started with under `~/.brig`, and reads it back when neither `--workspace` nor
-`BRIG_WORKSPACE` names one; `brig rm` and `brig reset` drop the entry with the
-sandbox. Pass either of them a directory that is not the one the sandbox is
+finds it without repeating the flag. brig records it in
+`~/.brig/sessions.json`, filed under the session's ref, and reads it back when
+neither `--workspace` nor `BRIG_WORKSPACE` names one; `brig rm` and `brig reset`
+drop the entry with the sandbox, and `brig ls` drops any whose sandbox has gone.
+
+That file is an index and not a record of the truth: the runtime stays the
+authority on what is actually mounted, and an unreadable index costs one restart
+rather than failing the command. Because it is filed under the ref, a session is
+found under either spelling of its agent -- `claude` and `claude-code` are one
+agent through an alias, and both reach the one entry. Pass either of them a directory that is not the one the sandbox is
 mounting and it restarts, as it always has -- a share cannot be moved on a
 running guest, and you asked for a different directory.
 
--- a/cmd/brig/main.go
+++ b/cmd/brig/main.go
@@ -895,6 +895,10 @@ func listSandboxes(args []string) error {
 	if err != nil {
 		return err
 	}
+	// The one verb that knows what the runtime actually has, so the one place
+	// an entry for a sandbox that went away without going through `brig rm`
+	// can be dropped. See wrap.PruneSessions.
+	pruneSessionIndex(list)
 	sort.Slice(list, func(i, j int) bool { return list[i].Name < list[j].Name })
 	rows := make([]sandboxRow, 0, len(list))
 	for _, inst := range list {
@@ -912,6 +916,20 @@ func listSandboxes(args []string) error {
 		fmt.Println("(none -- `brig run claude` starts one)")
 	}
 	return nil
+}
+
+// pruneSessionIndex hands the session index every instance the runtime has, so
+// that an entry naming a sandbox that is not among them goes.
+//
+// Every instance and not only brig's own: a sandbox brig would not recognise
+// still exists, and the question the index is asking is whether the sandbox is
+// there at all.
+func pruneSessionIndex(list []runtime.Instance) {
+	live := make([]string, 0, len(list))
+	for _, inst := range list {
+		live = append(live, inst.Name)
+	}
+	wrap.PruneSessions(live)
 }
 
 // sandboxRow is one line of the listing, gathered before anything is printed
@@ -955,7 +973,7 @@ func printSandboxes(w io.Writer, rows []sandboxRow) {
 // name, and reporting the derived one was reporting the wrong directory with
 // no sign that it was wrong.
 func workspaceOf(vmName string, rt runtime.Runtime) string {
-	if ws := wrap.RememberedWorkspace(vmName); ws != "" {
+	if ws := wrap.WorkspaceOfSandbox(vmName); ws != "" {
 		return ws
 	}
 	rest := strings.TrimPrefix(vmName, sandboxPrefix)
@@ -1018,8 +1036,8 @@ func reset(args []string) error {
 		// terms: this goes through the runtime directly rather than through a
 		// Config, because reset works from the instance list and a stopped
 		// sandbox need not correspond to a profile brig can still look up.
-		wrap.ForgetWorkspace(inst.Name)
-		wrap.ForgetSession(inst.Name)
+		wrap.ForgetSandbox(inst.Name)
+		wrap.ForgetSlugClaim(inst.Name)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "brig: could not remove %s: %v\n", inst.Name, err)
 			continue
--- a/cmd/brig/main_test.go
+++ b/cmd/brig/main_test.go
@@ -8,6 +8,8 @@ import (
 	"testing"
 
 	"github.com/brig-sh/brig/internal/profile"
+	"github.com/brig-sh/brig/internal/runtime"
+	"github.com/brig-sh/brig/internal/wrap"
 )
 
 // The registry is built at run time now rather than being a package-level
@@ -336,8 +338,11 @@ func TestWorkspaceOfPrefersTheRecordedWorkspace(t *testing.T) {
 	t.Setenv("BRIG_STATE_DIR", dir)
 	t.Setenv("BRIG_WORKSPACE", "/somewhere/else")
 
-	index := filepath.Join(dir, "workspaces.json")
-	body := `{"brig-claude-code-rc23test": "/private/tmp/ws-rc23"}`
+	// Keyed by the ref, and the sandbox is in the value: the listing has a
+	// sandbox name in hand and no ref, so this is the direction it reads.
+	index := filepath.Join(dir, "sessions.json")
+	body := `{"claude-code@rc23test":` +
+		`{"home": "/private/tmp/ws-rc23", "sandbox": "brig-claude-code-rc23test"}}`
 	if err := os.WriteFile(index, []byte(body), 0o600); err != nil {
 		t.Fatal(err)
 	}
@@ -354,6 +359,39 @@ func TestWorkspaceOfPrefersTheRecordedWorkspace(t *testing.T) {
 	}
 	if got := workspaceOf("brig-claude-code", nil); got != "/somewhere/else" {
 		t.Errorf("ls reported %q for an unrecorded sandbox, want the derived path", got)
+	}
+}
+
+// `brig ls` is where brig learns that a sandbox went away without going
+// through `brig rm` -- removed with the runtime's own CLI, say -- so it is
+// where the index is pruned. An entry whose sandbox is still listed stays,
+// stopped or not: a stopped sandbox is the same session, and its home is where
+// the work is.
+func TestListingPrunesTheSessionsWhoseSandboxIsGone(t *testing.T) {
+	dir := t.TempDir()
+	t.Setenv("BRIG_STATE_DIR", dir)
+
+	index := filepath.Join(dir, "sessions.json")
+	body := `{` +
+		`"claude-code@kept": {"home": "/ws/kept", "sandbox": "brig-claude-code-kept"},` +
+		`"claude-code@gone": {"home": "/ws/gone", "sandbox": "brig-claude-code-gone"}}`
+	if err := os.WriteFile(index, []byte(body), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	// What the runtime lists: the stopped sandbox brig still has, and one
+	// container of somebody else's, which is why the prune is handed the whole
+	// list rather than brig's own share of it.
+	pruneSessionIndex([]runtime.Instance{
+		{Name: "brig-claude-code-kept", State: "stopped"},
+		{Name: "some-other-container", State: "running"},
+	})
+
+	if got := wrap.WorkspaceOfSandbox("brig-claude-code-kept"); got != "/ws/kept" {
+		t.Errorf("ls pruned a session whose sandbox it just listed: %q", got)
+	}
+	if got := wrap.WorkspaceOfSandbox("brig-claude-code-gone"); got != "" {
+		t.Errorf("ls kept %q for a sandbox the runtime does not have", got)
 	}
 }
 
--- a/internal/wrap/config.go
+++ b/internal/wrap/config.go
@@ -197,8 +197,8 @@ func Load(t profile.Profile, o Options, rt runtime.Runtime) (*Config, error) {
 		return nil, fmt.Errorf("cannot determine the current directory: %w", err)
 	}
 
-	// The sandbox name is settled first because the workspace may be read back
-	// from an index keyed by it. See index.go.
+	// The sandbox name is settled first because the entry read back below is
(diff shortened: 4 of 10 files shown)


asapranidis:
green.

ananos:
Nice work on the ref-keyed index. The reasoning in `index.go` is very clear,
and the migration guard is a neat trick -- a current session index cannot parse
as a map of strings, so the claims migration cannot fire on it. Good to see
that pinned in both a unit test and the smoke script.

One thing we should fix before this lands.

`readIndex` returns an empty map when the unmarshal fails. But a file holding
the JSON literal `null` unmarshals *successfully*, into a nil map. Both write
sites then panic:

```
rememberSession: assignment to entry in nil map
claimSlug:       assignment to entry in nil map
```

I reproduced it with a throwaway test, with `~/.brig/sessions.json` and
`~/.brig/slug-claims.json` each containing `null`:

```
=== RUN   TestNilMapProbe/rememberSession
    PANIC in rememberSession: assignment to entry in nil map
=== RUN   TestNilMapProbe/claimSlug
    PANIC in claimSlug: assignment to entry in nil map
```

This is the case the doc comment on `readIndex` explicitly rules out:
"Failing instead would make a stray file in ~/.brig able to stop every brig
command on the host." A panic is worse than the failure we are avoiding there.

The fix is one line:

```go
if err := json.Unmarshal(blob, &index); err != nil || index == nil {
    return map[string]V{}
}
```

To be fair, this pre-exists on `main` in the non-generic `readIndex`, so it is
not something this PR introduces. But this is the PR that folds both files onto
one function and writes the contract comment, so it seems like the right place
to close it. A test with `null` in the file would keep it closed.

asapranidis:
Confirmed and fixed in d3ceb0e.

Reproduced it before changing anything — your probe is exactly right. `null`
unmarshals without an error and leaves the map nil, and the error was the only
thing `readIndex` checked, so both writes took a nil map:

```
=== RUN   TestNilMapProbe/rememberSession
    PANIC in rememberSession: assignment to entry in nil map
=== RUN   TestNilMapProbe/claimSlug
    PANIC in claimSlug: assignment to entry in nil map
```

The read now treats a nil map as it treats a parse error:

```go
if err := json.Unmarshal(blob, &index); err != nil || index == nil {
    return map[string]V{}
}
```

Your framing of where it belongs is the one I took: it predates the generic
reader — `main`'s non-generic `readIndex` has the same hole — but this is the
change that folds both files onto one function and writes the contract comment,
so it is the one that owes the contract. The doc comment now names the case,
since "every way the read can fail" was silently untrue for one of them.

`TestAnIndexHoldingJSONNullIsIgnoredRatherThanFatal` keeps it closed. It writes
`null` into both files and then asserts what the panic would have taken: that
each reads back non-nil, that `rememberSession` records the workspace, and that
`claimSlug` claims the sandbox. It panics on the parent commit and passes here.

`gofmt` clean, `make all` green

ananos:
Re-verified on d3ceb0e, thanks for the quick turnaround.

I re-ran my original probe unchanged against the new head rather than only
reading the diff, and both writes are clean now:

```
--- PASS: TestNilMapProbeVerify/rememberSession
--- PASS: TestNilMapProbeVerify/claimSlug
```

`TestAnIndexHoldingJSONNullIsIgnoredRatherThanFatal` passes too, and
`go build`, `go vet` and the full suite are green here (bar `internal/secret`,
which wants an interactive keychain and fails the same way on `main`).

Good call asserting the two writes rather than just the non-nil read. The read
was never where the cost landed, so a test that only checked it would have gone
green on the broken version too.

One note for the stack rather than for this PR: the fix lives only on this
branch, so #118, #119, #120 and #122 still carry the old `readIndex` until they
rebase. Nothing to do about it now, just worth knowing if someone reviews #122
on its own before this lands.

ananos (approved):
The nil-map panic is fixed and independently re-verified on d3ceb0e, and I
found nothing else in the ref-keyed index. Details in the comment above.

PR #118, opened 2026-09-01
feat(cli): collapse the management commands into noun groups

### LINKED ISSUE #13: Settle one CLI grammar for 0.2, and show it in the help text

asapranidis:
## Summary

brig had three grammars at once and no way to tell from a command line which applied. A plural noun
was a command of its own (`brig profiles`), two noun commands were also spelled at the top level
(`brig import`, `brig export`), the verb set differed under every noun, and several operations had a
second undocumented spelling (`list`, `save`, `load`) that could only be found by accident and then
written into a script.

This settles the management half: one noun, one verb set, one spelling per operation.

`profile` becomes `agent` **at the CLI surface only**. `internal/profile`, `profile.Profile`,
`BRIG_PROFILE_DIR` and `~/.brig/profiles/` keep their names, so a ~99-file mechanical rename does
not land inside a grammar change and compete with it for review attention. That leaves two words
correct at once for a while, which is a real cost and a smaller one than the alternative.

## Related issues

Fixes #13. Part of #47.

**Stacked on #117** — base is `feat/109-session-index-by-ref`, so the diff shows only this change.
Merge order: #113, #115, #117, then this. It has no logical dependency on the ref work, but #115
rewrote the dispatch switch and the `usage` const that this edits, so a sibling branch off `main`
would have conflicted in exactly those two places for no benefit.

## Changes

`brig agent ls|show|new|edit|rm|import|export`, mapping onto commands that already existed:

| today | vNext |
| --- | --- |
| `brig profiles`, `brig profile ls`, `brig profile list` | `brig agent ls` |
| `brig export <p>` (prints, takes `--json`) | `brig agent show <a>` |
| `brig export <p> <name>` (saves under a new name) | `brig agent new <name> --from <a>` |
| `brig profile import`, `brig profile load`, `brig import` | `brig agent import` |
| `brig profile export`, `brig profile save` | `brig agent export` |
| `brig profile edit`, `brig profile rm` | `brig agent edit`, `brig agent rm` |
| `brig policies` | `brig policy ls` |

**One real fix, not a rename.** `brig agent new <name> --from <a>` rewrites `name:` inside the file
it writes. Starting from the closest existing agent is the documented way to author one, and the
recipe brig printed for it failed on its own second step: the copy was written under the new name
while the profile inside kept the old one, so `brig agent edit <name>` had nothing to open and the
only command that could reach the file named a different agent and deleted it without asking.

**Every retired spelling still works** and prints one line naming its replacement, using the
`deprecated()` mechanism already in the tree. That includes the undocumented aliases: `list`,
`save`, `load`, `secret list`, `secret rm` and `policy list` are out of the vocabulary but not out
of the binary — §5 of the issue says they "go", which is right about the vocabulary and would break
scripts inside the window #47 sets if read as removal in 0.2.

**`brig secret` keeps `create|update|read|delete|ls|import`.** This is a deliberate exception to
"one verb set under every noun", agreed on #47, and it has a test pinning it so the next reader does
not complete #13 by changing it back. The reason is in the Credentials section below.

**`brig policy` and `brig telemetry` keep their verb sets.** `policy` was already conformant; only
the plural retires. `telemetry status|on|off` is already noun-verb. Both were absent from #47's
command tree because they landed after it was filed.

**Not in scope:** `brig agent secrets`, which #47's tree lists but nothing in the tree does today —
new capability rather than a rename, so it wants its own issue. And no `brig start`: `brig run` is
how a stopped session is returned to.

### Two places this reached beyond the CLI, deliberately

`brig agents` and `brig template` keep their commands and behaviour untouched, but their
deprecation notices pointed at `brig profiles` and `brig profile` — spellings this commit retires.
A notice whose replacement answers with another notice is worse than no notice, so only the strings
they name changed.

Three files under `internal/profile/` changed, and every line is a user-facing string or a comment
naming a retired command. No identifier, type, package, env var or directory moved. The two that
matter: `exportHeader`, which is written **into every exported agent file** ("Edit it, then: brig
agent import <this file>"), and `LegacyHint()`, which is printed to anyone with files in the
pre-profile directory. Leaving those would have brig instructing people to run deprecated commands.

## Testing

`script/smoke.sh` goes from 149 to 193 passing checks. The `== profiles ==` section is now
`== agents ==` in the new vocabulary, plus a table-driven deprecation matrix and checks for each new
spelling. Its style is kept — output captured to a file before asserting, never piped into
`grep -q`, for the SIGPIPE reason documented at the top of that file.

Unit tests: `cmd/brig/agent_test.go` covers the new spellings, the unknown-subcommand wording,
`show` refusing a destination, `new` requiring `--from`, the help text vocabulary, and the `secret`
exception pin. `cmd/brig/deprecated_test.go` gains three tables: the retired-spelling matrix, the
`secret rm` alias, and a negative test asserting current spellings print **nothing** — which is the
one that catches a notice fired on the wrong path.

`script/smoke.sh` also has failures that predate this branch and depend on host state. The sets are
identical on this branch and on its parent `fd87d67`, with one textual difference: an assertion
label renamed from "the run is the profile it was copied from" to "…the agent…". Same assertion,
same cause, no change in which checks fail.

## Manual testing

Unlike the rest of this stack, none of this needs a runtime — these commands do not boot anything.
From a checkout of this branch:

```bash
make build                              # builds ./brig
export BRIG_PROFILE_DIR=$(mktemp -d)    # keep your own agents out of the way
```

**The new vocabulary**

```bash
./brig agent ls
./brig agent show claude                # prints the annotated profile; --json also works
./brig agent ls --help
```

**Every retired spelling still works, and says what replaced it**

```bash
./brig profiles          # brig: `brig profiles` is now `brig agent ls`
./brig profile ls        # brig: `brig profile ls` is now `brig agent ls`
./brig policies          # brig: `brig policies` is now `brig policy ls`
./brig import            # brig: `brig import` is now `brig agent import`
./brig export claude     # brig: `brig export` is now `brig agent export`
./brig secret rm         # brig: `brig secret rm` is now `brig secret delete`
```

Each prints its notice on stderr and then does the job. Confirm the second half too — a notice that
replaced the behaviour would be a regression, not a deprecation.

**The notices that used to point at a deprecated spelling**

```bash
./brig agents            # must say `brig agent ls`, not `brig profiles`
./brig template ls       # must say `brig agent`, not `brig profile`
```

**The `secret` exception**

```bash
./brig secret ls         # no notice: this verb set is unchanged on purpose
```

**The fix: `name:` is rewritten in the copy**

```bash
./brig agent new mytool --from claude
grep '^name:' "$BRIG_PROFILE_DIR"/mytool.yaml     # must print: name: mytool
./brig agent edit mytool                          # must find it (this is what used to fail)
./brig agent rm mytool
```

`brig agent new mytool` with no `--from` should explain that it copies an agent and so needs one.

## Checklist

**`script/smoke.sh`**: as described under Testing — failures predate the branch, sets identical, 44
checks added.

**Real runtime**: struck through. These verbs boot nothing. The dispatch they route through is
shared with the lifecycle verbs, and `script/smoke.sh` exercises that whole path against a stub
runtime with no change in behaviour.

**Docs**: `usage` and the group help texts are updated, because they are in the file being edited
and cannot not change. `README.md` and `docs/profiles.md` are **not**, and this makes them more
wrong than they already were — the flag table was stale after #115 and now the command vocabulary
is too. #111 owns that rewrite and has a comment recording both. Flagging it here so a reviewer is
not misled into reading those files as current.

No new mount, no new forwarded variable, no new host path the guest can reach. Two things worth
stating anyway, both about the credential store rather than the guest.

**Why `brig secret` keeps its verbs.** The unified set in the issue would have made `set` mean
create-or-update and `get` mean read. A typo that turns a read into a create-or-update silently
overwrites a stored credential, and the recovery is to go and find the original value again. The
inconsistency is cheaper than that failure, which is why the exception exists and why it is pinned
by a test rather than left to judgement.

**`brig agent show` and `brig agent export` print an agent's spec, which names its secrets and never
their values** — the same rule the command already followed under its old spelling. The tests assert
on the rendered bytes, so a value appearing in a field would fail them.

## LLM usage

message and this description. Verification was re-run independently of the agent that wrote the
change, from a frozen tree: `gofmt`, `make all`, and the smoke failure sets diffed against
`fd87d67`. Every command in the Manual testing section was run against the built binary before
being written down, including the `name:` rewrite and each deprecation notice. Two of the agent's
decisions were reviewed against the diff rather than taken on trust: the changed replacement strings
in `brig agents`/`brig template`, and every line it touched under `internal/profile/`.
Accountability for every line remains the author's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,18 +122,29 @@ separate Linux re-implementation of it.
 | `brig ls` | every brig sandbox, running or merely holding its name, with its workspace |
 | `brig reset` | stop and remove every brig sandbox. Workspaces are untouched |
 | `brig info <agent>` | the boundary a run would trust -- sandbox, workspace, image, credentials **by name only** -- and whether the guest will be authenticated (`brig env` is the old spelling) |
-| `brig profiles` | the profiles, their images, and what each one refuses to forward |
-| `brig profile ls\|export\|import\|edit\|rm` | manage profiles (`brig export` is the short form of `brig profile export`). `export --json` for JSON instead of YAML |
-| `brig policies` | every policy that parses, by name and description |
+| `brig agent ls` | the agents, their images, and what each one refuses to forward |
+| `brig agent show <agent>` | print one agent's spec. `--json` for JSON instead of YAML |
+| `brig agent new <name> --from <agent>` | copy an agent under a new name, ready to edit |
+| `brig agent edit\|rm\|import\|export` | manage agents |
 | `brig policy ls\|create\|edit\|show\|rm` | manage policies. `show --json` for JSON instead of YAML |
 | `brig secret create\|read\|update\|delete\|ls` | keep secrets in your keyring. macOS only for now |
 | `brig secret import <profile>` | fill that profile's secrets from your host, once. macOS only for now |
 | `brig telemetry status\|on\|off` | report what is counted, or turn the counting on or off. See [Telemetry](#telemetry) |
 | `brig version` | |
 
-`brig template …` and `brig agents` are the older spellings and still work
-for one release, each printing a one-line note naming the new one; they no
-longer appear above. There is deliberately no `brig template edit`.
+The older spellings all still work and each prints a one-line note naming the
+new one, so nothing scripted breaks in this release; they no longer appear
+above. `brig profiles` and the whole `brig profile` group are now `brig agent`,
+`brig policies` is `brig policy ls`, and top-level `brig import` and
+`brig export` are `brig agent import` and `brig agent export`. So are the
+undocumented aliases that were never in this table -- `list`, `save`, `load`,
+`secret rm`. `brig template …` and `brig agents` remain from an earlier rename,
+and there is deliberately no `brig template edit`.
+
+`brig secret` keeps `create|read|update|delete`, and that is deliberate rather
+than an oversight. The tidier `set`/`get` pair would make one typo turn a read
+into a write, silently replacing a stored credential with no way back to the
+original value. The inconsistency is the cheaper of the two.
 
 A brig line has three places a token can stand, and they are not the same kind
 of place. Left of the verb are brig's global flags, a closed set -- an unknown
@@ -255,7 +266,7 @@ The verbs line up deliberately, so muscle memory carries over:
 | `sbx create` | `brig create` | |
 | `sbx exec` / `sbx ls` / `sbx stop` / `sbx rm` / `sbx reset` | same | |
 | `sbx cp` | n/a | the workspace is a live host directory, so there is nothing to copy across |
-| `sbx template ls\|save\|load\|rm` | `brig profile ls\|export\|import\|edit\|rm` | brig's profiles describe the *agent*, not just its image, and are YAML like sbx's kits |
+| `sbx template ls\|save\|load\|rm` | `brig agent ls\|export\|import\|edit\|rm` | brig's profiles describe the *agent*, not just its image, and are YAML like sbx's kits |
 | `sbx secret set` | `brig secret create` + a profile's `secrets:` | brig has a store of its own and binds from it. Or point a profile at your existing secret manager: whatever puts a value in brig's environment is enough |
 | `sbx -t/--template` | `brig -t/--image` | |
 | `sbx -m/--memory`, `--cpus`, `-d`, `--name` | same | |
@@ -356,7 +367,7 @@ typo in a name is a message rather than a silently lost secret.
 
 A profile declares the names it wants out of this store, so a stored secret
 reaches a sandbox on its own -- as a file where the agent reads one, and as an
-environment variable where it does not. `brig profiles` prints each profile's
+environment variable where it does not. `brig agent ls` prints each agent's
 list, and which of those names `brig secret import` can fill for you:
 
 ```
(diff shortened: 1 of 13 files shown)


ananos (approved):
Reviewed in full. I exercised all eighteen retired spellings and every one works
and names its replacement, including the undocumented `list`, `save`, `load`,
`secret list` and `policy list`. The error messages are good throughout, and
`brig agent new mine codex` refusing to guess the word order rather than picking
one is the right choice.

The alias and reserved-name refusals on the new `--from` path are correct:
`brig agent new claude --from claude-code` and `brig agent new desktop --from
claude-code` are both refused with the reason.

#126 is filed as a follow-up, not blocking: `brig agent ls extra`, `brig agent
edit a b` and `brig agent import a b` accept and ignore the trailing word, where
the top-level verbs refuse it. It is the one remaining inconsistency in the
grammar this PR is settling.

One correction to the description. It presents the `name:` rewrite in
`brig agent new <name> --from <a>` as "One real fix, not a rename". That fix is
already on `main`: `profile.ExportAs` exists there, `exportProfile` already
calls it, and the rationale comment about the recipe stopping at its second step
is already in `internal/profile/file.go`. This PR re-spells the command, which
is worth doing, but a reviewer reading that paragraph will go looking for a
behaviour change and a test that are not in the diff.

PR #119, opened 2026-09-01
feat(cli): switch the lifecycle verbs to session refs

### LINKED ISSUE #5: Accept a session name in every lifecycle command

asapranidis:
## Summary

`brig ls` printed `brig-claude-code-refactor`, which no verb accepted. Copying what the list showed
produced `unknown profile`. This makes the printed identifier the operand.

The operand half was already done by #108 (`split` returns a `session.Ref`, the label maps onto the
session path), so `brig stop claude@refactor` worked before this PR. What was missing: the verb
tree, and `ls` printing a ref at all.

## Related issues

Fixes #5. Part of #47. **Stacked on #118** — base `feat/13-agent-noun-groups`.
Merge order: #113, #115, #117, #118, this.

## Changes

Verb collapse, each retired spelling kept working with a one-line notice:

| retired | replacement |
| --- | --- |
| `brig exec <ref> -- cmd` | `brig sh <ref> cmd` |
| `brig shell <ref>` | `brig sh <ref>` |
| `brig create <ref>` | `brig run -d <ref>` |
| `brig reset` | `brig rm --all` |
| `brig env <ref>` | `brig info <ref>` |
| `-n`, `--name NAME` | `<agent>@<label>` |

- `brig sh <ref> [command...]`: with a command runs it, without one opens a login shell.
- Verbless `brig <ref>` resolves **last in the dispatch chain**, so a token is tried as a ref only
  after every verb has been checked. Accepted, in no help text or example.
- `brig ls` gains a `REF` column. `brig ls -q` prints refs only.
- `-n`/`--name` is mapped to a label and warned about rather than aliased: it is also Claude Code's
  own flag.

**No-index-entry case in `ls`.** A sandbox with no entry in `~/.brig/sessions.json` (created before
the index, or pruned) has no recorded ref. Order: recorded ref, else derive from the sandbox name,
else `-`.

Deriving is safe because it cannot address a different sandbox. A sandbox is named
`brig-<agent>-<slug>`, so re-resolving a decomposed ref rebuilds the same name. The
`claude-code` + `refactor` vs `claude-code-refactor` ambiguity picks between two readings that
compose to one string, so either is a valid address for that sandbox. The two copies of the
decomposition loop (`ls`'s workspace column already had one) are now one `splitSandboxName`.

Out of scope: `sh --cwd/-u/-t/-T`, multi-ref `rm`/`stop`, `brig diff`/`apply`, `brig start`. None
exist today; all are new capability.

## Testing

`script/smoke.sh`: 193 → 219 passing assertions.

Unit: `lifecycle_test.go` (142), `listing_test.go` (157), `deprecated_test.go` +111. The round-trip
test feeds `brig ls -q` into every verb and asserts each accepts it — that is the acceptance
criterion for this issue.

### Smoke failure set: 3 lines differ from the parent, not 0

Local run: 11 FAIL on both this branch and `845f89a`. The sets are **not** textually identical:

- parent `create prints the sandbox name` + `run -d prints the name` → this branch
  `run -d prints the sandbox name`. Two assertions became one: the `create` case was rewritten in
  the taught spelling, and the duplicate `run -d` case became setup for `rm --all`.
- this branch adds `sh with no command opens a login shell`, which fails locally.

All 11 local failures are downstream of the first: `brig run` exits 1 with `.claude/.credentials.json
sits on virtiofs rather than an ephemeral mount`, so no agent or shell exec runs against the stub on
that host. The equivalent assertion written against the parent's own `shell` verb fails there
identically. Re-run with `env -u GH_TOKEN`: same 11.

`script/smoke.sh` runs in CI (`ci.yml:62`) and passes there, so CI is the arbiter for whether the new
assertion holds.

## Manual testing

```bash
make build

./brig create claude      # brig: `brig create` is now `brig run -d`
./brig exec claude -- ls   # brig: `brig exec` is now `brig sh`
./brig shell claude        # brig: `brig shell` is now `brig sh`
./brig reset               # brig: `brig reset` is now `brig rm --all`
./brig env claude          # brig: `brig env` is now `brig info`
```

Each prints the notice and then does the job. `runtime unavailable` after the notice means the line
parsed.

For `ls`, the ref column and the round trip need a runtime. The stub from `script/smoke.sh` works by
hand — extract the `hull` heredoc and export `BRIG_RUNTIME_BIN`, `BRIG_STATE_DIR`,
`BRIG_BOOT_ASSETS`, `BRIG_HYPERVISOR=vz`, `BRIG_VERIFY=off`:

```bash
./brig run -d claude@refactor
./brig ls          # REF column reads claude-code@refactor
./brig ls -q       # claude-code@refactor
for r in $(./brig ls -q); do ./brig info "$r"; done     # every ref must be accepted

rm "$BRIG_STATE_DIR/sessions.json"
./brig ls          # REF still claude-code@refactor, derived from the sandbox name
```

## Checklist

`script/smoke.sh` is ticked on the basis of CI, which is where it runs clean; see the section above
for the local host failures and why they are not this change.

Real runtime: not booted by the author of this branch. Every verb was exercised against the stub,
which covers the whole path except the VM.

Docs: README updated for the verbs this changes. The structural rewrite is #111.

No new mount, no new forwarded variable, no new host path reachable from the guest. `brig sh` uses
the same exec path `exec` and `shell` used. `ls` prints sandbox names, refs and workspace paths, as
before.

## LLM usage

that wrote the change: `gofmt`, `make all`, verb probes against the built binary, and the `ls`
round-trip plus the derive fallback exercised against the stub runtime with `sessions.json` removed.

Commits on this branch are **unsigned**: SSH signing via 1Password cannot be approved
non-interactively. `4f2ed2f`, `930e948`, `822596e` and `85dee40` lower in the stack are unsigned for
the same reason. The `main` ruleset does not require signatures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,15 +113,13 @@ separate Linux re-implementation of it.
 
 | command | what it does |
 | --- | --- |
-| `brig run <agent> [args…]` | start the sandbox if needed, then run the agent. Arguments pass through untouched |
-| `brig create <agent>` | start the sandbox without attaching; prints its name |
-| `brig shell <agent> [cmd…]` | a login shell inside the sandbox, or one command in it |
-| `brig exec <agent> -- cmd…` | run one command inside the sandbox |
-| `brig stop <agent>` | stop the sandbox, keep it. Starting it again is a boot, not a fresh creation |
-| `brig rm <agent>` | stop and remove the sandbox. The workspace is untouched |
-| `brig ls` | every brig sandbox, running or merely holding its name, with its workspace |
-| `brig reset` | stop and remove every brig sandbox. Workspaces are untouched |
-| `brig info <agent>` | the boundary a run would trust -- sandbox, workspace, image, credentials **by name only** -- and whether the guest will be authenticated (`brig env` is the old spelling) |
+| `brig run <ref> [args…]` | start the sandbox if needed, then run the agent. Arguments pass through untouched. `-d` starts it and exits, printing its name |
+| `brig sh <ref> [cmd…]` | a login shell inside the sandbox, or one command in it |
+| `brig stop <ref>` | stop the sandbox, keep it. Starting it again is a boot, not a fresh creation |
+| `brig rm <ref>` | stop and remove the sandbox. The workspace is untouched |
+| `brig rm --all` | stop and remove every brig sandbox. Workspaces are untouched |
+| `brig ls [-q]` | every brig sandbox, running or merely holding its name, with its ref and workspace. `-q` prints the refs alone, for a script |
+| `brig info <ref>` | the boundary a run would trust -- sandbox, workspace, image, credentials **by name only** -- and whether the guest will be authenticated |
 | `brig agent ls` | the agents, their images, and what each one refuses to forward |
 | `brig agent show <agent>` | print one agent's spec. `--json` for JSON instead of YAML |
 | `brig agent new <name> --from <agent>` | copy an agent under a new name, ready to edit |
@@ -132,6 +130,10 @@ separate Linux re-implementation of it.
 | `brig telemetry status\|on\|off` | report what is counted, or turn the counting on or off. See [Telemetry](#telemetry) |
 | `brig version` | |
 
+A `<ref>` is the session: `claude` is that agent's default session, and
+`claude@refactor` is a session of its own. `brig ls` prints the ref of every
+sandbox, and every verb above takes one -- see [Named sessions](#named-sessions).
+
 The older spellings all still work and each prints a one-line note naming the
 new one, so nothing scripted breaks in this release; they no longer appear
 above. `brig profiles` and the whole `brig profile` group are now `brig agent`,
@@ -140,6 +142,15 @@ above. `brig profiles` and the whole `brig profile` group are now `brig agent`,
 undocumented aliases that were never in this table -- `list`, `save`, `load`,
 `secret rm`. `brig template …` and `brig agents` remain from an earlier rename,
 and there is deliberately no `brig template edit`.
+
+The lifecycle verbs collapsed in the same way. `brig exec` and `brig shell` are
+both `brig sh`, which runs a command when you give it one and opens a login
+shell when you do not -- one verb, and which of the two you get is said by the
+line rather than by the word you reached for. `brig create` is `brig run -d`,
+`brig reset` is `brig rm --all`, and `brig env` is `brig info`. `-n`/`--name` is
+the label: `brig run claude@refactor` is what `brig run claude --name refactor`
+was. That one is mapped and warned about rather than quietly reinterpreted,
+because `--name` is also Claude Code's own flag.
 
 `brig secret` keeps `create|read|update|delete`, and that is deliberate rather
 than an oversight. The tidier `set`/`get` pair would make one typo turn a read
@@ -196,7 +207,8 @@ working; one that wants to branch on the reason now can.
 ```bash
 brig run claude -p "summarise this repo"   # headless, arguments passed through
 brig run claude -- --name not-a-session    # --name reaches claude, not brig
-brig shell codex 'ls -la ~'                # one command, in a login shell
+brig sh codex 'ls -la ~'                   # one command, in a login shell
+brig sh codex                              # a login shell, no command
 ```
 
 ### Named sessions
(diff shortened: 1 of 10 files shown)


ananos:
The verbless form is a good addition, and reading it last in the dispatch chain
is the right call. An agent named `ls` must not shadow the listing, and the
comment says exactly that.

One thing I would change. The fallback swallows the ref diagnosis that #115
works hard to produce:

```
$ brig run claude@BAD
brig: session label "BAD" is not usable as it stands: it would have to become
      "bad". ... Type "bad" if that is the session you want

$ brig claude@BAD
brig: unknown command "claude@BAD" (try `brig help`)
```

The comment justifies this with "the reader typed the token where a command
goes", which is fair for a bare word. For a token with an `@` in it I am not so
sure. No brig verb has an `@`, so such a token can only have been meant as a
ref. Same story for `brig claude@my-really-long-label` and `brig nosuch@x`.

Maybe we could report `refErr` (or the unknown-agent error) when the token
contains an `@`, and keep "unknown command" for everything else? The verbless
form would then get the same messages the verbed one already has. What do you
think?

asapranidis:
Taken, in 4864ffe. Your reading of the separator is the one that decides it:
`@` is reserved in `internal/session` for exactly one job, no verb has one, and
`ParseRef` cuts on the first — so an agent name cannot carry one either. A
token with an `@` in it has a single possible reading however badly it is
spelled, which makes "unknown command" a wrong answer rather than a cautious
one.

Scoped the way you suggested, so the existing justification survives where it
is still true. A bare word keeps the fallback, because `brig nosuchthing` is a
mistyped verb as readily as a mistyped agent, and `brig ls` with an agent named
`ls` is still the listing — the ref is tried last exactly as before. This only
changes what happens once it has failed.

```
$ brig claude@BAD
brig: session label "BAD" is not usable as it stands: it would have to become
      "bad". ... Type "bad" if that is the session you want
$ brig claude@a@b
brig: session ref "claude@a@b" has more than one "@". A ref is an agent, or an
      agent and one label: agent@label
$ brig @refactor
brig: session ref "@refactor" names no agent. ...
$ brig nosuch@refactor
brig: unknown profile "nosuch". `brig agent ls` lists them
$ brig nosuchthing
brig: unknown command "nosuchthing" (try `brig help`)
```

One deviation from your sketch, on the unknown-agent half. Rather than
reporting it in the fallback, the token falls through to the run line, which is
where the verbed form already reports it. Two spellings of one message would
have to be kept in step by hand; this way they are the same message because
they are the same line of code. It also picks up something I had not expected:
that path returns a `notFoundError`, so `brig nosuch@refactor` now exits 3 like
`brig run nosuch@refactor` instead of flattening to 1. The fallback was
mis-classifying the failure as well as mis-wording it.

`session.IsRefShaped` is exported for the test rather than a `strings.Contains`
at the call site, so `refSep` keeps one definition.

`TestARefShapedTokenIsDiagnosedAsARef` pins it as equality with the verbed form
rather than against fixed strings: the two spellings ask the same question, so
a later change to how brig answers it has to reach both or the pair stops
matching. It fails on all five tokens at the parent commit.

`gofmt` clean, `make all` green, and `script/smoke.sh` fails the same 11
host-dependent assertions here as on the parent.

ananos:
One thing left, and it is the other half of the mis-classification you found.

The wording is right for both cases now. The exit code is not. The verbless
branch returns `refErr` bare, where `split` wraps the same error in `usagef`,
so one mistake exits two ways:

```
TOKEN              brig X   brig run X
claude@BAD         1        2
claude@a@b         1        2
claude@            1        2
@refactor          1        2
nosuch@refactor    3        3   <- ok, this one falls through to the run line
```

README says 1 is "general failure -- something ran and did not finish" and 2 is
"usage error". Nothing ran, so 1 is the wrong class. It is the same fallthrough
you already fixed for the unknown-agent case, on the `refErr != nil` branch
instead.

Suggested fix:

```go
-	return refErr
+	return usagef("%s", refErr)
```

`TestARefShapedTokenIsDiagnosedAsARef` does not catch this because it compares
`.Error()` strings, which are already equal. The error type differs, not the
message. Adding the code to the same comparison would pin it:

```go
if exitCode(verbless) != exitCode(verbed) {
    t.Errorf("brig %s exits %d, brig run %s exits %d",
        ref, exitCode(verbless), ref, exitCode(verbed))
}
```

Table above captured from the built binary at 91fc803. The rest of this PR
checks out: all five tokens give the ref diagnosis, and a bare word still gets
"unknown command".

PR #120, opened 2026-09-01
feat(session): drop the slug length cap and stop truncating

### LINKED ISSUE #114: Raise the session label cap to 20, and warn where it moves a guest home

asapranidis:
## Summary

A session label was capped at ten characters. `experiment` fit exactly; `refactoring` did not.

The cap existed only to bound `Slug`'s truncation, and truncation is what made two different labels
land on one sandbox and one guest home. So the truncation goes and the cap goes with it — there is
nothing left for a length limit to protect against.

The issue as filed said raise it to 20. The decision was to remove it. If an unbounded label turns
out to be a problem, that gets addressed then.

## Related issues

Fixes #114. Part of #47. **Stacked on #119** — base `feat/5-lifecycle-verbs-refs`.
Merge order: #113, #115, #117, #118, #119, this.

## Changes

- `Slug` no longer truncates. `maxSlug` is deleted, not renumbered.
- `ParseRef`'s length check is gone. Its remaining rule — a label must already be what `Slug` would
  make of it — is now purely the character test, at any length.
- `session.LegacySlug` and `legacyBudget = 10` are new, and exist only so the notice below can
  recognise a name the old budget would have cut. `LegacySlug` has one caller
  (`internal/wrap/config.go`) and derives no path.
- The three `ref_test.go` cases built on `maxSlug` were rewritten rather than retargeted: a long
  label is now valid, so the property under test changed.

### `claimSlug` stays, and this is the reason

Removing truncation removes one source of collision, not all of them. Sanitising is not one-to-one:
`Foo`, `foo!` and `foo` are all `foo`, so two `--name` values can still land on one sandbox. That
refusal is what stops two agents sharing a guest home with whichever credentials arrived last, and no
length budget was ever protecting against it.

Its logic is unchanged. Its comments and two user-facing messages are not: they said a name
"shortens to" a sandbox, which is no longer how a collision happens.

Two test fixtures had to move for the same reason, and neither is a weakened assertion:

- `internal/wrap/slugclaim_test.go` collided `acme-corp-prod` with `acme-corp-staging`, which
  collide *only* by truncation. After this change they are distinct slugs, so the test would have
  failed at its own "the two names did not collide" guard. Repointed at `acme-corp-prod` versus
  `Acme Corp Prod` — a sanitisation collision, which is the class that survives. Same substitution in
  `script/smoke.sh`'s `== session collision ==` block.
- `cmd/brig/hygiene_test.go:309` asserted `claude@my-very-long-label` is refused, which was a pure
  length refusal. That label is now valid, so the case became a sanitisation refusal.

### The migration notice

Any existing session whose `--name` is longer than ten characters now slugs differently, so it gets a
new sandbox and a new guest home and the old pair is orphaned. The work in the workspace survives,
being on the host; the guest's in-memory state and an in-sandbox login do not.

It fires only when the slug the old budget would have produced differs from the new one **and** the
old workspace directory is on disk, and it names both homes and both sandboxes:

```
brig: session "refactoring-api" used to be shortened to "refactorin" and now keeps its
name in full, so it has a new home and a new sandbox:
<workspace>/ws-refactoring-api (brig-claude-code-refactoring-api) instead of
<workspace>/ws-refactorin (brig-claude-code-refactorin).
```

Computed in `Load` and held on the `Config` until `BuildEnv` prints it, beside `envWarnings`, which
is the existing precedent for a decision `Load` makes with no writer available. Printing it from
`cmd/brig` beside `LegacyHint()` would have matched the issue's tone reference but missed
`cmd/brigd`, which also calls `BuildEnv`.

The old home is read from the session index first and derived as `<base>-<oldSlug>` only as a
fallback, so a session started under `--workspace` or `BRIG_WORKSPACE` is named where it actually is.

### Accepted, and deliberately unguarded

- The bound is now the filesystem's: 255 bytes per path component. A label near that fails at
  directory creation with a filesystem error rather than a brig message.
- The sandbox is `brig-<agent>-<slug>`, so a long label is also subject to the runtime's own name
  limits.

No length guard, no warning and no truncation fallback was added for either.

## Testing

`script/smoke.sh`: 219 → 225 passing assertions, `FAIL` set byte-identical to the parent
(`fc476e4`), verified by diff. Six new assertions:

```
== the slug migration ==
  the migration notice names the home that was left behind
  the migration notice names the home in use now
  a session with nothing orphaned is not warned
  a name the old budget left alone is not warned
  a long label gets a sandbox of its own
  a long label brig would rewrite is still refused
```

Unit: `internal/wrap/slugmigration_test.go` (new), plus the rewritten `ref_test.go`,
`session_test.go` and `slugclaim_test.go` cases.

The 11 pre-existing `FAIL` lines are host-specific: `brig run` aborts at
`.claude/.credentials.json sits on virtiofs rather than an ephemeral mount`, so nothing needing an
in-guest exec runs against the stub. The same suite passes in CI (`ci.yml:62`).

## Manual testing

No runtime needed for the label rules:

```bash
make build

./brig run 'claude@my-really-long-label-that-is-way-over-ten'   # accepted (parses)
./brig run 'claude@Refactor'                                    # refused: would become "refactor"
./brig run 'claude@my really long label'                        # refused: would become "my-really-long-label"
```

`runtime unavailable` means the line parsed, i.e. the label was accepted.

The migration notice needs a sandbox. Using the stub from `script/smoke.sh` (extract the `hull`
heredoc, export `BRIG_RUNTIME_BIN`, `BRIG_STATE_DIR`, `BRIG_WORKSPACE`, `BRIG_BOOT_ASSETS`,
`BRIG_HYPERVISOR=vz`, `BRIG_VERIFY=off`):

```bash
mkdir -p "$BRIG_WORKSPACE-refactorin"          # the home the old budget would have made
./brig run -d claude --name refactoring-api    # notice names both homes and both sandboxes

# and the two quiet cases
rm -rf "$BRIG_WORKSPACE-refactorin"
./brig run -d claude --name refactoring-api    # silent: nothing orphaned
./brig run -d claude --name short              # silent: the old budget left it alone
```

## Checklist

`script/smoke.sh` is ticked on CI, where it runs clean; the local host failures are described above.

Real runtime: not booted. Exercised against the stub, which covers the whole path except the VM.

Docs: the README sentence stating a ten-character limit is corrected. The structural rewrite is #111.

The failure this area guards against is two sessions on one guest home, each handed different
credentials, with whichever booted last leaving its own behind. `claimSlug` is what refuses it and it
is unchanged — see the section above for why removing truncation does not remove the need for it, and
why the surviving collision class is now the primary test case rather than an additional one.

The migration notice exists so that a changed slug is not a silent relocation: without it, a session
would come back on a new empty home with no indication that the old one still holds its state.

No new mount, no new forwarded variable, no new host path reachable from the guest.

## LLM usage

that wrote the change: `gofmt`, `make all`, label-rule probes against the built binary, and all three
migration-notice cases (fires, and both quiet cases) exercised against the stub runtime. The claim
that `claimSlug` was untouched was checked against the diff and corrected — its logic is unchanged,
its comments and messages are not.

Commit is unsigned: SSH signing does not work non-interactively in the environment it was produced
in. The `main` ruleset does not require signatures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,18 +221,18 @@ brig run claude --name refactor    # the same session, the older spelling
 A session of its own: its own workspace (`~/brig/claude-code-refactor`), its
 own sandbox, and the name you typed reaches the agent as its display name.
 
-Paths use a short lowercase form of the name -- letters, digits, dot, dash and
-underscore, ten characters -- so `my_project` and `my-project` stay separate,
-while `Foo` and `foo` do not (macOS filesystems ignore case, and two names
-sharing one directory but not one VM is a bug waiting to happen). If the slug
-differs from what you typed, brig says which directory it used.
-`brig info claude@refactor` prints the one in use.
+Paths use a lowercase form of the name -- letters, digits, dot, dash and
+underscore -- so `my_project` and `my-project` stay separate, while `Foo` and
+`foo` do not (macOS filesystems ignore case, and two names sharing one directory
+but not one VM is a bug waiting to happen). Nothing is shortened, so the name you
+pick is the directory you get. If the slug differs from what you typed, brig says
+which directory it used. `brig info claude@refactor` prints the one in use.
 
 The `@` form is the stricter of the two: a label brig would have to rewrite --
-uppercase, a space, more than ten characters -- is refused rather than quietly
-shortened, and the message names what it would have become. That is what makes
-the label safe to use as an address: nothing is silently mapped onto a directory
-you did not ask for. `--name` keeps its older, lenient behaviour.
+uppercase, a space, a slash -- is refused rather than quietly rewritten, and the
+message names what it would have become. That is what makes the label safe to use
+as an address: nothing is silently mapped onto a directory you did not ask for.
+`--name` keeps its older, lenient behaviour.
 
 **Every verb takes the ref.** A named session is addressed as
 `<agent>@<label>`, and the older agent-plus-`--name` spelling still works
--- a/cmd/brig/hygiene_test.go
+++ b/cmd/brig/hygiene_test.go
@@ -306,7 +306,9 @@ func TestParseRefusesABadRef(t *testing.T) {
 		{"claude@"},
 		{"claude@a@b"},
 		{"claude@Refactor"},
-		{"claude@my-very-long-label"},
+		// Long is not what refuses this -- a label is not shortened, so there
+		// is no length to refuse -- the spaces are.
+		{"claude@my very long label"},
 		{"@refactor"},
 	} {
 		_, _, _, err := parse(args)
--- a/cmd/brig/main.go
+++ b/cmd/brig/main.go
@@ -61,7 +61,7 @@ usage:
 A <ref> is the session. claude is that agent's default session, and
 claude@refactor is a session of its own -- its own workspace, its own sandbox,
 and the label reaching the agent as its display name. A label brig would have
-to shorten is refused rather than shortened. brig ls prints the ref of every
+to rewrite is refused rather than rewritten. brig ls prints the ref of every
 sandbox, and every verb above takes one.
 
 flags (before the agent's own arguments; -- ends brig's parsing):
@@ -354,9 +354,9 @@ func run(args []string) error {
 	if err != nil {
 		return err
 	}
-	// The slug is shortened and sanitised, so the directory a named session
-	// gets rarely reads back as the name typed. Say which directory it actually
-	// is whenever the two differ. A name that shortens onto another session's
+	// The slug is sanitised, so the directory a named session gets does not
+	// always read back as the name typed. Say which directory it actually is
+	// whenever the two differ. A name that sanitises onto another session's
 	// sandbox is refused later, at EnsureRunning; this only names the directory.
 	if cfg.RawName != "" && cfg.RawName != cfg.Slug {
 		fmt.Fprintf(os.Stderr, "brig: session %q uses %s (sandbox %s)\n",
--- a/internal/session/ref.go
+++ b/internal/session/ref.go
@@ -23,21 +23,19 @@ const refSep = "@"
 // places that have to agree: the sandbox name, which is the profile's name
 // plus "-" plus the slug (see internal/wrap/config.go), and the workspace
 // directory the guest gets as its home. Slug is what turns a name into both,
-// and Slug truncates at maxSlug -- so a cap any longer than that would let two
-// labels differing only past character maxSlug slug the same, land on one
-// sandbox and share one home directory. Refusing the label rather than
-// truncating it is what removes that collision, instead of detecting it later
-// and asking the user to pick again.
+// so a label Slug would rewrite is a label that addresses a directory nobody
+// typed. Refusing it is what keeps a ref an address, instead of quietly
+// mapping one spelling onto a session started under another.
 //
-// So the rule is: a label must already be what Slug would make of it. Below
-// the cap Slug cannot truncate, which is why the length check comes first and
-// why a single `label != Slug(label)` is then an exact test for "not
-// slug-clean" -- the character rules stay in Slug and are not restated here,
-// so the two cannot drift.
+// So the rule is: a label must already be what Slug would make of it. Slug
+// changes characters and nothing else -- it does not shorten -- so the single
+// `label != Slug(label)` is an exact test for "not slug-clean" at any length,
+// and there is no length for a label to be refused for. The character rules
+// stay in Slug and are not restated here, so the two cannot drift.
 //
 // --name is deliberately not held to this. It keeps its older, lenient
-// behaviour, truncation and the warning that names the directory it actually
-// got; the '@' form is the strict one.
+// behaviour, sanitising the name and reporting the directory it actually got;
+// the '@' form is the strict one.
 func ParseRef(s string) (Ref, error) {
 	agent, label, hasLabel := strings.Cut(s, refSep)
 	if strings.Contains(label, refSep) {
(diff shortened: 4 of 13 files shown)


ananos:
Two things here, and the first is a question rather than a defect claim.

**The migration notice may not fire for the sessions it is for.**

`slugMigrationNotice` reads the old home from the session index first, so that a
session started under `--workspace` is named where it actually is. I do not
think that read can hit once the whole stack is one release:

- #117 deletes `workspaces.json` on sight, which is where a pre-upgrade
  session's custom workspace was recorded.
- After this PR `Slug` never truncates, so no `sessions.json` entry can carry a
  truncated slug as its label.

So `rememberedWorkspace(sessionKey(profile, old), oldVM)` misses, and it falls
back to `base + "-" + old`. For a `--workspace` session that path is not where
the directory is, so `os.Stat` fails and the function returns nil. No notice, for
exactly the session most likely to lose an in-guest login.

It works if #117 to #119 ship in a release before this one, since then the index
does hold truncated-slug entries. What is the release plan? If they all land in
0.2 then the index read is dead and the notice covers default-workspace sessions
only, which is worth saying in the comment rather than implying the wider cover.

**No upper bound replaces the cap.**

`Slug` no longer truncates and `ParseRef` has no length check, so a label is
unbounded. Over ~255 bytes it fails at `mkdir` rather than at parse:

```
mkdir 255 chars -> ok
mkdir 300 chars -> [Errno 63] File name too long
```

The workspace component is `<profile>-<slug>`, so the usable budget is 255 minus
the profile name and a dash. I take the point in the description that an
unbounded label can be addressed if it becomes a problem. A check in `ParseRef`
or `Resolve` against that budget would keep the failure where the user can act
on it, and is a few lines.

Neither blocks this from my side. Checked at parse level with the runtime
stubbed out, on db5415a.

ananos (approved):
Removing the truncation rather than raising the cap is the right call, and the
reasoning that truncation was the collision source, not length, is correct.
Keeping `claimSlug` for the sanitisation collisions that survive is also right,
and moving the two test fixtures from truncation collisions to sanitisation
collisions is an honest change rather than a weakened assertion.

I checked `LegacySlug` against the old `Slug` ordering (trim, cut, trim) and
they agree.

The two points in my comment above are not blocking: the migration notice may
not fire for `--workspace` sessions depending on the release plan, and nothing
bounds a label now, so a very long one fails at `mkdir` instead of at parse.
Happy for both to be handled here or separately.

PR #121, opened 2026-09-01
feat(policy): attach, detach, and check policy bindings

marrgko:
## Summary

Adds binding a policy to a profile or a session.

`brig policy attach <policy> <profile> [-n NAME]` binds a policy to every
run of a profile, or -- with `-n` -- to one session by name. `detach`
reverses it. The effective policy set for a run is the union of the
profile's inline `policy:` list, the profile's attachments, and the
session's attachments, deduplicated by name (`EffectivePolicies`).

`brig policy check <profile> [-n NAME]` reports what applies to a run and
whether brig can enforce anything against the target at all: a
`kind: shell`/`kind: gui` profile has no agent to enforce anything
against, and a bound name that no longer resolves to a policy cannot be
enforced either. Either one exits non-zero.

Because a policy name can be bound from three places at once, `rm` and
`edit` refuse to break a binding -- deleting the file, or renaming it away
from the bound name -- unless `--force`. `brig policy ls` shows what binds
each policy, under the policy it binds.

## Changes

- `brig policy attach|detach <policy> <profile> [-n NAME]` -- bind a policy
  to a profile or one session, and reverse it. Refuses when the target
  cannot be enforced (`kind: shell`/`kind: gui`), or when the policy is
  already declared inline on the profile.
- `internal/policy.EffectivePolicies` -- the union rule: inline + profile
  attachments + session attachments, deduplicated.
- `brig policy check <profile> [-n NAME]` -- what is effectively bound, and
  whether anything can be enforced against the target at all.
- `brig policy rm`/`edit` refuse to delete a file or save a rename that
  would leave a binding pointing at nothing, unless `--force`.
- `brig policy ls` shows what binds each policy.
- `Profile` gains an inline `policy:` field; `internal/policy/attachments.go`
  holds the separate `attachments.yaml` record `attach`/`detach` write to.
- `docs/policies.md` and `README.md` updated for all of the above.

## Checklist

- ~~I have exercised the change against a real runtime (`brig run <agent>`), if it touches the run, exec or credential path~~ (this slice writes only to `attachments.yaml` and profile/policy files on the host; it does not touch the run or boot path)

## LLM usage

Authored with assistance from Claude Code (Sonnet 5); I reviewed every commit and am accountable for the change.


ananos:
Read through this one and exercised the commands. `make all` is green here and
the branch is already rebased onto `c06f9bc`, so it sits on the merged stack.

The mechanics are solid. `attachments.yaml` is written through a temp file and
renamed, every map is nil-guarded before it is written into, attach is
idempotent (three attaches of one policy leave one entry), and detach fully
unbinds. Every refusal in the description does what it says: `kind: shell` is
turned away, `rm` refuses a bound policy and names what binds it, `detach` on an
inline policy explains to edit the profile instead, and an unknown profile is
named.

Two things.

**1. Nothing at the point of use says a policy is not enforced.**

`docs/policies.md` is admirably direct about this:

> Nothing today makes an agent's outbound traffic actually respect a policy you
> write.

The CLI does not repeat it, and the CLI is where someone acts:

```
$ brig policy attach egress claude-code
attached egress to claude-code

$ brig policy check claude-code
egress
[exit 0]
```

I confirmed `EffectivePolicies` has exactly one caller, `cmd/brig/policy.go:751`,
which is `check` itself. Nothing in `internal/wrap` consults a policy, so no
boot or run path reads one.

Read from the terminal, "attached" and a `check` that prints the policy name and
exits 0 say the rule is live. The word `check` carries most of that weight: it
is the verb you reach for to confirm something is in force. The two messages
that do mention enforcement point the other way as well, since both describe
enforcement as a thing that happens:

```
cannot enforce any policy on x: x is kind: shell ...
x is bound to y, which does not exist as a policy -- nothing can enforce what is not there
```

For a security feature the gap between recorded and enforced is where false
confidence forms, and a user who never opens `docs/policies.md` has no way to
learn it from the tool. A line on `attach` and on `check` would close it, for
example:

```
attached egress to claude-code
note: no policy is enforced at runtime yet; this records the binding only
```

Worth deciding deliberately rather than leaving to the docs, since the docs
already made the honest choice and only the CLI is out of step with them.

**2. A corrupt `attachments.yaml` surfaces the YAML library rather than brig.**

```
$ printf 'not: [valid\n' > attachments.yaml
$ brig policy check claude-code
brig: error converting YAML to JSON: yaml: line 1: did not find expected ',' or ']'

$ printf 'profiles: "a string"\n' > attachments.yaml
$ brig policy check claude-code
brig: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string
      into Go struct field Attachments.profiles of type map[string][]string
```

Neither names the file, so there is nothing to act on: "converting YAML to JSON"
and "Go struct field Attachments.profiles" are the implementation talking. Every
other refusal in this PR names the thing at fault and what to do about it, which
is why these two stand out.

Failing rather than tolerating looks right to me here, and deliberately unlike
`internal/wrap/index.go`, which treats a corrupt file as empty. A binding record
that silently reads as "nothing is bound" is worse than a refusal. So this is
about the wording, not the behaviour. Naming the path would be enough.

`null` and an empty file are both handled and read as no attachments, which is
the right split.

Checked on macOS arm64 against the built binary, with `XDG_CONFIG_HOME` and
`BRIG_PROFILE_DIR` pointed at scratch directories. I did not boot a guest, which
this PR does not touch.

marrgko:
Thanks for exercising the commands rather than reading them. Both points are
addressed, one commit each, so they can be reviewed separately.

Paths below are shortened; the runs used a scratch `XDG_CONFIG_HOME` and
`BRIG_PROFILE_DIR`.

## 1. The CLI now says a binding is recorded, not enforced -- c2bb2d0

You were right that `check` carries most of the weight here. It is the verb you
reach for to confirm something is live, and it was answering with a name and an
exit code of zero.

```
$ brig policy attach egress claude-code
attached egress to claude-code
note: no policy is enforced at runtime yet; this records the binding only

$ brig policy check claude-code
egress
note: no policy is enforced at runtime yet; this records the binding only
```

Three decisions behind that output.

**It goes to stderr**, where this CLI already puts every advisory. `check` prints
one policy name per line, and anything looping over that output would otherwise
read the note as a name:

```
$ brig policy check claude-code 2>/dev/null
egress
```

**It is only printed where the answer would otherwise read as a verdict.** A
refusal already names what cannot be enforced, and "no policy applies" leaves
nothing to misread, so neither carries the note:

```
$ brig policy check codex
no policy applies to codex

$ brig policy check ubuntu
no policy applies to ubuntu
brig: cannot enforce any policy on ubuntu: ubuntu is kind: shell, which has no agent to hook an egress rule into
```

**The sentence is one exported constant**, `policy.NotEnforcedNote`, rather than
two string literals. It is a claim about what brig does, printed by two
commands: the moment they disagree, one of them is lying about the tool it
belongs to, and two copies leave nothing to catch an edit that changes one and
not the other.

`docs/policies.md` states the stderr and one-constant reasoning next to the
worked example; the absence of the note on the refusal is visible in the
`check` example there rather than spelled out.

## 2. A corrupt attachments.yaml now names the file -- e3e37d7

Behaviour is unchanged, as you suggested -- failing rather than tolerating, and
deliberately unlike `internal/wrap/index.go`. Only the wording changed, and the
reason for the split is now written above `LoadAttachments` so it does not read
as an oversight later.

```
$ printf 'not: [valid\n' > ~/.config/brig/policies/attachments.yaml
$ brig policy check claude-code
brig: /home/you/.config/brig/policies/attachments.yaml is not a valid attachment record: error converting YAML to JSON: yaml: line 1: did not find expected ',' or ']'

$ printf 'profiles: "a string"\n' > ~/.config/brig/policies/attachments.yaml
$ brig policy check claude-code
brig: /home/you/.config/brig/policies/attachments.yaml is not a valid attachment record: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Attachments.profiles of type map[string][]string
```

The library's own text is kept after the colon rather than replaced. It is the
only thing that says *where* in the file the problem is -- `line 1` in the first
case, the field name in the second -- which is what you need once you know which
file to open.

`null` and an empty file still read as no attachments, both at exit 0.

`TestLoadAttachmentsNamesTheFileItCouldNotParse` covers both shapes, since they
come from different layers and neither names the file on its own.

marrgko:
Going through the surrounding code while fixing the two above turned up five
more things. They are separate commits so they can be taken or dropped
individually. One thing I found and did **not** fix is at the bottom.

Paths shortened; runs used a scratch `XDG_CONFIG_HOME` and `BRIG_PROFILE_DIR`.

## `create` could overwrite brig's own record -- 1e5d1ba

`LoadAll` skips `attachments.yaml` so it is not read as a broken policy. Nothing
stopped `create` from writing to that same path, which would have destroyed the
attachment record and produced a policy no command could see.

```
$ brig policy create attachments
brig: "attachments" is reserved: brig keeps attachments.yaml in the policy directory for its own records, so a policy cannot be called that. Pick another name
[exit 1]
```

Refused regardless of `--force`: forcing is for replacing a policy you own, not
for overwriting brig's bookkeeping. `edit` makes the same refusal on a rename --
that file is not the reserved one, so saving would destroy nothing, but it would
leave a policy under a name `create` cannot produce, reachable only through
`edit`. Two ways in with different rules is a bug either way.

## `attach -n` could record a session no run would start -- 7ca0c19

A session's sandbox and workspace are named from its slug. `brig run -n "My
Work"` starts `my-work`, so recording `My Work` bound a policy to nothing, and
`check -n my-work` would report nothing bound.

```
$ brig policy attach egress claude-code -n "My Work"
brig: session "My Work" is not usable as it stands: it would have to become "my-work". Type "my-work" if that is the session you want

$ brig policy attach egress claude-code -n Desktop
brig: session "Desktop" becomes "desktop", which the claude-desktop profile already uses. Pick another name

$ brig policy attach egress claude-code -n "!!!"
brig: session "!!!" has no usable characters. Sessions use letters, digits, dot, dash and underscore
```

Reserved is tested against the slug, the way `Resolve` tests it, and tested
before the rewrite refusal so a name is never turned away with advice that is
itself refused -- `-n Desktop` would otherwise be told to type `desktop`, which
`claude-desktop` owns. The rewrite refusal is the one rule here that `Resolve`
has no equivalent of.

`run --name` stays lenient and sanitises, but that is older behaviour kept for
compatibility. `-n` here is new and carries no such debt.

The check is in `attach` only, not in the shared parser. `detach` and `check`
read the record and have to be able to name whatever is in it, including a key a
hand edit put there -- holding them to this rule would refuse the exact spelling
a listing prints and leave nothing able to remove it:

```
$ brig policy ls
egress
                bound to: claude-code, claude-code -n My Work
$ brig policy detach egress claude-code -n "My Work"
detached egress from claude-code -n My Work
[exit 0]
```

## `attach` announced a binding it had not saved -- bcc3504

`Save` ran after the `Printf`, so a directory it could not write printed success
on stdout and exited non-zero. Anything reading stdout would believe the binding
landed. Built at the parent commit for contrast:

```
$ brig policy attach egress codex 2>/dev/null     # before
attached egress to codex
[exit 1]

$ brig policy attach egress codex 2>/dev/null     # after
[exit 1]
```

`detach` had the same order and is fixed with it. `create` and `rm` already
printed after their write.

## `(no such policy)` said something untrue -- 5f98344

This changes the second string you quoted under point 1.

A name can be absent from the loaded set for two reasons, and brig cannot always
tell them apart: nothing declares it, or the file that declares it did not parse.
An unparseable file has no readable `name:` to match against. In the second case
the file is sitting right there, named on stderr one line above:

```
$ brig policy check claude-code                    # before
brig: ignoring 1 unusable policy file(s):
  egress.yaml: error converting YAML to JSON: yaml: line 4: did not find expected ',' or ']'
egress (no such policy)
brig: claude-code is bound to egress, which does not exist as a policy -- nothing can enforce what is not there
```

`does not exist` argues against the fix. Now:

```
$ brig policy check claude-code                    # after
brig: ignoring 1 unusable policy file(s):
  egress.yaml: error converting YAML to JSON: yaml: line 4: did not find expected ',' or ']'
egress (not loaded)
brig: claude-code is bound to egress, which no policy loads under -- nothing can enforce what did not load
```

Same commit: `ls` now shows a binding left behind by `--force`. Both `rm --force`
and a forced rename say so at the time and then nothing does, so the listing was
the one place that leftover could reappear:

```
$ brig policy rm egress --force
removed /home/you/.config/brig/policies/egress.yaml
$ brig policy ls
egress          (not loaded; bound to claude-code)
```

Previously that printed `no policies yet`, which contradicted the state it was
in: having nothing to manage and having a binding to clean up are different.

## `Bindings` read the registry itself -- d38b3aa

Not a reachable bug: `profile.Load` runs at `cmd/brig/main.go:169` before any
verb, so the registry is always populated by the time this is called. It takes
its profiles as an argument now, the way `CheckCoverage` and `EffectivePolicies`
already do. A caller that had not loaded the registry would otherwise have got
an answer with every inline binding silently missing, and `rm` would then delete
a policy a profile declares inline without the refusal that exists to stop it.
Sorting moved in with it, on a copy, since the walk order decides the printed
order.

Also folded in here: `sessionGivenEmpty` was walking the flag set by hand
instead of calling the existing `seen` helper in `main.go`.

---

## Not fixed: concurrent writes to attachments.yaml

`attach` and `detach` are read-modify-write with no lock. Each reads the whole
record, changes one entry, and renames a complete replacement over it. The
rename is atomic, so the file is never torn -- but a concurrent writer's changes
are gone, and nothing reports it. Five runs, eight `attach` in parallel each
time, all on distinct policies:

```
run 1: 8 parallel attach -> 4 of 8 recorded; commands reporting failure: 0
run 2: 8 parallel attach -> 4 of 8 recorded; commands reporting failure: 0
run 3: 8 parallel attach -> 6 of 8 recorded; commands reporting failure: 0
run 4: 8 parallel attach -> 6 of 8 recorded; commands reporting failure: 0
run 5: 8 parallel attach -> 5 of 8 recorded; commands reporting failure: 0
```

Every one of those commands exited 0.

I left it alone deliberately. brig does use `flock`, but only in
`cmd/brigd/main.go` to keep a second daemon from starting; no bookkeeping file
is written under a lock. `internal/wrap/index.go` has exactly this shape, and
`claimSlug` on top of it -- read the whole index, change one entry, rename a
replacement over it. Locking `attachments.yaml` alone would make it the only
guarded one of the three, and guarding the others is a decision about the whole
tool rather than about this PR. Interactive use does not hit it. A script that
attaches in a loop does.

Happy to do it as its own PR if you want it before this lands.

marrgko:
Going through the surrounding code while fixing the two above turned up five
more things. They are separate commits so they can be taken or dropped
individually. One thing I found and did **not** fix is at the bottom.

Paths shortened; runs used a scratch `XDG_CONFIG_HOME` and `BRIG_PROFILE_DIR`.

## `create` could overwrite brig's own record -- 1e5d1ba

`LoadAll` skips `attachments.yaml` so it is not read as a broken policy. Nothing
stopped `create` from writing to that same path, which would have destroyed the
attachment record and produced a policy no command could see.

```
$ brig policy create attachments
brig: "attachments" is reserved: brig keeps attachments.yaml in the policy directory for its own records, so a policy cannot be called that. Pick another name
[exit 1]
```

Refused regardless of `--force`: forcing is for replacing a policy you own, not
for overwriting brig's bookkeeping. `edit` makes the same refusal on a rename --
that file is not the reserved one, so saving would destroy nothing, but it would
leave a policy under a name `create` cannot produce, reachable only through
`edit`. Two ways in with different rules is a bug either way.

## `attach -n` could record a session no run would start -- 7ca0c19

A session's sandbox and workspace are named from its slug. `brig run -n "My
Work"` starts `my-work`, so recording `My Work` bound a policy to nothing, and
`check -n my-work` would report nothing bound.

```
$ brig policy attach egress claude-code -n "My Work"
brig: session "My Work" is not usable as it stands: it would have to become "my-work". Type "my-work" if that is the session you want
```

Two more refusals come with it, both `Resolve`'s: a value with nothing usable in
it, and one that is some profile's own workspace. Reserved is tested against the
slug and before the rewrite refusal, so `-n Desktop` is not told to type
`desktop`, which `claude-desktop` owns.

`run --name` stays lenient and sanitises, but that is older behaviour kept for
compatibility. `-n` here is new and carries no such debt.

The check is in `attach` only, not in the shared parser: `detach` and `check`
must still be able to name whatever is already in the record, including a key a
hand edit put there. Holding them to this rule would refuse the exact spelling a
listing prints and leave nothing able to remove it.

## `attach` announced a binding it had not saved -- bcc3504

`Save` ran after the `Printf`, so a directory it could not write printed success
on stdout and exited non-zero. Anything reading stdout would believe the binding
landed. Built at the parent commit for contrast:

```
$ brig policy attach egress codex 2>/dev/null     # before
attached egress to codex
[exit 1]

$ brig policy attach egress codex 2>/dev/null     # after
[exit 1]
```

`detach` had the same order and is fixed with it. `create` and `rm` already
printed after their write.

## `(no such policy)` said something untrue -- 5f98344

This changes the second string you quoted under point 1.

A name can be absent from the loaded set for two reasons, and brig cannot always
tell them apart: nothing declares it, or the file that declares it did not parse.
An unparseable file has no readable `name:` to match against. In the second case
the file is sitting right there, named on stderr one line above:

```
$ brig policy check claude-code
brig: ignoring 1 unusable policy file(s):
  egress.yaml: error converting YAML to JSON: yaml: line 4: did not find expected ',' or ']'
egress (no such policy)                                                    # before
brig: claude-code is bound to egress, which does not exist as a policy -- nothing can enforce what is not there

egress (not loaded)                                                        # after
brig: claude-code is bound to egress, which no policy loads under -- nothing can enforce what did not load
```

`does not exist` argues against the fix, one line under the file it would fix.

Same commit: `ls` now shows a binding left behind by `--force`. Both `rm --force`
and a forced rename say so at the time and then nothing does, so the listing was
the one place that leftover could reappear:

```
$ brig policy rm egress --force
removed /home/you/.config/brig/policies/egress.yaml
$ brig policy ls
egress          (not loaded; bound to claude-code)
```

Previously that printed `no policies yet`, which contradicted the state it was
in: having nothing to manage and having a binding to clean up are different.

## `Bindings` read the registry itself -- d38b3aa

Not a reachable bug -- `profile.Load` runs at `cmd/brig/main.go:169` before any
verb. It takes its profiles as an argument now, the way `CheckCoverage` and
`EffectivePolicies` already do, so a caller that had not loaded the registry
cannot get an answer with every inline binding silently missing and have `rm`
delete an inline-declared policy without its refusal. Sorting moved in with it,
on a copy, since the walk order decides the printed order.

Also here: `sessionGivenEmpty` was walking the flag set by hand instead of
calling the existing `seen` helper in `main.go`.

---

## Not fixed: concurrent writes to attachments.yaml

`attach` and `detach` are read-modify-write with no lock. Each reads the whole
record, changes one entry, and renames a complete replacement over it. The
rename is atomic, so the file is never torn -- but a concurrent writer's changes
are gone, and nothing reports it. Five runs, eight `attach` in parallel each
time, all on distinct policies:

```
run 1: 8 parallel attach -> 4 of 8 recorded; commands reporting failure: 0
run 2: 8 parallel attach -> 4 of 8 recorded; commands reporting failure: 0
run 3: 8 parallel attach -> 6 of 8 recorded; commands reporting failure: 0
run 4: 8 parallel attach -> 6 of 8 recorded; commands reporting failure: 0
run 5: 8 parallel attach -> 5 of 8 recorded; commands reporting failure: 0
```

Every one of those commands exited 0.

I left it alone deliberately. brig does use `flock`, but only in
`cmd/brigd/main.go` to keep a second daemon from starting; no bookkeeping file
is written under a lock. `internal/wrap/index.go` has exactly this shape, and
`claimSlug` on top of it -- read the whole index, change one entry, rename a
replacement over it. Locking `attachments.yaml` alone would make it the only
guarded one of the three, and guarding the others is a decision about the whole
tool rather than about this PR. Interactive use does not hit it. A script that
attaches in a loop does.

Happy to do it as its own PR if you want it before this lands.

ananos (approved):
Approving. Both points from the earlier review are addressed, and I checked them against the built binary rather than reading the diff.

The binding now says what it is at the point of use. `attach` and `check` both print that nothing enforces it yet, which was the substance of the first finding: `check` printing a policy name and exiting 0 was where someone would come to believe a policy was in force.

```
$ brig policy check claude-code
egress
note: no policy is enforced at runtime yet; this records the binding only
```

The unparseable record names the file, and the raw struct-field text is gone. Worth saying that the exit codes are the right way round here, which I got wrong on my first pass because I was reading `rc` through a pipe: `check` and `attach` both exit 1 on a corrupt record, and `ls` warns and still lists at 0. Listing policies does not depend on the binding record, so degrading there and failing in the other two is the correct split.

The five you found on your own all hold up:

- `policy create attachments` refused, and refused under `--force` too. Forcing over brig's own bookkeeping is not the same thing as replacing a policy you own, and the `edit` rename making the same refusal closes the second way in.
- `attach -n "My Work"` refused, naming `my-work` rather than quietly recording a binding no run would ever match. That is the same refuse-rather-than-rewrite rule the session work settled on, so the two agree.
- The record reaches disk before the success line.
- A dangling binding shows as `egress (not loaded)` at exit 1, and `ls` names what it is bound to. That was the case I would have raised next if you had not.

`internal/policy` and `cmd/brig` pass here; the only failures are `internal/secret`, which fails identically on `main` in a background job with no interactive keychain.

One wording nit, not worth another round: "claude-code is bound to egress, which no policy loads under" does not quite parse. Maybe "... bound to egress, and no policy of that name loads". Up to you.

PR #122, opened 2026-09-01
feat(wrap): mount a named project beside the guest home

### LINKED ISSUE #6: Take the workspace as a positional argument on run and create
### Problem

The workspace can be set with `-w`, `--workspace` or `BRIG_WORKSPACE`, and
the docs put projects under a brig-owned home.
The usual expectation is different:

```
cd ~/src/myproject
brig run claude .
```

That line does not do what it looks like. brig's argument parser gives the
second bare word to the agent, so `.` becomes the agent's first argument.

### Proposed solution

Give `run` and `create` a second positional argument:

```
brig run <profile> [workspace] [flags] [-- agent args...]
```

Agent arguments come after `--`. The default is unchanged: with no workspace
argument and no flag, brig mounts what it mounts today.

For one release, warn when a second bare word is present and `--` is absent.
Name both readings in the warning, so a user can pick one before the meaning
changes.

### Alternatives considered

- Treat the second bare word as a workspace only when it names an existing
  directory. This keeps most current command lines working, but the meaning
  of an argument then depends on the filesystem, which is hard to explain and
  hard to script.
- Keep the flag as the only way to set a workspace. This costs nothing, but
  the current-directory workflow stays two extra words long, and that is the
  first thing a new user types.

asapranidis:
## Summary

`cd ~/src/myproject && brig run claude .` did not do what it looks like: `.` reached the agent as its
first argument. This makes it the project directory.

The reason it was not simply the workspace: the workspace **is** the guest's home, so pointing it at
a repository would put the agent's dotfiles in that repository. #47 rejected a positional home for
that reason. This adds a second mount instead — the home stays brig-owned, the project mounts beside
it, and the agent starts in the project.

## Related issues

Fixes #6. Part of #47. **Stacked on #120** — base `feat/114-drop-slug-cap`.
Merge order: #113, #115, #117, #118, #119, #120, this.

#6 and #47 were never in conflict; they argued about two different mounts. That is recorded in a
comment on the issue.

## Changes

- `brig run <ref> [project] [agent args...]`. `run` only — `sh` gets no positional, since a second
  bare word there is already the guest command.
- The project mounts at `/work/<basename>`, **outside every profile's `GuestHome`**
  (`/home/claude`, `/home/codex`, `/root/work`). That layout is the guarantee: a project cannot be
  mistaken for home state or collide with the agent's dotfiles, by construction rather than
  convention. Basename rather than the full path so the agent's prompt names the project the way its
  owner does.
- `GuestCwd` points at the project. The cwd-under-home derivation is untouched when no positional is
  given.
- `--home PATH` for the guest home. `-w` and `--workspace` keep working and print a notice.
- Running the same ref with a different project recreates the sandbox, reusing `EnsureRunning`'s
  existing stale-share path. Shares are `--shared-dir` arguments to `hull run` and the `Runtime`
  interface has no add-share method, so a project cannot be attached to a live sandbox. Recreating is
  cheap because the home is a host directory that survives it.
- The trust key is computed for the project path when one is mounted. `trustGuestCwd` pre-trusts the
  directory the agent starts in; without this the agent would start in `/work/<basename>` and prompt
  for trust on a directory brig had just handed it. Git identity does not move — it writes
  git-over-HTTPS files into the guest home, which is global config, not per-repository.
- `sessionEntry` gains `Project string \`json:"project,omitempty"\``, the field #109 deferred here.
  `omitempty` because it records the project a session last ran with rather than part of its
  identity; the identity is the ref, which is the home.
- The envelope names the project: `PROJECT  <path> (read-write, mounted at /work/<basename>)`.

### The one-release notice, and its cost

A bare word after the ref previously ended brig's parsing and reached the agent, so this reassigns
that word. The notice names both readings and points at `--`:

```
brig: `<path>` is now the project directory this run mounts, and brig starts the agent
in it. It used to be the agent's first argument instead. If that is what you meant, put
it after --: `brig run <ref> -- <path>`. This notice goes in the next release.
```

**It fires on every positional use, including intended ones.** The two readings are syntactically
identical, and #6 rejected the alternative of deciding by whether the word names an existing
directory — that makes an argument's meaning depend on the filesystem, which is hard to explain and
hard to script. So the new feature nags for one release. That is the accepted cost of not silently
changing what an existing command line does.

`brig run <ref> -- <word>` is silent and still passes the word to the agent.

### One existing command line broke, and it is in this repo

`script/smoke.sh` drove `brig run ubuntu uname -a`, which now reads `uname` as the project and
refuses the run. Changed to `brig run ubuntu -- uname -a` — same assertion text, corrected command.
This is exactly the break the notice exists for.

I scanned `README.md` and `docs/*.md` for the same shape. No documented example breaks: every
`brig run <agent> <word>` hit is `brig run <agent>` followed by a `#` comment, which is not an
argument.

## Testing

`script/smoke.sh`: 225 → 244 passing. FAIL 11 → 16.

**No regressions**: the set of assertions that passed on the parent and fail here is empty,
verified with `comm` in both directions. The suite passes in CI (`ci.yml:62`).

The 5 additional FAIL lines are new assertions added by this change, and all five need an exec
inside the guest — which the stub runtime cannot complete, on the parent commit as much as here.
CI is where they are actually exercised, and it is green.

The 19 new passing assertions cover both shares, the index entry, the notice and its absence after
`--`, reuse on the same project, recreate on a different one, the refusal of a non-directory,
`--home`, and the `-w`/`--workspace` notices.

Unit: `internal/wrap/project_test.go` (291), `cmd/brig/project_test.go` (135), plus additions to
`running_test.go` and `main_test.go`. The regression test that matters most is
`TestLoadWithoutAProjectIsUnchanged` — with no positional, the mount set and `GuestCwd` are as
before.

## Manual testing

Against a real runtime, on the `ubuntu` profile: it is `kind: shell`, its guest home is
`/root/work`, and it declares no credential files, so nothing here depends on a configured agent.

```bash
make build

export BRIG=./brig
mkdir -p /tmp/rtproj /tmp/other
git -C /tmp/rtproj init -q
echo marker > /tmp/rtproj/marker.txt
echo other  > /tmp/other/other.txt
```

**1. The mount layout.** The guest home stays brig's own; the project mounts beside it.

```bash
$BRIG run -d ubuntu@rt /tmp/rtproj
$BRIG sh ubuntu@rt ls -a /work/rtproj      # marker.txt and .git
$BRIG sh ubuntu@rt ls -a /root/work        # the guest home, no marker.txt
ls -a /tmp/rtproj                          # unchanged: no dotfile, nothing added
```

That last line is the point of the design. If a dotfile appears in `/tmp/rtproj`, the mount is
wrong however well the rest behaves.

**2. The envelope names both.**

```bash
$BRIG info ubuntu@rt
```

A `PROJECT` line giving the host path, `/work/rtproj`, and `read-write`, beside the `WORKSPACE`
line for the home.

**3. A verb that names no project inherits it.** This is the case a stub cannot reach, and the one
that was wrong until `a60b575`.

```bash
$BRIG sh ubuntu@rt pwd            # /work/rtproj
cat ~/.brig/sessions.json         # the entry still carries "project"
$BRIG ls                          # still running: no restart happened
$BRIG sh ubuntu@rt                # a login shell, also in /work/rtproj
```

Before the fix the first line printed the guest home, because the flagless verb tore the sandbox
down and rebuilt it without the mount. The index key is the better signal: it was being dropped,
which lost the mount for every later command too.

**4. Arguments pass through beside a project.**

```bash
$BRIG run ubuntu@rt /tmp/rtproj 'pwd; ls -a'
```

`/work/rtproj` and its contents. No `--` and no `-lc`: brig wraps a `kind: shell` tail as
`bash -lc <tail>` already, so the whole script goes as one quoted argument.

**5. The one-release notice.** Every command above that named a positional printed it. It fires on
every positional use, not only on the old meaning — brig cannot tell "mount this" from "pass this
to the agent", they are the same syntax, and deciding by whether the word names an existing
directory would make an argument's meaning depend on the filesystem.

```bash
$BRIG run ubuntu@rt -- 'echo an-argument-not-a-project'
```

Silent, and the string reaches the guest: after an explicit `--` the line has already said the
rest is the agent's.

**6. A different project recreates; the same one does not.**

```bash
$BRIG run -d ubuntu@rt /tmp/other     # warns that it is restarting
$BRIG sh ubuntu@rt pwd                # /work/other
$BRIG sh ubuntu@rt ls /work/other     # other.txt
$BRIG run -d ubuntu@rt /tmp/other     # same project: no restart
```

A share cannot be attached to a live sandbox, so a different project means a rebuild. The
workspace survives on the host; the guest's memory-only state does not.

**7. `--no-project` detaches, and the detach sticks.**

```bash
$BRIG run -d --no-project ubuntu@rt
$BRIG sh ubuntu@rt pwd            # /root/work
$BRIG sh ubuntu@rt ls /work       # empty or absent: the mount is gone
cat ~/.brig/sessions.json         # no "project" key for ubuntu@rt
$BRIG sh ubuntu@rt pwd            # still /root/work, one command later
```

**8. The two refusals.** Both exit 1 rather than picking a winner.

```bash
$BRIG run --no-project ubuntu@rt /tmp/rtproj    # names both the flag and the directory
$BRIG sh --no-project ubuntu@rt                 # --no-project belongs to `brig run`
```

**9. A session with no project is unchanged.** This is the regression guard for every existing
invocation.

```bash
$BRIG run -d ubuntu@plain
$BRIG sh ubuntu@plain pwd         # /root/work
$BRIG info ubuntu@plain           # no PROJECT line
cat ~/.brig/sessions.json         # no "project" key for ubuntu@plain
```

No positional notice either.

**10. The line the feature exists for.**

```bash
cd /tmp/rtproj
$BRIG run -d ubuntu@dot .
$BRIG sh ubuntu@dot pwd           # /work/rtproj
$BRIG sh ubuntu@dot ls            # marker.txt
```

The basename, resolved from `.` against the directory the command ran in.

**Cleanup.**

```bash
$BRIG rm ubuntu@rt ; $BRIG rm ubuntu@plain ; $BRIG rm ubuntu@dot
$BRIG ls                          # empty
rm -rf ~/brig/ubuntu-rt ~/brig/ubuntu-plain ~/brig/ubuntu-dot /tmp/rtproj /tmp/other
```

## Checklist

`script/smoke.sh` is ticked on CI. Locally 16 fail, of which 11 predate the branch and 5 are new
assertions blocked by the same host condition; no assertion that passed on the parent fails here.

Real runtime: not booted. Exercised against the stub, which covers the whole path except the VM.
This one is worth a real boot before merge more than the others in the stack, because it adds a mount.

Docs: README teaches the positional and `--home`. The structural rewrite is #111.

**This widens what the guest can see, and that is the point of the change.** Until now a sandbox saw
one host directory: its home. It now sees a second, the project, read-write.

- It is mounted only when named. No positional, no second mount — asserted by
  `TestLoadWithoutAProjectIsUnchanged`.
- It is mounted at `/work/<basename>`, outside `GuestHome`, so it cannot shadow or be shadowed by
  home state.
- The trust key moves with the guest cwd, so brig is not pre-trusting a directory other than the one
  the agent is actually started in.
- No new forwarded variable, and no change to credential resolution or delivery.

`docs/security.md` describes the boundary as one directory. That sentence is now incomplete; #111
owns the rewrite and this PR does not touch it.

## LLM usage

that wrote the change: `gofmt`, `make all`, the smoke sets diffed against `db5415a` in both
directions with `comm`, the notice and its absence after `--` probed against the built binary, and
`README.md` plus `docs/*.md` scanned for command lines the new grammar would break.

`Fixes: #6` sits on the first commit rather than the last; GitHub closes the issue on merge either
way. All three commits are unsigned — SSH signing does not work non-interactively in the environment
they were produced in. The `main` ruleset does not require signatures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ separate Linux re-implementation of it.
 
 | command | what it does |
 | --- | --- |
-| `brig run <ref> [args…]` | start the sandbox if needed, then run the agent. Arguments pass through untouched. `-d` starts it and exits, printing its name |
+| `brig run <ref> [project] [args…]` | start the sandbox if needed, then run the agent. A directory named after the ref is mounted as this run's project and the agent starts in it. Remaining arguments pass through untouched. `-d` starts it and exits, printing its name |
 | `brig sh <ref> [cmd…]` | a login shell inside the sandbox, or one command in it |
 | `brig stop <ref>` | stop the sandbox, keep it. Starting it again is a boot, not a fresh creation |
 | `brig rm <ref>` | stop and remove the sandbox. The workspace is untouched |
@@ -152,6 +152,11 @@ the label: `brig run claude@refactor` is what `brig run claude --name refactor`
 was. That one is mapped and warned about rather than quietly reinterpreted,
 because `--name` is also Claude Code's own flag.
 
+`-w`/`--workspace` is `--home`, on the same terms: both spellings keep working
+and each says so. The word went because there are two host directories on a run
+line now -- the guest's own home and the project you name -- and "workspace"
+could not say which of them it meant.
+
 `brig secret` keeps `create|read|update|delete`, and that is deliberate rather
 than an oversight. The tidier `set`/`get` pair would make one typo turn a read
 into a write, silently replacing a stored credential with no way back to the
@@ -162,7 +167,9 @@ of place. Left of the verb are brig's global flags, a closed set -- an unknown
 flag there is a usage error naming the token, rather than an operand quietly
 swallowed. Between the verb and the agent are the run-line flags below. Right of
 the agent the vocabulary is the agent's, and `--` ends brig's parsing outright,
-so an agent flag spelled like one of brig's still reaches it.
+so an agent flag spelled like one of brig's still reaches it. The one exception
+is `run`'s project directory, which is the second bare word on the line -- see
+[The project you name](#the-project-you-name).
 
 brig does still read its own flags after the agent, because `brig run claude -q`
 is a line that works. But once an unknown token has ended brig's reading, a brig
@@ -173,7 +180,7 @@ the token, so a line that works today keeps working.
 | --- | --- |
 | `-n, --name NAME` | a session of its own: own workspace, own sandbox. Also written `<agent>@<label>` |
 | `--image IMAGE` | guest image to boot (`-t` still works, with a note) |
-| `-w, --workspace PATH` | host directory to mount as the guest home |
+| `--home PATH` | host directory to mount as the guest home (`-w`/`--workspace` still work, each with a note) |
 | `--mem MB` | guest memory (`--memory` also works; `-m` works with a note) |
 | `--cpus N` | guest vCPUs |
 | `-d, --detach` | with `run`: start the sandbox, print its name, exit |
@@ -210,6 +217,47 @@ brig run claude -- --name not-a-session    # --name reaches claude, not brig
 brig sh codex 'ls -la ~'                   # one command, in a login shell
 brig sh codex                              # a login shell, no command
 ```
+
+### The project you name
+
+```bash
+cd ~/src/myproject
+brig run claude .
+```
+
+That line does what it looks like. The directory is mounted at
+`/work/myproject` in the guest and the agent starts there.
+
+`/work`, not the guest home. The home is the agent's own -- its dotfiles, its
+onboarding state, its caches -- so a project mounted there could be mistaken for
+home state or collide with the agent's files. Under `/work` it cannot, by the
+mount layout rather than by a convention anyone has to remember. Both mounts are
+read-write, and `brig info` and the pre-run envelope name each of them.
+
+The project is per run, not part of the session. The session is its ref, and the
+ref is the home: `brig run claude ~/src/a` and `brig run claude ~/src/b` are the
+same session with the same home, one after the other. The one consequence worth
+knowing is that the second one restarts the sandbox -- a share is bound at boot
+and cannot be attached to a live guest -- and brig says so when it does. Nothing
+persistent is lost: the home is a host directory that survives the restart, and
+so is each project.
+
+Only `run` takes one. On `sh` the second bare word is already the guest command,
+and brig's own parsing still stops at the project, so a flag after it is the
+agent's exactly as it was before.
+
+```bash
+brig run claude ~/src/myproject -p "what does this do?"  # project, then args
+brig run claude -- ~/src/myproject                      # a path FOR the agent
+brig run claude --home ~/homes/claude ~/src/myproject    # both, each named
+```
+
+**This is a breaking change, for one release.** That word used to end brig's
+parsing and reach the agent, so `brig run claude .` passed `.` to the agent.
+Anything brig now reads as a project it says so about, naming the reading it
+lost beside the one it gained, and `--` is how to keep the old one. The notice
+goes in the next release. A word that names no directory is refused rather than
+mounted, so a line that meant the agent gets an error and not a surprise.
 
 ### Named sessions
 
@@ -272,13 +320,14 @@ To drop the workspace too, remove the directory `brig info` prints. And
 `brig rm --all` stops and removes every brig sandbox at once, named ones
 included, leaving all workspaces alone.
 
-**`-w` is remembered.** A session created with `--workspace` keeps that
+**The home is remembered.** A session created with `--home` keeps that
 directory for every later verb, so `brig sh claude@refactor ls` finds it
 without repeating the flag. brig records it in `~/.brig/sessions.json`, filed
-under the session's ref, and reads it back when neither `--workspace` nor
+under the session's ref, and reads it back when neither `--home` nor
 `BRIG_WORKSPACE` names one; `brig rm` drops the entry with the sandbox, and
 `brig ls` drops any whose sandbox has gone. That file is also where `brig ls`
-reads the ref it prints.
+reads the ref it prints, and where the project a session last ran with is
+recorded -- which is how the next run knows whether it has to restart.
 
 That file is an index and not a record of the truth: the runtime stays the
 authority on what is actually mounted, and an unreadable index costs one restart
@@ -294,7 +343,7 @@ The verbs line up deliberately, so muscle memory carries over:
 
 | `sbx` | `brig` | note |
 | --- | --- | --- |
-| `sbx run <agent> [path]` | `brig run <agent> [-w path]` | brig takes the workspace as a flag rather than a positional, so it never has to guess where agent arguments begin |
+| `sbx run <agent> [path]` | same | the positional is the project brig mounts and starts the agent in; `--home` is the separate directory that is the guest's own |
 | `sbx create` | `brig run -d` | starts the sandbox and exits, printing its name |
 | `sbx ls` / `sbx stop` / `sbx rm` | same | |
 | `sbx exec` | `brig sh` | one verb for a command and for a login shell |
--- a/README.md
+++ b/README.md
@@ -181,6 +181,7 @@ the token, so a line that works today keeps working.
 | `-n, --name NAME` | a session of its own: own workspace, own sandbox. Also written `<agent>@<label>` |
 | `--image IMAGE` | guest image to boot (`-t` still works, with a note) |
 | `--home PATH` | host directory to mount as the guest home (`-w`/`--workspace` still work, each with a note) |
+| `--no-project` | with `run`: mount no project, whatever this session ran with last |
 | `--mem MB` | guest memory (`--memory` also works; `-m` works with a note) |
 | `--cpus N` | guest vCPUs |
 | `-d, --detach` | with `run`: start the sandbox, print its name, exit |
--- a/README.md
+++ b/README.md
@@ -167,14 +167,15 @@ of place. Left of the verb are brig's global flags, a closed set -- an unknown
 flag there is a usage error naming the token, rather than an operand quietly
 swallowed. Between the verb and the agent are the run-line flags below. Right of
 the agent the vocabulary is the agent's, and `--` ends brig's parsing outright,
-so an agent flag spelled like one of brig's still reaches it. The one exception
-is `run`'s project directory, which is the second bare word on the line -- see
-[The project you name](#the-project-you-name).
+so an agent flag spelled like one of brig's still reaches it.
 
 brig does still read its own flags after the agent, because `brig run claude -q`
-is a line that works. But once an unknown token has ended brig's reading, a brig
-flag further along belongs to the agent -- and brig now says so, without taking
-the token, so a line that works today keeps working.
+is a line that works. What ends brig's reading is the first token it does not
+own. On `run` that is not the project directory -- the second bare word is
+brig's own operand, so brig reads past it and on to the next token; see
+[The project you name](#the-project-you-name). Once an unknown token has ended
+brig's reading, a brig flag further along belongs to the agent -- and brig says
+so, without taking the token, so a line that works today keeps working.
 
 | flag | what it does |
 | --- | --- |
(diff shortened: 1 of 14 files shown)


ananos:
Mounting the project outside every profile's `GuestHome` is the right
structural call, and the PR body defends it well.

One case the root guard misses. It tests `filepath.Base(abs)`, and
`filepath.Abs` cleans lexically but does not resolve symlinks. So a symlink to
`/` goes straight through:

```
$ ln -s / ~/myroot
$ brig run claude ~/myroot
```

`mountProject` accepts this and sets `GuestProject=/work/myroot`. I reproduced
it with a throwaway test:

```
mountProject("/")             -> refused, good
mountProject(symlink -> "/")  -> err=<nil>, guest="/work/myroot"
```

`/Users/..` is caught correctly, since `Abs` cleans that one lexically. So the
lexical half works, it is only symlinks that get past.

To be precise about what I checked: I verified that `mountProject` accepts the
symlink. I did not boot a guest, so I have not confirmed what the VMM does with
it -- presumably it follows the link and exports the host root, but that is
worth someone checking on a real runtime before we decide how much it matters.

Nothing here is an escalation as such, since the user typed the path. But the
guard exists for the accidental case, and a `~/mnt -> /` symlink is exactly how
someone hits it by accident. A `filepath.EvalSymlinks` before the check would
cover it.

ananos:
`--no-project` is a good addition, and the three-way switch in `Load` reads
well. Keeping "said nothing" and "asked for none" apart is the right
distinction to draw.

There is a problem underneath it though, and it is the positional rather than
the flag. The project path is a bare word, and #115's rule is that the first
token brig does not own ends brig's reading. So every brig flag written after
the path goes to the agent:

```
$ brig run claude /tmp/projx --mem 4096
brig: --mem is one of brig's own flags, but here it is the agent's ...
```

Same for `-d`, `--offline` and `--no-project` -- I checked all four. So
`brig run claude ~/src/myproj --mem 4096 -d --offline`, which is the natural
shape of the line this feature exists for, boots in the foreground with default
memory and the network on. brig warns, but it does not act.

That rule made sense when the second bare word was the agent's command:
`brig run claude -p hi --quiet` really is all the agent's. For a project
directory I think it is harder to defend, since the path is brig's own operand
rather than the start of the agent's argv.

It also makes the new contradiction check unreachable in the spelling people
will type:

```
$ brig run --no-project claude /tmp/projx     # flag before the ref
brig: --no-project and the project "/tmp/projx" ask for different things; drop one

$ brig run claude /tmp/projx --no-project     # flag after the path
# no error: the project is mounted, and --no-project goes to the agent
```

`TestNoProjectWithAPositionalIsRefused` only covers the first spelling, which
is presumably why CI is green on this.

What I would suggest: on a `run` line, the project positional should not end
brig's reading, so brig keeps reading its own flags after it the way it already
does after the ref. Another approach would be to move the contradiction check
into the tail, so it can see a `--no-project` that landed there. The first one
seems cleaner to me, but up to you.

All of the above was checked at parse level against the built binary with
`BRIG_RUNTIME` stubbed out, so no guest was booted on my side.

asapranidis:
Fixed in db3ce40. Reproduced it first — `mountProject(<link> -> "/")` returned
nil and set `GuestProject=/work/myroot`, exactly as you had it.

One reframing, because it changes what the fix is for. This is not only the
accidental case. The guard is a boundary, and a symlink is a general way around
*any* check brig makes on a directory — the person who follows a link is not
necessarily the person who made it, and a planted `mnt -> /` is a way to get a
mount someone did not intend to grant. That makes it worth closing on its own
terms rather than as a nicety for people with tidy home directories.

`EvalSymlinks` now runs before the check, as you suggested. One deviation:
it feeds the **check only**, not `Project`/`GuestProject`. Resolving those too
looked right and isn't — it broke seven existing tests, and the breakage was the
useful part: on macOS every path under `/tmp` and `/var` resolves elsewhere, so
`brig run claude /tmp/x` would start reporting and recording `/private/tmp/x`,
and every session whose project sits under a symlinked parent would see a
changed path and restart. A link is also the name its owner uses for the
project, so `/work/<link>` is the right guest path. The VMM resolves the host
side at mount time regardless.

So: resolved for the judgement, as typed for the mount.

```
$ brig run claude ~/myroot          # myroot -> /
brig: cannot mount /Users/…/myroot as this run's project: it resolves to /,
      which has no name to mount it under; name a project directory rather
      than a filesystem root

$ brig run claude /
brig: cannot mount / as this run's project: it has no name to mount it under; …

$ brig run claude ~/proj            # proj -> ~/src/thing
PROJECT   /Users/…/proj (read-write, mounted at /work/proj)
```

The resolution is named only when there is one worth naming — "/ resolves to /"
would have been noise on the direct case.

A path that stats but cannot be resolved is refused rather than falling back to
the lexical check, since the lexical check is precisely the one that does not
hold for it.

Two things I did not do, both worth saying out loud rather than leaving implied.
I have not booted a guest either, so your presumption about what the VMM does
with the link is still unverified — the fix means brig never hands it one, which
is the part that is now settled. And a residual TOCTOU window remains: a link
repointed between this check and the VMM's mount would still be followed.
Closing that means mounting the resolved path, which is the change I backed out
above; I would rather leave it open and named than pay that cost silently. Say
if you disagree — it is a real gap, just a much narrower one than the hole you
found.

The guard is still filesystem-root-only. A link to `/Users` or to `$HOME` is
accepted, exactly as typing those directly is.

`TestALinkIsResolvedBeforeTheRootGuardAndMountedUnderItsOwnName` pins both
halves: a link to `/` is refused, and an ordinary link mounts under its own
name. It fails on the parent commit.

`gofmt` clean, `make all` green, `script/smoke.sh` unchanged against the parent
(16 host-dependent failures both sides).

asapranidis:
Fixed in 8ca34d1, taking your first option.

One correction to the framing. This is not an exception to #115's rule — it is
that rule applied. `split`'s doc comment already says the boundary is "the first
token it does not own", and on a `run` line the positional is a token brig owns,
like the ref before it. It ended the parsing only because `split` did not know
the verb, which the same comment says two paragraphs earlier. So `split` now
takes the verb, the positional stops being a boundary, and `projectFor` is gone
because that is now the same decision.

Parsing after the positional ends where it ends after the ref: at the next bare
word, which is where the agent's argv starts, or at the first flag brig does not
own.

```
brig run claude /tmp/p --mem 4096 -d --offline   → brig's, all four
brig run claude /tmp/p --resume                  → agent's
brig run claude /tmp/p npm test                  → agent's
brig run claude /tmp/p -d npm test               → -d brig's, npm test agent's
brig run claude /tmp/p --no-project              → refused
brig sh claude ls -la                            → unchanged, guest's
```

Two things found on the way that were not in your report.

`--` discarded a project named before it. `brig run claude ~/proj -- pwd` lost
the mount, because the `--` branch returned an empty positional — unreachable
before, since `--` could only be met before the second bare word. It now keeps
what was already read: the marker speaks for what follows it, not for what came
before. Caught by `script/smoke.sh`, not by the unit tests.

The breaking change you would expect is real and I did not avoid it: `brig run
claude . -q` read `-q` as the agent's and now reads it as brig's. Mitigating it
— the word before it is already changing meaning this release with a notice,
that line already printed a warning saying brig had not read the flag, and `--`
still hands both to the agent. Called out in the commit body.

Tests: `TestBrigFlagsAfterTheProjectAreStillBrigs` covers the four shapes above
plus the boundary cases, `TestNoProjectAfterThePositionalIsRefused` covers both
spellings of the contradiction, and `TestOnlyRunKeepsTheSecondBareWord` now
asserts through `split` per verb rather than through `projectFor`.

README updated: the flag-positions section no longer calls the positional an
exception, and the project section states the boundary.

Verified beyond the stub, on the `ubuntu` profile with a real runtime, since
the stub missed a mount bug on this PR once already:

```
$ brig run ubuntu /tmp/…/myproject --mem 2048 'pwd; ls'
Starting VMM: … --mem-mib 2048 … --share-rw /tmp/…/myproject share1 …
/work/myproject
marker
$ brig sh ubuntu 'pwd; ls'      # flagless verb still inherits it
/work/myproject
marker
```

`gofmt` clean, `make all` green

ananos (approved):
Both findings are fixed. I verified them on 181cfcf against the built binary
rather than from the diffs.

Symlink guard: a link to `/` is refused, an ordinary link still mounts under
its own name at `/work/<link>`.

Flags after the positional: `--mem 4096`, `-d` and `--offline` all reach brig
and land in the options, `--resume` and `npm test` still go to the agent, `-d
npm test` splits correctly, `brig sh claude ls -la` is unchanged, and a project
named before `--` survives. That last one was your catch, not mine.

On the TOCTOU window you named: I agree with leaving it open. Mounting the
resolved path costs the `/tmp` -> `/private/tmp` rewrite on macOS and a restart
for every session under a symlinked parent, which is worse than the window it
closes. Named and open is the right call.

Two minor things I am not blocking on, recorded so they are not lost:

1. Share specs are built as `host + ":" + guest` in both `hull.go` and
   `nerdctl.go`. The project path is user input now, so a directory with a `:`
   in its name gives `/…/my:proj:/work/my:proj`. Pre-existing and reachable
   before via `--workspace`, just easier to hit now.

2. `brig create claude --no-project` is refused, though `create` is documented
   as exactly `run -d`. Goes away when `create` does.

pmoust:
@claude-fleet review

PR #128, opened 2026-09-02
test(smoke): make the smoke test pass on a developer's machine

asapranidis:
## Summary

`script/smoke.sh` was broken on macOS in two independent ways. Neither can
happen on a Linux CI runner, so both went unnoticed.

**Cases failed from the first `brig run` onwards.** The stub's `stat -f -c %T`
grepped its mount table for the exact path it was given and answered `virtiofs`
on a miss. A file inside a tmpfs has no mount entry of its own, so the stub
called `/home/claude/.claude/.credentials.json` `virtiofs` where a real guest
says `tmpfs` -- exactly the fstype `verifySecretFile` refuses rather than put a
credential on host disk. The first `brig run` exited 1 and everything downstream
of a booted sandbox failed with it. brig is right here; the stub was lying.

Needs a credential to deliver, so only macOS hits it: `claude-credentials` is
`required: false`, sourced from the macOS keychain or
`~/.claude/.credentials.json`. A Linux runner has neither and skips the whole
path.

**Then it hung.** From a terminal, the script stopped after
`ok an older hull boots the tag` and waited on a keypress with no prompt in
sight. The next case drives a bad signature, which under `BRIG_VERIFY=warn` asks
instead of refusing. `Config.confirm` checks brig's own stdin --
`out="$(brig ... 2>&1)"` redirects stdout and stderr but *not* stdin, so brig
found the terminal, wrote `Boot it anyway? [y/N]` into the captured stream where
nobody could see it, and blocked. On a runner there's no tty, `confirm` refuses
immediately, exit 5 -- which is what the case asserts. Its label already said
"with no terminal to ask"; nothing arranged for that, it was just true by
accident.

## Related issues

None.

## Changes

- Stub `stat -f` walks up to the nearest recorded mount point instead of
  matching the exact path -- what the kernel does, and what `guestFake.fstypeOf`
  in `internal/wrap/secretfiles_test.go` already did. The walk matches on field
  5, the mount point: field 4 is the mount's root and reads `/` on every line,
  so a whole-line match would hand a path under no mount whichever fstype was
  recorded last, rather than the `virtiofs` fallback. `fuseblk` for bind mounts
  and the `virtiofs` fallback are unchanged, and no assertion relied on the old
  behaviour.
- `exec < /dev/null` at the top of the script. Central rather than per call: the
  no-terminal assumption belongs to the whole script, and cases added later
  inherit it. The two `profile import -` cases pipe into brig and keep their own
  stdin.

Harness only, no production code.

## Checklist

- [ ] ~~I have added or updated tests covering the change~~ -- this *is* the test
- [ ] ~~I have run `go test ./... -race`~~ -- no Go code changed
- [ ] ~~I have exercised the change against a real runtime~~ -- stub only
- [ ] ~~I have updated the affected docs~~ -- no documented behaviour changed

## Testing

Rebased onto `c06f9bc`; both commits applied cleanly. macOS arm64, where both
failures reproduce:

| | before | after |
|---|---|---|
| stdin not a tty | `FAILURES` | `PASS`, 269 assertions, exit 0 |
| real pty on stdin | hung; killed at 60s | `PASS`, exit 0 in 15s |

Failure counts are host-specific -- they depend on having a credential to
deliver -- so the table says what the run does, not how many cases it takes with
it. The hang was reproduced under a `pty.openpty()` stdin first and re-run after
the fix, not inferred from reading `confirm`. Also runs clean under the system
bash 3.2 (`/bin/bash`), not just bash 5. `make all` green.

The mount-point anchoring was checked directly: with mounts
`/home/claude/.claude tmpfs` and `/home/claude fuseblk` recorded in either
order, `.credentials.json` answers `tmpfs`, `work.txt` answers `fuseblk`, and
`/nonexistent` and `/home` answer `virtiofs`. `/`, `//`, `/a` and `/a/b/c/d/e`
all terminate.

## LLM usage

Authored with assistance from Claude Code; I have reviewed every line and am
accountable for the change.

### DIFF
--- a/script/smoke.sh
+++ b/script/smoke.sh
@@ -95,8 +95,23 @@ case "$verb" in
         # `stat -f -c %T <path>` asks what a path sits on; `stat -c ...` asks
         # about the file itself.
         if [ "${2:-}" = -f ]; then
+          # A path inherits the filesystem of the nearest mount point above
+          # it, so walk up until one matches rather than asking only about the
+          # path itself: a credential file created inside a tmpfs has no mount
+          # entry of its own, and answering virtiofs for it is a fstype the
+          # guest would never report. Match on field 5, the mount point --
+          # field 4 is the mount's root and reads `/` on every line, so a
+          # looser match answers for a path under no mount at all.
           target="${5:-}"
-          line=$(grep " $target " "$STUB_STATE.mounts" 2>/dev/null | tail -1)
+          line=""
+          probe="$target"
+          while [ -n "$probe" ]; do
+            line=$(awk -v p="$probe" '$5 == p' "$STUB_STATE.mounts" 2>/dev/null | tail -1)
+            [ -n "$line" ] && break
+            [ "$probe" = / ] && break
+            probe="${probe%/*}"
+            [ -z "$probe" ] && probe=/
+          done
           case "$line" in
             *tmpfs) printf 'tmpfs\n' ;;
             *fuseblk) printf 'fuseblk\n' ;;
--- a/script/smoke.sh
+++ b/script/smoke.sh
@@ -10,6 +10,17 @@
 set -uo pipefail
 cd "$(dirname "$0")/.."
 
+# Nothing here is answered by a person. Several cases assert the outcome brig
+# reaches when it has nobody to ask -- a bad signature refused, a profile rm
+# that will not guess -- and brig decides that by looking at its own stdin.
+# `out="$(brig ... 2>&1)"` redirects stdout and stderr and leaves stdin alone,
+# so run from a terminal those same cases found one, printed a [y/N] into the
+# captured stream where nobody could see it, and blocked on the answer. CI has
+# no terminal and never hit it. Detach stdin once, here, so the script behaves
+# the same either way; the two `profile import -` cases feed brig through a
+# pipe and supply their own.
+exec < /dev/null
+
 WORK="$(mktemp -d)"
 trap 'rm -rf "$WORK"' EXIT
 export STUB_LOG="$WORK/argv.log"


ananos:
Both failures are real and the diagnosis is right in each case. I reproduced the
hang and checked the stub change against current `main`.

**The hang reproduces.** Under a real pty on stdin, `script/smoke.sh` on `main`
was still running after 100s and I had to kill it. With both commits applied it
finishes in 18s, exit 0. So `exec < /dev/null` is doing what the body says.

**The stat -f change is right, and there is a second implementation that agrees
with it.** `guestFake.fstypeOf` in `internal/wrap/secretfiles_test.go:159` already
walks up to the nearest mount, with the comment "the way stat -f does", and
`secretfiles_test.go:281` asserts `/home/x/.claude/.credentials.json` is `tmpfs`.
The shell stub was the odd one out, so this brings the two harnesses into
agreement rather than moving the goalposts.

**No assertion changed.** Applied to current `main`, the suite still passes with
269 assertions and the list is identical, none added, removed or renamed.

Two things before this lands.

**1. The branch is 21 commits behind `main`.** It was cut from `8340bb4`, before
the #47 stack merged, so its `script/smoke.sh` is the pre-stack version. Both
commits cherry-pick onto `c06f9bc` cleanly and the suite passes there, so this
is a rebase rather than a rework. Worth doing because the numbers in the
description were measured against a script that no longer exists: `main`'s
smoke.sh now has 269 assertions where the branch's has 147, so "11 cases failed"
is not a count anyone can reproduce against `main` today.

I could not reproduce the 11 failures myself. On this machine `main` passes
non-interactively, 269/269, with `~/.claude/.credentials.json` present. So I can
confirm the stub change is correct on its own terms but not the failure count.

**2. The walk-up matches the wrong field once it reaches `/`.**

The mount stub writes mountinfo lines, where field 4 is the root and field 5 the
mountpoint:

```
1 1 0:1 / /home/claude/.claude rw - tmpfs
```

Every line therefore contains `" / "`, so when the loop shortens `probe` to `/`
the `grep " $probe "` matches on that root field and `tail -1` takes whichever
mount was recorded last. A path under no mount gets that fstype instead of the
`virtiofs` fallback:

```
mounts: /home/claude/.claude tmpfs, then /home/claude fuseblk
  /nonexistent -> fuseblk

same path, mounts file written in the other order
  /nonexistent -> tmpfs
```

So the answer depends on mount order, and the description's "the `virtiofs`
fallback [is] unchanged" does not hold. Nothing asserts on such a path today, so
it is latent, but it is the same class as the bug being fixed: a stub that
answers something the guest would not.

Anchoring on the mountpoint field fixes it:

```sh
line=$(awk -v p="$probe" '$5==p' "$STUB_STATE.mounts" 2>/dev/null | tail -1)
```

With that, `/home/claude/.claude/.credentials.json` is `tmpfs`,
`/home/claude/work.txt` is `fuseblk`, and `/nonexistent` and `/home` fall back to
`virtiofs`.

Termination is fine either way: `probe` shrinks each pass and the `/` case
breaks, and I checked `/`, `//`, `/a` and `/a/b/c/d/e` all return.

asapranidis:
Both points addressed.

**1. Rebased onto `c06f9bc`.** Both commits applied cleanly, no conflicts. The
description's numbers were measured against the pre-#47 script, so I re-measured
against the rebased branch: 269 assertions, `PASS`, exit 0, both under bash 5
and the system bash 3.2. Under a real pty it finishes in 15s where `main` was
still running when I killed it at 60s.

I dropped the "11 cases failed" line rather than substituting a new count. The
number depends on the host having a credential to deliver, which is why you saw
269/269 on `main` and I see failures -- so it is not a figure anyone can check.
The description now says what the run does, not how many cases it takes down.

**2. The `/` match was real, and it is fixed.** Reproduced it standalone before
changing anything, with the two mounts recorded in each order:

```
mounts: /home/claude/.claude tmpfs, then /home/claude fuseblk
  /nonexistent -> fuseblk     /home -> fuseblk

same mounts, other order
  /nonexistent -> tmpfs       /home -> tmpfs
```

Order-dependent, and never the `virtiofs` fallback -- the description's claim
that the fallback was unchanged did not hold.

Anchored on field 5 as you suggested:

```sh
line=$(awk -v p="$probe" '$5 == p' "$STUB_STATE.mounts" 2>/dev/null | tail -1)
```

With that, both orders agree: `.credentials.json` is `tmpfs`, `work.txt` is
`fuseblk`, `/nonexistent` and `/home` are `virtiofs`. `/`, `//`, `/a` and
`/a/b/c/d/e` all return, and a recorded `/` mount is now reachable where the
loop would previously have matched it by accident on every path.

Folded into the walk-up commit rather than added on top -- it is the same
commit's concern, and the branch was being force-pushed for the rebase anyway.
The commit message now records why the match is anchored where it is. `make all`
green.

ananos (approved):
Both fixes verified independently, and the branch is current. Approving.

I re-ran my own probes rather than reading the diff.

**The root-field match is gone.** I extracted the stub verbatim from this branch's `smoke.sh` and drove it with the two mounts recorded in each order:

```
order A (tmpfs child first)          order B (fuseblk parent first)
  .credentials.json -> tmpfs           .credentials.json -> tmpfs
  work.txt          -> fuseblk         work.txt          -> fuseblk
  /nonexistent      -> virtiofs        /nonexistent      -> virtiofs
  /home             -> virtiofs        /home             -> virtiofs
```

Both orders agree, and `/`, `//` and `/a/b/c/d/e` all terminate. A `/` mount recorded on purpose is now reachable on purpose: with one in the table, `/nonexistent` reads `fuseblk` instead of falling through.

**The hang.** Under a real pty this branch is 269 assertions, 0 FAIL, exit 0, in 15.5s, the same under bash 5 and the system bash 3.2. A fresh control on `c06f9bc` in the same harness hung at 247 assertions and I killed it at 240.8s.

Good call on dropping the "11 cases failed" line rather than substituting a new one. It matches what I see from here -- 269/269 on `main`, failures on yours -- so it was never a figure anyone else could check.

One thing worth flagging for merge order now that #129 is open. That branch comes off `main` and inherits this hang: I killed it at 240.1s at the same case (`an older hull boots the tag`). With this branch merged into it, the pair gives 300 ok / 0 FAIL in 19.3s under a pty. So this one wants to go in first.

Issue #10, opened 2026-08-25
Print the execution envelope before the agent starts, and name it `brig inspect`


pmoust:
### Problem

`brig run` starts an agent without saying what boundary the user is about to
trust. The facts exist: `Status()` already reports the workspace, the sandbox,
the runtime, the image, the pull policy, the forwarded credential names and
the guest login state. Only `brig env` prints
them, and a user runs `brig env` after something has already gone wrong.

The name is part of the problem. The bug report template tells reporters to
run `brig status`, which is not a command.

### Proposed solution

1. Print a short block before the agent starts:

```
SESSION      review
PROFILE      claude-code
SANDBOX      brig-claude-review (hull)
WORKSPACE    /Users/panos/src/brig (read-write)
IMAGE        ghcr.io/brig-sh/claude-code@sha256:... (verified)
CREDENTIALS  claude-credentials, GH_TOKEN
```

2. Add `brig inspect <session>` for the full report, with `--json`. Keep
   `brig env` as an alias for one release, through `deprecated()`.
3. Add `--quiet` to suppress the block, for scripts and for the second run of
   the day.

This is display only. The block prints what brig already resolves. It does not
add network policy, isolation policy or a new check. A field brig cannot
determine today is left out rather than guessed.

### Alternatives considered

Print the block only on the first run of a session. It is quieter, but the
facts change between runs: a token expires, a profile is edited, an image is
pulled again. A block that appears sometimes is a block nobody reads.

PR #129, opened 2026-09-03
feat(cli): say less by default, and put the machinery behind --verbose

### LINKED ISSUE #24: Say less by default, and put the machinery behind `--verbose`
### Problem

A plain `brig run` prints the machinery of three layers before the agent says
anything.

- brig's own progress, for example `brig: starting sandbox brig-claude...`.
- The runtime's output, passed straight through: hull's stdout and stderr go
  to the terminal, nerdctl's stderr does the same, and the boot-asset download
  prints its own progress.
- brig's advice. There are 43 places in the wrapper library that print a
  warning or a status line.

Every line is true and useful once. Together they read as work being narrated,
and the lines that matter, an expired credential or an unverified image, sit
in the same column as a pull progress bar.

### Proposed solution

Set one rule: **by default, print what the user has to act on. Print the rest
under `--verbose`.**

1. Default output on `run`: the envelope from #10, then the agent. Nothing
   else unless something needs a decision or a fix.
2. `-v`, `--verbose`: brig's own progress lines, and the runtime's output.
3. `-q`, `--quiet`: identifiers and errors only, for scripts.
4. Capture the runtime's output instead of passing it through. Print it on
   failure, where it is evidence, and under `--verbose`, where it was asked
   for.
5. Warnings stay in the default output. A warning is an action.
6. One long operation deserves one line, not a stream: a first pull says it is
   pulling, and says when it is done.

Audit the message sites once against the rule, and give each one a level.

### Alternatives considered

- Add `--quiet` and leave the default as it is. It helps scripts and leaves
  the human with the noisy path.
- Route everything through a log level environment variable. It is the usual
  answer, and it makes the quiet path something a user has to discover.

None. This changes what is printed, not what is done. The rule cuts the other
way for security-relevant facts: the envelope becomes more visible when the
progress noise around it is gone.

asapranidis:
## Summary

A default `brig run` printed brig's progress, the runtime's output and brig's warnings
together, so an expired credential sat in the same column as a pull progress bar.

Three levels now. A default run prints warnings and identifiers. `--verbose` adds the
execution envelope, brig's progress and the runtime's output. `-q` drops ordinary
warnings. Verification prints at every level, including `-q`.

| | envelope | verification problems | other warnings | progress, runtime output | identifiers, errors |
| --- | --- | --- | --- | --- | --- |
| `-q` | no | **yes** | no | no | yes |
| default | no | yes | yes | no | yes |
| `--verbose` | yes | yes | yes | yes | yes |

## Related issues

Fixes #24. Part of #47. Branches from `main`; no stacking.

## Changes

**`wrap.Verbosity`** is `Quiet(-1)/Normal(0)/Verbose(1)`. `Normal` is the zero value, so a
hand-built `Config` is unchanged. `warnf` requires `>= Normal`, `progressf` requires
`>= Verbose`, `alertf` has no level check.

**The runtime's output is captured, not passed through.** `RunSpec.Progress` decides:
`nil` buffers it and prints it with the error if the boot fails, discards it if the boot
succeeds; a non-nil writer streams it live. Capped at 64K, keeping the tail.
`Replace` is untouched — it execs the runtime over this process, so there is nothing to
capture and nothing buffered in front of an interactive agent.

**`--verbose` and `-q` take the global position** that #108 established closed and empty.
`--verbose` has no short form: `-v` is Claude Code's version flag, codex's verbose flag and
Docker's volume flag.

**`-q` moved from the run line.** `brig run claude -q` works today, so both positions are
accepted this release; the run-line spelling prints a notice and v0.3 drops it.
`brig ls -q` is unchanged.

**Verification is above the level `-q` can silence.** It used `warnf`, so `-q` suppressed
it and a run that booted an unchecked image told a script nothing. `alertf` covers
`BRIG_VERIFY=off`, every signature refusal, gap, mismatch, `NotOurs`, `NoTooling`,
`Unresolved`, both boot-asset failures, the no-terminal refusal, and `cannot boot by
digest`. Credential expiry, workspace notices and deprecations stay at `warnf`.

**A `VERIFY` envelope row and a result line.** Moving the per-check success lines behind
`--verbose` left a default run silent about verification. `PrintPreRunEnvelope` runs before
`EnsureRunning`, so the mode is known at envelope time and the outcome is not: the row
carries the mode under `--verbose`, and `sayVerified` prints the result at default level.
Nothing is printed when nothing was positively checked.

**The envelope moves behind `--verbose`.** Nine rows, and the largest part of a default
run. This reverses part of #10 and part of #24's own text, on a decision taken during
review. `brig info` prints it without booting and is unaffected at every level.

## Behaviour change worth flagging

A successful run can now print nothing. On a warm sandbox, `brig run -d` prints only the
sandbox name and `brig sh` prints nothing at all — verification does not re-run, and the
envelope is behind `--verbose`. That follows the table, and it is the first time brig is
silent on success.

## Testing

`script/smoke.sh`: 253 → 284 passing assertions, FAIL set identical to `c06f9bc`.

The 16 failures predate this branch. They are the stub `stat -f` defect fixed by #128; with
that branch applied I measured this one at **288 ok / 0 FAIL**.

Unit tests cover: the levels; each verification outcome a fake cosign can drive, all at
`Quiet`; an ordinary warning still suppressed by `-q`; the envelope's presence per verb and
level; `brig info` at every level. A source scan of `verify.go` fails on any `c.warnf`
outside `sayVerified`, so a verification site added later cannot land at the suppressible
level.

## Manual testing

```bash
make build
export BRIG=./brig
```

**The three levels.** Any profile; `ubuntu` needs no credentials.

```bash
$BRIG run -d ubuntu              # warnings, then the sandbox name
$BRIG --verbose run -d ubuntu    # envelope, brig's progress, the runtime's output
$BRIG -q run -d ubuntu           # the sandbox name
```

**`-q` drops ordinary warnings but not verification.**

```bash
GH_TOKEN='op://x/y' $BRIG run -d ubuntu      # warns about the unresolved reference
GH_TOKEN='op://x/y' $BRIG -q run -d ubuntu   # silent
BRIG_VERIFY=off $BRIG -q run -d ubuntu       # still reports that nothing was checked
```

**Flag positions.**

```bash
$BRIG -v run ubuntu          # refused, naming -v
$BRIG --verbose run ubuntu   # accepted
$BRIG -q run ubuntu          # accepted
$BRIG run ubuntu -q          # works, prints a notice naming the global position
$BRIG ls -q                  # refs only, unchanged
```

**A failed boot still reports what the runtime said.** Point `BRIG_RUNTIME_BIN` at a script
that writes to stderr and exits non-zero:

```bash
$BRIG run -d ubuntu          # brig's error, then the runtime's last lines, indented
$BRIG -q run -d ubuntu       # same: an error is not suppressed
```

On a successful boot the same output is discarded. Check both directions.

**Verification, on a profile with a signed image and cosign installed.**

```bash
$BRIG run -d claude              # "image and boot assets verified"
$BRIG -q run -d claude           # no result line
$BRIG --verbose run -d claude    # VERIFY row, result line, per-check detail
BRIG_VERIFY=off $BRIG run -d claude   # reported once, not twice
```

**`brig info` is unaffected.**

```bash
$BRIG info claude                # full envelope including the VERIFY row
$BRIG -q info claude             # same
```

It does not boot, so it shows the mode and no result line.

## Checklist

`script/smoke.sh` is ticked on CI, where it runs clean; the local failures are the #128 stub
defect described above.

Real runtime: not booted. Verified against the stub runtime, which covers the whole path
except the VM.

Docs: README and `docs/quickstart.md` updated for the flags and the new default output.

No new mount, no new forwarded variable, no new host path reachable from the guest.

This changes what is printed about the boundary, in both directions. A default run no longer
prints the envelope, which named the sandbox, workspace, image, credentials and network.
Against that, verification is now unsuppressible: `-q` previously silenced every signature
refusal and `BRIG_VERIFY=off`, so a script could boot an unchecked image and see nothing.
`brig info` remains the full report and is unchanged.

Credential values are not printed at any level, as before.

## LLM usage

Authored with assistance from Claude Code (Opus 5). Verification was re-run independently of
the agent that wrote the change: `gofmt`, `make all`, each commit built and tested in
isolation, and every command in the Manual testing section exercised against the built
binary. The level table and the failed-boot behaviour were checked in both directions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,12 +170,25 @@ the agent the vocabulary is the agent's, and `--` ends brig's parsing outright,
 so an agent flag spelled like one of brig's still reaches it.
 
 brig does still read its own flags after the agent, because `brig run claude -q`
-is a line that works. What ends brig's reading is the first token it does not
-own. On `run` that is not the project directory -- the second bare word is
+is a line that works -- `-q` has moved left of the verb, and the old spelling
+keeps working for this release and says so. What ends brig's reading is the
+first token it does not own. On `run` that is not the project directory -- the second bare word is
 brig's own operand, so brig reads past it and on to the next token; see
 [The project you name](#the-project-you-name). Once an unknown token has ended
 brig's reading, a brig flag further along belongs to the agent -- and brig says
 so, without taking the token, so a line that works today keeps working.
+
+| global flag | what it does |
+| --- | --- |
+| `--verbose` | brig's own progress and the runtime's own output. No short form: `-v` is Claude Code's version flag, codex's verbose flag and Docker's volume flag, so brig does not claim the letter |
+| `-q, --quiet` | identifiers and errors only, for a script. It drops the execution envelope and brig's warnings; on `ls` it prints the refs alone. Still read after the verb for this release, with a note |
+
+By default a run prints the execution envelope, then anything you have to act
+on, then the agent. brig's own progress lines and the runtime's own output wait
+for `--verbose` -- except that a boot which fails quotes what the runtime said
+whether or not you asked, because that is the only evidence there is. A long
+download gets one line saying it started and one saying it finished, rather than
+a stream.
 
 | flag | what it does |
 | --- | --- |
@@ -486,7 +499,7 @@ brig:   GH_TOKEN(secret)
 The block at the top is the execution envelope: the boundary the run would
 trust, printed before the sandbox boots. `brig run` prints the same block before
 it starts one, so what you preview is what you get. Names only, never values.
-`--quiet` drops it.
+`brig -q` drops it.
 
 A secret a profile declares but the store does not have fails the run before
 any sandbox is created, naming every one that is missing and the command that
--- a/README.md
+++ b/README.md
@@ -180,15 +180,23 @@ so, without taking the token, so a line that works today keeps working.
 
 | global flag | what it does |
 | --- | --- |
-| `--verbose` | brig's own progress and the runtime's own output. No short form: `-v` is Claude Code's version flag, codex's verbose flag and Docker's volume flag, so brig does not claim the letter |
-| `-q, --quiet` | identifiers and errors only, for a script. It drops the execution envelope and brig's warnings; on `ls` it prints the refs alone. Still read after the verb for this release, with a note |
+| `--verbose` | the execution envelope, brig's own progress and the runtime's own output. No short form: `-v` is Claude Code's version flag, codex's verbose flag and Docker's volume flag, so brig does not claim the letter |
+| `-q, --quiet` | identifiers and errors only, for a script. It drops brig's warnings; on `ls` it prints the refs alone. A verification that did not hold is printed even here. Still read after the verb for this release, with a note |
 
-By default a run prints the execution envelope, then anything you have to act
-on, then the agent. brig's own progress lines and the runtime's own output wait
-for `--verbose` -- except that a boot which fails quotes what the runtime said
-whether or not you asked, because that is the only evidence there is. A long
-download gets one line saying it started and one saying it finished, rather than
-a stream.
+By default a run prints what you have to act on, then the agent: warnings,
+errors, and one line saying verification held. The execution envelope, brig's
+own progress lines and the runtime's own output wait for `--verbose` -- except
+that a boot which fails quotes what the runtime said whether or not you asked,
+because that is the only evidence there is. A long download gets one line saying
+it started and one saying it finished, rather than a stream. `brig info` prints
+the envelope on demand, without booting anything.
+
+Verification is the exception at both ends. A check that did **not** hold --
+a signature that failed, a bundle brig does not publish, a missing `cosign`,
+`BRIG_VERIFY=off` -- prints at every level, `-q` included: it is the one claim
+brig exists to make, and a scripted run is where an unchecked image matters
+most. A check that held prints one line, which `-q` drops. The policy those
+held under is the envelope's `VERIFY` row.
 
 | flag | what it does |
 | --- | --- |
@@ -491,6 +499,7 @@ SANDBOX      brig-claude-code (hull)
 ISOLATION    microVM (hull, vz backend)
 WORKSPACE    /Users/you/brig/claude-code (read-write)
 IMAGE        ghcr.io/brig-sh/claude-code-stock:latest (pull missing)
+VERIFY       warn, against brig's own trust policy
 CREDENTIALS  GH_TOKEN
 brig: forwarding to guest:
 brig:   GH_TOKEN(secret)
@@ -520,7 +529,7 @@ repository *and* the workflow file, so a signature from anywhere else fails.
 
 | situation | what happens |
 | --- | --- |
-| image under `ghcr.io/brig-sh/`, signature verifies | one line saying so, boots |
+| image under `ghcr.io/brig-sh/`, signature verifies | one line saying so, boots. `--verbose` adds the per-check detail |
 | image published by someone else | **warning**, boots. Bring-your-own images are a supported way to use brig |
 | `cosign` not installed | **warning**, boots. "Could not check" is not "failed" |
 | image under `ghcr.io/brig-sh/`, signature does **not** verify | **stops and asks** `[y/N]`. With no terminal it refuses |
--- a/cmd/brig/hygiene_test.go
+++ b/cmd/brig/hygiene_test.go
@@ -233,10 +233,12 @@ func TestBrigFlagsOverlapWithClaudeCodeOnlyWhereKnown(t *testing.T) {
 	}
 }
 
-// The global position -- left of the verb -- is closed. #24 filled it with
-// --verbose and -q; everything else is still a token to name rather than a
-// command to run or a word to forward, and #11 and #30 are what fill in the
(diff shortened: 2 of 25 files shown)


ananos:
Nice piece of work, and the security half of it is a real improvement -- `-q` silencing every signature refusal was worse than I had realised.

I checked the load-bearing claims rather than taking them on trust.

**Verification is genuinely unsuppressible.** `alertf` has no level check, every refusal in `verify.go` uses it, and `sayVerified` is the only `warnf` left. To see whether the source scan actually bites I turned the `verifyTag` refusal at `verify.go:102` into a `warnf`:

```
--- FAIL: TestVerifySitesDoNotReachForTheWarningLevel (0.00s)
    alert_test.go:164: verify.go:102: verifyTag uses warnf, which -q silences.
```

The `fn == ""` guard at the end is the right instinct. A scan that quietly matches nothing is the usual failure mode for that kind of test.

**The failed boot reports.** With a stub runtime that exits non-zero, both the default level and `-q` quote the tail back indented, `FATAL: no space left on device` and all. `make all` is green here except `internal/secret`, which fails identically on `main` (no interactive keychain in a background job), and `go test ./... -race` finds nothing. `cmd.Stdout, cmd.Stderr = said, said` is the same interface value, so `os/exec` reuses one pipe and there is only ever one writer into the buffer.

## One thing I would fix before this lands: a cold run is silent

The image pull happens inside `hull run` / `nerdctl run` (`--pull missing`), so it goes to the held narration. At the default level that is `nil`, and nothing brackets it. With a stub that pulls six layers:

```
$ brig run -d ubuntu
brig: BRIG_VERIFY=off, so the guest image is not checked before it boots
brig-ubuntu
                    # 6.2s elapsed, nothing else printed
```

On a real first pull of an agent image that is minutes rather than seconds.

`RunSpec.Progress`'s own comment names "the image pull" as what it holds, and `runtimeNotice` says a download that takes a minute has to say so "or the terminal looks hung". But `noticef` only ever brackets the boot-asset download, so the longest wait on a first run is the one that goes unannounced.

The description flags that a successful run can now print nothing, though only for a warm sandbox. The cold case is the one I would worry about: a new user's first `brig run` looks like a hang, and some of them will Ctrl-C it.

Cheapest fix consistent with what you have already built is to put the "starting sandbox" line at the notice level rather than behind `--verbose`:

```go
noticef(spec.Notice, "starting sandbox %s...", spec.Name)
```

or bracket the run the way `pullAssets` brackets its own download. Either way it stays out of `-q`, which is the level that asked for none of it.

## Minor, on the numbers

284 ok + 16 FAIL is 300 assertions, so clearing the 16 should give 300 rather than the 288 in the description. Your delta is right though -- you measure +31 and so do I (269 -> 300 with no tty).

Also worth stating explicitly: this branch inherits `main`'s pty hang, so #128 wants to merge first. I killed this one at 240.1s at the same case. With #128 merged in, it is 300 ok / 0 FAIL in 19.3s under a real pty.

Happy to approve as soon as the cold-run line is in. Everything else here I would take as is.

ananos (approved):
Approving. The mechanism is right and I checked the parts that carry weight rather than taking the description on trust.

Verification is genuinely unsuppressible now. `alertf` has no level check, every refusal in `verify.go` goes through it, and `sayVerified` is the only `warnf` left. I planted a violation to see whether the source scan actually bites, turning the `verifyTag` refusal at `verify.go:102` into a `warnf`:

```
--- FAIL: TestVerifySitesDoNotReachForTheWarningLevel (0.00s)
    alert_test.go:164: verify.go:102: verifyTag uses warnf, which -q silences.
```

The `fn == ""` guard at the end is the right instinct, since a scan that quietly matches nothing is the usual way that kind of test rots.

The failed boot reports at both levels: with a stub runtime exiting non-zero, default and `-q` each quote the tail back indented. `go test ./... -race` is clean, and `cmd.Stdout, cmd.Stderr = said, said` is the same interface value, so `os/exec` reuses one pipe and only one goroutine ever writes into the buffer. `make all` is green here apart from `internal/secret`, which fails identically on `main` in a background job with no interactive keychain.

On flag positions, `-v` is refused naming itself, `--verbose` on the run line is handed to the agent with a notice pointing at the global position, `brig run ubuntu -d -q` prints the retirement notice, and `brig info` is unchanged at every level. `--quiet` and `--verbose` together are refused.

One thing stays open, and I am taking it as a follow-up rather than holding the PR for it. A cold run is silent at the default level: the image pull happens inside `hull run` / `nerdctl run`, so it lands in the held narration, and nothing brackets it. With a stub pulling six layers I measured 6.2s elapsed with nothing on stderr, which on a real first pull is minutes rather than seconds. `runtimeNotice` exists for exactly this and says so in its own comment, but `noticef` only brackets the boot-asset download. The cheapest fix consistent with what is already here is to put the "starting sandbox" line at the notice level:

```go
noticef(spec.Notice, "starting sandbox %s...", spec.Name)
```

Worth doing before the release that ships quiet-by-default, since a first `brig run` that prints nothing for a few minutes reads as a hang. Happy for it to be its own PR.

Numbers, for the record: with #128 in, this branch is 300 ok / 0 FAIL in 19.3s under a real pty, and 300 / 0 with no tty. Your delta of +31 matches mine.

Issue #15, opened 2026-08-25
Give each sandbox its own network, and add an offline mode


pmoust:
### Problem

Sandboxes share one network. The security doc states it plainly, under
"Things brig does not claim":

> It does not isolate one sandbox from another. Sandboxes share a network, not
> just a host: brig asks every runtime for its shared network, unconditionally
> and with no setting to say otherwise.

On `hvi` that is one gateway on `10.87.0.0/24`, with brig handing out the
addresses. Two agents holding different credentials sit on one broadcast
domain, and each can reach what the other is listening on. The doc calls this
"a real hole in the narrow-blast-radius argument", and the current advice is
to run them on separate hosts.

There is also no way to run an agent with no egress at all.

### Proposed solution

Take it in two steps, and ship the first one alone if the second is not ready.

**Step one: separation.** Give each sandbox its own network identity, so one
sandbox cannot reach another. On `hvi`, brig already owns the user-mode
gateway process and its subnet, which is where this fits. On `vz` and on
Linux, use the runtime's per-instance network.

**Step two: `--offline`.** A sandbox with no route out. The agent runs, the
workspace is there, nothing leaves.

Report the posture in the envelope from #10 and in `--json`, in both steps.
The posture is a fact about the sandbox, not a setting hidden in an
environment variable.

This ticket stops there. Egress allowlists, an approval flow and L7 rules are
a separate discussion, and neither step above depends on them.

### Alternatives considered

- Keep the shared network and document it. That is the state today. It leaves
  the "narrow blast radius" argument with a hole in it that the doc has to
  keep apologising for.
- Go straight to deny-by-default egress with allowlists. It is the better end
  state, and it is a much larger job. Separation and `--offline` are useful on
  their own, and they do not have to be redone afterwards.

### Effect on the sandbox boundary

It narrows the boundary in the one place the security doc says is open.

pmoust:
Scope update. This ticket now covers three controls, enforced at the runtime level rather than in the agent:

1. No inter-sandbox comms by default. Each sandbox gets its own network; on `hvi` that is one gateway per sandbox on its own `10.87.N.0/24`, carved from `10.87.0.0/16`. The default becomes `isolated`, and `network: shared` (or `BRIG_NETWORK=shared`) is the way back.
2. A complete no-egress mode: `--offline`, or `network: offline` in a profile. Maps to the runtime's `--net none` / `--network none`, which both runtimes already accept.
[...]

PR #130, opened 2026-09-04
feat(network): enforce an attached policy, on a gateway per sandbox

ananos:
## Summary

`docs/policies.md` ships a policy document format, five verbs and a record of what is bound to what, under a heading that admits the whole thing does nothing: *"Nothing today makes an agent's outbound traffic actually respect a policy you write."* `brig policy attach` and `brig policy check` print the same admission to the terminal. This is the enforcement point that note was waiting for.

A boot now resolves every policy bound to the run (inline `policy:`, profile attachment, session attachment), loads them, and puts the rules on the network gateway serving that sandbox.

Which requires the sandbox to have one of its own. An egress policy belongs to a gateway: it covers every member of that gateway's network, and no rule can name one member. hull's own gateway documentation states the consequence — *"A sandbox that needs an egress policy of its own needs a gateway of its own, so run one gateway per sandbox."* brig ran one gateway for the host, so it had nowhere to put a policy.

So this also implements **`--network isolated` on hull, which is accepted today and does nothing**. brig passed the posture straight through as `--net isolated`; hull's `--net` takes `none` or `shared` and reads anything that is not `none` as networked, so the flag was accepted, the value ignored, and the sandbox joined the shared gateway anyway. That is a silent gap between what the envelope reports and what the sandbox got.

**The default does not change.** A sandbox nobody attached a policy to has unrestricted egress, exactly as before. Measured on a real default sandbox against hull from `main`: it lands on the shared `198.18.0.0/24` (not a network of its own), its gateway is started with **zero** `--egress` flags, and from inside the guest `example.com` returned 200, `api.anthropic.com` 404 and `1.1.1.1` 301. No shipped profile binds a policy, and a test over the whole set holds that. Filtering is opt-in, per profile or per session.

## Related issues

Refs #15 (step one, on macOS: each sandbox gets a network of its own). Does not close it — `--offline` already exists as a posture, and #15 also asks for the posture in `--json`, which this does not touch.

## Changes

- **`--network isolated` on hull** gives the sandbox a gateway process of its own on a `/30` of its own. hull is told `shared`, the only thing it can be told; isolation is which gateway the guest is pointed at, not a word on the run line.
- **A bound policy is enforced** at that gateway: `--egress-default`, `--egress-allow`, `--egress-deny` from the merged rule set. Read once, at boot — editing a policy changes what the next boot enforces, never what a running sandbox is under, and no environment variable overrides it.
- **A policy makes a run isolated** whether or not it asked to be, and the envelope's `NETWORK` row says so rather than reporting a posture brig quietly changed. A new `POLICY` row names what applies, only when something does.
- **Isolated networks come from `198.18.1.0/24`**, adjacent to the shared network's `/24` inside the same RFC 2544 range #93 moved to. It has to be a different `/24`: the shared gateway serves the whole of that one, so a `/30` taken from there would be a network inside a network with two gateways claiming the same addresses.
- **Three refusals, never a silent drop**, because a sandbox that reports a policy and enforces nothing is worse than one with no policy — someone will rely on it:
  - a policy on `vz`, `qemu` or a container runtime, none of which brig filters (the refusal names the backend that does);
  - a policy against a runtime whose `network-gateway` has no `--egress-default` — every hull up to and including `0.1.0-rc21`. Asked of the binary, not derived from a version string;
  - a bound policy name that does not load. `check` reported this before a boot; now the boot refuses it too.
- **Several policies are unioned**, and that needed deciding rather than falling out. A rule in any of them applies, and the default is the strictest any names — one `deny` makes the run deny-by-default. The union costs what it costs: a host allowed by the second policy is reachable though the first alone would have denied it. Attaching is granting, and the docs now say so.
- **Lifecycle.** `stop` and `rm` take the gateway with them, gated on the verb having succeeded rather than a fresh `hull ps` (which would put another process on the path of every stop to answer what the exit status already answers). A sandbox that dies without either is reaped by `PruneNetworks` — the same `NetworkPruner` the nerdctl adapter implements with `network rm`, over a different primitive: here a sandbox's network is a process, so tidying one is stopping it.
- **`EnforcementNote`** replaces `NotEnforcedNote`: the commands now say where a binding is acted on rather than that it never is.

## Checklist

- [x] `make all` passes (vet, test, build) — except `internal/secret`, whose keychain tests fail identically on unmodified `origin/main` in this environment (`security: … User interaction is not allowed`; a background shell cannot unlock the login keychain). Verified against a clean worktree at `9c699a1`.
- [x] `script/smoke.sh` passes, including a new negative case: a policy attached, a `vz` run refused, and nothing reaching the runtime.
- [x] I have run `go test ./... -race`
- [x] I have exercised the change against a real runtime (`brig run <agent>`) — against a hull built from `main` at `a7a5d1e`, the first that takes the egress flags. **A real guest was booted behind a policy gateway on the `/30` this code assigns, and denied by the rules**: an allowed name returned 404 from Anthropic's edge, a denied name failed to resolve, and an address dialled directly failed to connect. Recorded in [`docs/manual-tests/egress-policy.md`](docs/manual-tests/egress-policy.md). hull's own `internal/netgw` egress e2e suite ran green against the same build.
- [x] I have updated the affected docs (`README.md`, `docs/security.md`, `docs/policies.md`)

This moves the second half of the boundary, so it is worth stating plainly.

`docs/security.md` said, under "Things brig does not claim", that brig does not sandbox the agent from the network and that there is no per-sandbox policy. On `hvi` both are now false, and the page says so: a policy is enforced at the gateway, and a run that cannot enforce one is refused. On `vz`, `qemu` and Linux both remain true, and the page still says that too.

It also sharpens the sandbox-to-sandbox claim. The doc's measured position was that separation on macOS is "a property that holds, not a guarantee brig makes" — brig asked for a shared network and happened to get guests that could not address each other, with no test that would notice if a runtime change removed it. `--network isolated` on hvi is now the guarantee rather than the accident.

**The negative test is the one that matters**, and it is in `script/smoke.sh` driving the real binary: with a policy attached and the run pinned to a backend that cannot enforce it, the boot is refused *and* nothing reaches the runtime. It passes with the feature working and fails with it deleted; "the sandbox booted" would not.

Nothing here touches credential resolution, forwarding or the workspace.

## What the real-runtime pass found

Running it against a hull that can enforce turned up three defects, fixed in the second commit. Worth reading as evidence that the first commit alone was not finished:

- **A policy attached after the sandbox booted was reported and not enforced.** `EnsureRunning` finds the sandbox up and returns without a boot, so the live sandbox stays behind whatever gateway it booted with — while the envelope prints a `POLICY` row for rules nothing is applying. That is the exact failure the refusals elsewhere exist to prevent, sitting on the one path that skips the boot. A network is a third thing fixed at boot, so it now joins the switch that already recreates a sandbox whose share went stale. Asked of the runtime through an optional interface, the way `NetworkPruner` already is.
- **An offline sandbox carrying a policy was refused.** The check asked whether the rules could be enforced without asking whether there was anything to enforce them on. A sandbox with no network satisfies every rule set.
- **Every gateway wrote to one `gateway.log`.** Harmless with one gateway per host, not with one per sandbox — and it is what the "did not come up" error points at, so a shared name sent a reader to another sandbox's output.

`PruneNetworks` had **no test at all**, which is the one thing standing between a sandbox that died without a `brig stop` and a host accumulating 28.7 MB a time. It now has four, driving real processes rather than a stub, because what it decides on is a pid it signals and an argv it reads back to confirm that pid is still the gateway it recorded.

## Notes for review

Two things I did not do, deliberately:

- **`CheckCoverage` still refuses `kind: shell` and `kind: gui`** on the grounds that they have "no agent to hook an egress rule into". That reasoning belongs to a tool-call enforcer; a gateway does not care what runs inside the guest, so the restriction is now unnecessary. I left it alone rather than widen someone else's design inside this change. Worth a follow-up.
- **Two stale shared gateways are running on this machine** on the pre-#93 socket name (`~/.brig/gateway.sock`), left behind when the subnet moved. Nothing reaps the shared gateway — `PruneNetworks` here deliberately does not touch it, since it serves sandboxes a single run knows nothing about. Pre-existing and out of scope, but it is a leak with no owner.
- **`198.18.0.0/15` is not a private range.** It is IANA *Benchmarking* space (RFC 2544), reserved and not globally routable, which is a different property from RFC 1918 private. #93 never claimed otherwise and its collision-avoidance reasoning is sound; I extended the same range by one `/24` rather than reopen the choice inside an unrelated feature. If it is worth revisiting, it wants its own issue.

## LLM usage

Authored with assistance from Claude Code; I have reviewed every line and am accountable for the change.


internal/wrap/run.go
```diff
@@ -232,6 +232,15 @@ func (c *Config) EnsureRunning(set creds.Set) error {
 		// Recreate rather than fail: all persistent state lives in the
 		// workspace on the host, so restarting costs nothing but the boot.
 		switch stale := c.projectShareStale(); {
+		case c.networkStale():
```
pmoust, line 242: The policy refusal is skipped when the sandbox is already running. supports() is only reached through Run, and NetworkStale answers false on every backend except hvi.

With BRIG_HYPERVISOR=vz: boot a sandbox, then run brig policy attach no-net claude-code, then run brig run claude. Running is true, so c.networkStale() calls hull.NetworkStale, which returns false at its first line because hypervisorOrDefault("vz") is not hvi. The switch falls to the default branch and returns deliverSecretFiles(). Run never happens, so supports() never refuses. The user gets a POLICY row and an isolated NETWORK row on a vmnet sandbox that filters nothing.

The smoke test passes only because it runs brig rm --all first. The support check needs to run on the already-running path as well.

internal/runtime/gateway.go
```diff
@@ -104,32 +215,73 @@ func ensureGateway(bin string) (string, error) {
...
+// reusing it would enforce the policy as it stood when the sandbox last
+// booted, or route guests on a network the process on the other end does not
+// serve. Restarting is safe here because this runs before the guest attaches.
+func ensureIsolatedGateway(bin, name string, index int, policy Egress) (string, error) {
+	sock, err := isolatedSocket(name)
+	if err != nil {
+		return "", err
+	}
+	if policy.Filtered() {
+		if err := gatewayEnforces(bin); err != nil {
```
pmoust, line 242: gatewayEnforces starts hull network-gateway --help before the reuse check, so every policied boot pays for the probe even when no gateway will be started.

The probe runs here. Only after it does ensureIsolatedGateway check gatewayReachable and recordedSpec, and return the existing socket when they match. On the common path, the same sandbox restarted under the same policy, the extra process is pure waste on the boot path.

Move the probe below the reuse check, into the branch that calls startGateway. The protection is the same and it costs nothing on a reuse.

internal/runtime/gateway.go
```diff
@@ -104,32 +215,73 @@ func ensureGateway(bin string) (string, error) {
...
+			return "", err
+		}
+	}
+	want := gatewaySpec(index, policy)
+	if gatewayReachable(sock) {
+		if recordedSpec(sock) == want {
+			return sock, nil
+		}
+		shutDownGateway(name)
```
pmoust, line 256: When a gateway must be replaced, a surviving old process satisfies the readiness check. The boot then proceeds under the old rules while the spec file records the new ones.

shutDownGateway is best effort. If the .pid file is missing, or ownsGateway is false, or hull does not die inside the 3s gatewayStopTimeout, the old gateway keeps listening. startGateway then ignores cmd.Start() and the exit of the child on purpose, so the new hull fails with "socket is in use" unseen. Success is judged only on gatewayReachable(sock), and the old process answers it. That socket is returned as ready.

writeGatewayRecord has already written the new spec beside it, so recordedSpec now claims rules that the running process never received. On the next run NetworkStale compares equal and never restarts. The sandbox stays under the previous policy for good, while brig reports the current one.

pmoust (commented):
Nothing in the nerdctl adapter reads RunSpec.Egress, and the only refusal is in the supports() function of hull. On Linux, or with BRIG_RUNTIME=nerdctl, attach a policy with egress.default: deny and run brig run claude. wrap/config.go raises the posture to isolated. The envelope prints a POLICY row that names the policy and a NETWORK row that says isolated. The spec carries the rules. nerdctl.Run ignores spec.Egress. 

The sandbox boots with unrestricted egress. docs/policies.md says that everywhere else the boot is refused rather than left unenforced, and EnforcementNote tells every attach and check user the same thing. 

This is the case the PR exists to prevent. Give the nerdctl adapter the same refusal.

internal/runtime/gateway.go
```diff
@@ -104,32 +215,73 @@ func ensureGateway(bin string) (string, error) {
...
+// because its network is in its name and it carries no rules.
+func startGateway(bin, sock, subnet, gatewayIP string, policy Egress, spec string) (string, error) {
 	if err := os.MkdirAll(filepath.Dir(sock), 0o700); err != nil {
 		return "", fmt.Errorf("could not create the gateway directory: %w", err)
 	}
 
 	// A stale socket file left by a crashed gateway is hull's problem, not
 	// ours: it claims the path and removes a file nothing is listening on.
-	logPath := filepath.Join(filepath.Dir(sock), "gateway.log")
+	logPath := gatewayLogPath(sock)
```
pmoust, line 279: Nothing removes the per-socket log file of a gateway. ~/.brig grows one sandbox-NAME.log for every sandbox that was ever isolated and keeps it for ever.

startGateway opens gatewayLogPath(sock) with O_APPEND. stopGatewayAt removes only the .pid and .spec files. Stop, Remove and PruneNetworks remove nothing else. A sandbox that is created, run and removed leaves its log behind with no owner. A long-lived sandbox appends to the same file on every boot, with no rotation and no cap.

internal/runtime/gateway.go
```diff
@@ -154,6 +310,169 @@ func ensureGateway(bin string) (string, error) {
...
+// not have to repeat it.
+func stopGatewayAt(sock string) {
+	if pid, ok := gatewayPID(sock); ok && ownsGateway(pid, sock) {
+		_ = syscall.Kill(pid, syscall.SIGTERM)
+		deadline := time.Now().Add(gatewayStopTimeout)
+		for time.Now().Before(deadline) && gatewayReachable(sock) {
+			time.Sleep(gatewayPollInterval)
+		}
+	}
+	_ = os.Remove(gatewayPIDPath(sock))
```
pmoust, line 501: stopGatewayAt deletes the pid and spec records even when the process was not stopped. That puts a surviving gateway beyond the reach of PruneNetworks.

SIGTERM is sent once and no SIGKILL follows it. After the 3s poll the function removes gatewayPIDPath and gatewaySpecPath unconditionally. It removes them even when gatewayPID returned !ok, so no signal was sent at all. PruneNetworks sweeps only sandbox-*.pid entries, so with the record gone the orphaned gateway is invisible to brig reset, to a later brig stop, and to shutDownGateway. It holds a socket and 28.7 MB for the life of the login session.

Remove the record only after the process is confirmed gone.

internal/policy/merge.go
```diff
@@ -0,0 +1,110 @@
...
+// reports the same condition before a boot rather than during one.
+func Resolve(p profile.Profile, session, dir string) (Egress, []string, error) {
+	names, err := EffectivePolicies(p, session, dir)
+	if err != nil {
+		return Egress{}, nil, err
+	}
+	if len(names) == 0 {
+		return Egress{}, nil, nil
+	}
+	entries, err := LoadAll(dir)
```
pmoust: Resolve treats any LoadAll error as fatal. That breaks the documented contract of LoadAll, and it lets one unrelated bad file in the policy directory refuse every policy-bound boot.

LoadAll says that a file which fails to parse does not stop the others from loading, and that the caller must read the returned map even when err is not nil. It returns a non-nil error for an unparseable file and also for two files that declare the same name. Resolve returns that error, Load raises it to strictErr, and Load returns it.

So an editor backup no-net.yaml.bak renamed to .yaml, a half-written document, or two files with the same name field, makes brig run claude fail outright for a profile whose own policy loaded without trouble. brig policy ls and brig policy check still work, which makes the cause hard to find.

internal/runtime/hull.go
```diff
@@ -344,9 +363,49 @@ func supports(spec RunSpec, hv string) error {
...
+	return "shared"
+}
+
+// isolatedNet reports whether this sandbox gets a network of its own.
+//
+// Asking for the isolated posture is one way in. Carrying a policy is the
+// other, and it is not a choice: an egress policy belongs to a gateway and
+// covers every member of its network, so a sandbox answering to rules of its
+// own cannot be sharing a gateway with sandboxes that do not.
+func isolatedNet(net string, policy Egress) bool {
```
pmoust, line 421: isolatedNet is read only under hv == "hvi", so --network isolated is still a no-op on vz and qemu, while the envelope keeps reporting a network of the sandbox's own.

Run brig run claude --network isolated on the default vz backend. hullNet now maps isolated to shared, hull.Run never enters the gateway branch, and the guest lands on the same vmnet network as every other sandbox. Network.Line() still prints "isolated (a network of this sandbox's own)". Nothing warns and nothing refuses.

This is the same shape of silent gap that the PR closes on hvi, left open on the backend that most runs use. docs/security.md admits it in prose, but the terminal does not.

internal/runtime/hull.go
```diff
@@ -534,17 +593,127 @@ func (h *hull) replaceCmd(spec ExecSpec) (argv, env []string) {
...
-// Remove also gives the sandbox's gateway address back. A stopped sandbox
-// keeps its address so that starting it again is the same machine on the same
+// Remove also gives the sandbox's address and network back. A stopped sandbox
+// keeps both so that starting it again is the same machine on the same
 // network; a removed one has no such claim.
 func (h *hull) Remove(name string) error {
 	err := h.quiet("rm", name, false)
+	releaseGateway(name, err)
 	releaseGatewayIP(name)
+	releaseSandboxNet(name)
```
pmoust: Remove frees the /30 of the sandbox and its shared address even when hull rm failed. releaseGateway just above it is gated on success.

releaseGateway(name, err) returns early on err, and its comment explains why: taking the network from a sandbox that can still be running turns one failure into two. releaseGatewayIP and releaseSandboxNet on the next two lines have no such gate. When hull rm fails and the sandbox survives, its network index goes back into the pool, and the next sandbox to boot can claim the same /30.

The same call is also on the path of every boot, at wrap/run.go:271, "Clear a stale stopped instance holding the name". So the assignment of an isolated sandbox is deleted and derived again on each start, which contradicts the promise in this file that a sandbox comes back on the network it had.

internal/runtime/hull.go
```diff
@@ -534,17 +593,127 @@ func (h *hull) replaceCmd(spec ExecSpec) (argv, env []string) {
...
+// direction is not trusted in the other either.
+func (h *hull) NetworkStale(name, hypervisor, net string, e Egress) bool {
+	if hypervisorOrDefault(hypervisor) != "hvi" || net == "none" {
+		// No gateway either way, so nothing here can differ. vz takes its
+		// network from vmnet, which brig neither chose nor can inspect.
+		return false
+	}
+	want := ""
+	if isolatedNet(net, e) {
+		index, err := sandboxNet(name)
```
pmoust: NetworkStale is a predicate that writes to disk. It calls sandboxNet, which allocates and persists a network index as a side effect.

NetworkStale runs on every brig run, brig sh and brig exec against a running sandbox. When isolatedNet is true it calls sandboxNet(name), which writes sandbox-nets.json. A run that is then refused, for a bad image, a failed verification, or a policy the backend cannot enforce, has still consumed one of only 64 networks. Only brig rm gives it back. On a host that cycles through sandbox names the space fills, and later boots die with "the isolated address space 198.18.1.0/24 is full".

The comparison needs a read-only lookup. Allocation belongs on the boot path.

internal/wrap/egress.go
```diff
@@ -0,0 +1,45 @@
+package wrap
+
+import (
+	"github.com/brig-sh/brig/internal/policy"
+	"github.com/brig-sh/brig/internal/runtime"
+)
+
+// runtimeEgress is the one place a policy document becomes something a runtime
```
pmoust, line 8: The doc comment for runtimeEgress has no blank line after it, so godoc attaches the whole block to networkStale and runtimeEgress ends up with no documentation at all.

Lines 8 to 15 describe runtimeEgress, "This function is the seam...". Line 16 then starts "// networkStale reports whether..." with no blank line between them and no declaration in between. Go reads one comment block and binds it to func (c *Config) networkStale. runtimeEgress at line 33 gets nothing. A reader of go doc wrap sees networkStale described as a converter from policy to runtime.

Put a blank line before the networkStale comment.

internal/runtime/sandboxnet.go
```diff
@@ -0,0 +1,200 @@
...
+	sandboxNetBits   = 30
+	sandboxBlockSize = 4
+
+	// How many sandboxes the space holds. Well past what a host can run one
+	// microVM each for, and past what their gateways alone would fit in
+	// memory: 64 of them is nearly 2 GB before a single guest boots.
+	sandboxNets = 256 / sandboxBlockSize
+)
+
+// sandboxNetStore is the file holding the assignments.
```
pmoust: This file repeats gatewayip.go almost line for line instead of sharing the allocator that is already there.

sandboxNetStore, readSandboxNets, writeSandboxNets, sandboxNet, releaseSandboxNet and lowestFreeNet are the same JSON map, the same write by temp file and rename, the same fallback for a corrupt file, the same scan for the lowest free slot and the same sorted "space is full" message as gatewayIPStore, readGatewayIPs, writeGatewayIPs, gatewayCIDR, releaseGatewayIP and lowestFreeHost in gatewayip.go. About 90 duplicated lines, different only in the store filename, the range bounds and the error text.

One allocator that takes a store path and a range would carry both. A fix to one, the missing lock for example, would then reach the other.

internal/runtime/sandboxnet.go
```diff
@@ -0,0 +1,200 @@
...
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+	return os.Rename(tmp.Name(), path)
+}
+
+// sandboxNet returns the index of the network this sandbox sits on, assigning
+// one if it has none. Everything addressable about the sandbox is derived from
+// it: the subnet, the gateway on it, and the guest's own address.
+func sandboxNet(name string) (int, error) {
```
pmoust, line 69: sandboxNet reads, changes and writes the shared assignment file with no lock. Two boots at the same time can hand the same /30 to two sandboxes and lose one of the two entries.

Two brig run invocations for different sandboxes start together. Both call readSandboxNets and see the same map. Both compute lowestFreeNet and get the same index. Both write. The rename makes the last writer win, so one sandbox has no recorded index and is assigned again on its next boot, and meanwhile both are addressed on the same 198.18.1.N/30.

The same race is already in gatewayip.go. That is one more reason to share a single allocator and fix it once.

cmd/brig/policy_test.go
```diff
@@ -1724,10 +1724,10 @@ func TestAttachSaysThePolicyIsNotEnforcedYet(t *testing.T) {
 			t.Fatal(err)
 		}
 	})
-	if !strings.Contains(note, policy.NotEnforcedNote) {
+	if !strings.Contains(note, policy.EnforcementNote) {
```
pmoust, line 1727: Two tests now assert the opposite of what their names and their failure messages say.

TestAttachSaysThePolicyIsNotEnforcedYet at line 1709 and TestCheckSaysThePolicyIsNotEnforcedYet at line 1738 were switched to EnforcementNote, which says that a policy is enforced on hvi. Their names and their t.Errorf text, "attach did not say the binding is not enforced", now describe behaviour the code no longer has. A reader who scans the test list is told the feature does not exist.

Rename both tests and reword both messages.

internal/wrap/run.go
pmoust: Fixed. EnsureRunning calls checkBackend before it asks whether the sandbox is running, so the refusal reaches the path that joins a running sandbox. hull.CanRun is supports for a caller that has booted nothing, and canrun_test.go covers it.

internal/runtime/gateway.go
pmoust: Fixed. ensureIsolatedGateway does the reuse check first, and probes only when policy.Filtered() is true. A restart under the same policy now starts no extra process.

internal/runtime/gateway.go
pmoust: Fixed. The socket is checked again after shutDownGateway, and the boot is refused, naming the socket to clear, rather than started on top of a process that would not go. startGateway also records the spec only for a pid that owns the socket, so a loser of the race cannot write a record.

internal/runtime/gateway.go
pmoust: Fixed. clearGatewayRecord removes the log beside the pid and spec files. TestStopClearsTheRecordsAndTheLog covers it.

internal/runtime/gateway.go
pmoust: Fixed. stopGatewayAt escalates from SIGTERM to SIGKILL, clears the record only once the process is confirmed gone, and returns false when it survives, so the record stays as the only handle on it. TestStopEscalatesWhenSigtermIsIgnored covers it.

internal/policy/merge.go
pmoust: Fixed. The LoadAll error is held as loadErr and fails the resolve only when a bound name is missing, where it travels with the refusal as the likely reason. One unrelated bad document in the directory no longer refuses every policy-bound boot.

internal/runtime/hull.go
pmoust: Fixed. supports refuses --network isolated when the backend is not hvi, with the reason and the alternative in the message. hullNet maps the posture to what hull actually takes, so the value is no longer passed through as networked. Both Run and the join path reach the refusal through CanRun.

internal/runtime/hull.go
pmoust: Fixed. releaseGatewayIP and releaseSandboxNet are inside if err == nil now, on the same terms as releaseGateway, with the reason in the comment.

internal/runtime/hull.go
pmoust: Fixed. NetworkStale calls lookupSandboxNet, which reads and never assigns. A sandbox with no network recorded is reported stale, which is the right answer for a sandbox that cannot be running on a network it was never given.

internal/wrap/egress.go
pmoust: Fixed. networkStale moved to internal/wrap/network.go, so the comment block ends at runtimeEgress and both functions are documented.

internal/runtime/sandboxnet.go
pmoust: Fixed. Both allocators are netAlloc in netalloc.go now, taking the store path, the bounds and the wording of the full message. gatewayip.go and sandboxnet.go keep only what differs between them.

internal/runtime/sandboxnet.go
pmoust: Fixed. netAlloc.slot takes an exclusive flock on a lock file beside the store and holds it across the read, the pick and the write. The lock file is separate from the map so the atomic rename cannot swap the file out from under a held descriptor.

cmd/brig/policy_test.go
pmoust: Fixed. Both tests are now TestAttachSaysWhereThePolicyIsEnforced and TestCheckSaysWhereThePolicyIsEnforced, and the failure messages say where the binding is enforced.

pmoust:
Re-reviewed at ab0d63c. Every finding from the earlier review is addressed, and each inline thread has a reply that says what closed it. go vet is clean and the tests pass in internal/runtime, internal/wrap, internal/policy and cmd/brig.

The one finding with no line to anchor to is also fixed. internal/runtime/nerdctl.go now has CanRun, which refuses a filtered policy on any networked sandbox instead of dropping it, and Run calls it. CanRun is on the Runtime interface, hull implements it as supports, and internal/wrap/network.go asks for it through the RunChecker assertion, so a runtime with no opinion is simply not asked.

Two of the fixes are worth naming because they close more than the finding did. netalloc.go replaces the two copies of the allocator with one type, so the missing lock is fixed in both places at once and cannot go missing in one of them again. supports now refuses --network isolated off hvi, which was a silent no-op before this branch and is not one the branch introduced.

pmoust (approved):

PR #131, opened 2026-09-04
fix(cli): class a verbless bad ref as a usage error, not failure

### LINKED ISSUE #127: Classify a mistyped ref without a verb as a usage error, so it exits 2 like brig run
## Describe the bug

A mistyped session ref exits with two different codes depending on whether a
verb was typed. The message is the same either way, which is correct and was
fixed in #119. The classification was not: the verbless branch returns the
`ParseRef` error bare, where `split` wraps the same error in `usagef`.

`cmd/brig/main.go:303`:

```go
if refErr != nil {
    return refErr
}
```

A bare error falls through `exitCode` to 1. README documents 1 as "general
failure -- something ran and did not finish" and 2 as "usage error -- an unknown
flag, a stray argument, a value in the wrong place". Nothing ran here, so 2 is
the right class.

## To reproduce

1. `brig claude@BAD; echo $?`
2. `brig run claude@BAD; echo $?`

## Expected behavior

Both exit 2. The two spellings are the same mistake and already print the same
message.

## Actual behavior

```
TOKEN              brig X    brig run X
claude@BAD         1         2
claude@a@b         1         2
claude@            1         2
@refactor          1         2
nosuch@refactor    3         3   <- correct, this one falls through to the run line
```

The unknown-agent case is right because it falls through to the run line and
picks up `notFoundError`. Only the `refErr != nil` branch is misclassified.

## Suggested fix

```go
-	return refErr
+	return usagef("%s", refErr)
```

`TestARefShapedTokenIsDiagnosedAsARef` (`cmd/brig/lifecycle_test.go:174`) does
not catch this because it compares `.Error()` strings, which are already equal.
The error type differs, not the message. Adding the code to the same comparison
would pin it:

```go
if exitCode(verbless) != exitCode(verbed) {
    t.Errorf("brig %s exits %d, brig run %s exits %d",
        ref, exitCode(verbless), ref, exitCode(verbed))
}
```

## Environment

- Host OS: macOS 26.3.1
- Architecture: arm64
- Installed how: `make build`
- Commit: c06f9bc

Raised on #119 before it merged, in
https://github.com/brig-sh/brig/pull/119#issuecomment-5510215817, and filed here
because that PR is now closed. Table above captured from a binary built at
c06f9bc.

Low impact: the message is right in both spellings, so this only affects a
script branching on exit status rather than reading output. It is a documented
contract though, and the contract is the reason the codes exist.

pmoust:
`brig claude@BAD` exited 1. `brig run claude@BAD` exited 2. Same mistake, same message, two exit codes.

Nothing runs in either case, so 2 is the right class, and it is what `split` already gives the verbed form. The verbless branch returned the `ParseRef` error bare, which falls through `exitCode` to 1. It is now wrapped in `usagef`. The message is unchanged, so the two spellings still read the same word for word.

`nosuch@refactor` is untouched. It falls through to the run line and exits 3 both ways.

`TestARefShapedTokenIsDiagnosedAsARef` compared only the message strings, which were already equal, so it never caught this. It now compares the exit code as well, over every token in the table.

Fixes: #127

### DIFF
--- a/cmd/brig/lifecycle_test.go
+++ b/cmd/brig/lifecycle_test.go
@@ -174,6 +174,14 @@ func TestARefShapedTokenIsDiagnosedAsARef(t *testing.T) {
 		if verbless.Error() != verbed.Error() {
 			t.Errorf("brig %s said\n  %v\nbut brig run %s said\n  %v", ref, verbless, ref, verbed)
 		}
+		// The message being equal is not enough: a bare ParseRef failure falls
+		// through to exit 1 while the verbed form's usage error exits 2, so the
+		// two read alike and a script still tells them apart. Pin the class too,
+		// including nosuch@refactor, which already reaches exit 3 both ways.
+		if exitCode(verbless) != exitCode(verbed) {
+			t.Errorf("brig %s exits %d, brig run %s exits %d",
+				ref, exitCode(verbless), ref, exitCode(verbed))
+		}
 		if strings.Contains(verbless.Error(), "unknown command") {
 			t.Errorf("brig %s was reported as an unknown command: %v", ref, verbless)
 		}
--- a/cmd/brig/main.go
+++ b/cmd/brig/main.go
@@ -328,8 +328,14 @@ func run(args []string) error {
 		// vocabulary of commands would send the reader looking for a verb they
 		// did not type, past a diagnosis that already names what is wrong with
 		// the ref.
+		//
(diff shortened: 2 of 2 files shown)


ananos (approved):
Nice one. The unknown-agent case still falls through to the run line, so only a genuine ParseRef failure is reclassified -- which is the right half. Message untouched, only the exit code joins them.

PR #132, opened 2026-09-04
test(brig): cover flag overlap for every shipped profile

### LINKED ISSUE #124: Cover every shipped profile in the brig/agent flag overlap test, not just claude-code
## Problem

`TestBrigFlagsOverlapWithClaudeCodeOnlyWhereKnown` in `cmd/brig/hygiene_test.go`
is a good guard: it fails when a brig flag collides with one of claude-code'"'"'s,
and also when a documented overlap stops colliding, so the map cannot outlive
the collision it records.

It only knows claude-code'"'"'s vocabulary. brig ships eight profiles: claude-code,
claude-desktop, codex, cursor, gemini, grok, opencode and ubuntu. A brig flag
that collides with any of the other seven is added silently, and brig reads it
first, so the agent never sees it.

Concretely, #115 adds `--mem`. On a line like `brig run codex /path --mem 4096`
brig now takes it. That is correct for codex, which has no `--mem`, but nothing
in the tree checks that.

## Proposed solution

Extend the test to cover every shipped profile, so adding a flag has to declare
any overlap for all of them rather than for one.

## Alternatives considered

Leaving it as is and relying on review. That is what happens today, and it works
until someone adds a flag that an agent other than claude-code already uses.

## Environment

- Host OS: macOS 26.3.1
- Architecture: arm64
- Installed how: `make build`

## Additional context

Found while reviewing #115. Not a blocker: no overlap exists today beyond the
four already documented in `knownOverlaps`. Filed so the coverage gap is
tracked rather than rediscovered.

pmoust:
The flag-overlap guard knew one agent. brig ships eight profiles, so a brig flag colliding with any of the other seven was added in silence. brig reads the run line first, so the agent never sees the flag.

The test is now table-driven over all eight, keeping both directions per profile: an undocumented collision fails, and so does a stale entry.

Each flag list carries its provenance, because the lists are not equally strong. claude-code, codex, cursor and grok are `--help` captures from a machine with those agents installed. gemini and opencode are their published CLI references, which can lag or lead the binary. claude-desktop is a GUI and ubuntu is a shell, so neither execs an agent CLI; both get an empty list and a comment, so the profile is accounted for rather than forgotten.

The claude-code list was captured at 2.1.252 and is refreshed to 2.1.261. It was missing `--permission-prompt-tool`, `--permission-prompts` and `--system-prompt-snapshot`.

Collisions this surfaces, all newly recorded:

| brig flag | agent |
| --- | --- |
| `--image`, `-m` | codex |
| `--workspace`, `-w` | cursor |
| `-d`, `-m` | grok |
| `-d`, `-m`, `-w` | gemini |
| `-n`, `-m` | opencode |

No brig flag is changed to dodge a collision. Most of these spellings are already retiring, and each entry says so.

Fixes: #124

### DIFF
--- a/cmd/brig/hygiene_test.go
+++ b/cmd/brig/hygiene_test.go
@@ -155,81 +155,211 @@ func TestBrigFlagsDeclareAPosition(t *testing.T) {
 // brig reads the flag and the agent never sees it, and `--` is the only way
 // through.
 //
-// The overlaps below are what that costs today. They are the reason #47
-// retires -n, -t, -w and -m in v0.3, and the map is expected to shrink as it
-// does; a spelling that starts colliding fails here first.
+// The overlaps below are what that costs today, one map per shipped profile.
+// They are the reason #47 retires -n, -t, -w and -m in v0.3, and the maps are
+// expected to shrink as it does; a spelling that starts colliding, or one whose
+// collision has gone, fails here first so a map cannot outlive what it records.
 //
-// claudeCodeFlags is hand-maintained. No profile spec declares its agent's
+// The flag lists are hand-maintained. No profile spec declares its agent's
 // flags -- internal/profile carries the image, the binary and the credentials
-// and nothing about the CLI's grammar -- so there is nothing to derive this
-// from. It was captured from `claude --help` (Claude Code 2.1.252) and covers
-// that one agent: the other shipped profiles' CLIs are not on this machine and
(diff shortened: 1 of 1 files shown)


ananos (approved):
Good extension. Tiering the evidence (local capture vs published reference) and dating the captures is the right call -- it says how strong each list actually is. The reverse check per profile keeps a map from outliving the collision it documents, which is what makes this maintainable.

PR #133, opened 2026-09-04
fix(cli): warn about tail flags only on verbs that forward a tail

### LINKED ISSUE #123: Scope the agent-tail flag warning to the verbs that forward a tail
## Describe the bug

`agentTail` warns about brig'"'"'s own flags found in the agent tail, but it runs for
every verb, including the ones that forward no tail at all. On those verbs the
warning contradicts the error printed immediately after it.

## To reproduce

1. `brig stop claude extra -q`

## Expected behavior

Either no warning, since `brig stop` has no agent to forward `-q` to, or a
warning that matches what the verb actually does.

## Actual behavior

Two messages that disagree. The first says the token belongs to the agent, the
second says the verb takes no tail:

```
brig: -q is one of brig's own flags, but here it is the agent's: brig stopped
      reading the line at the argument before it. Put -q before the profile for
      brig to read it.
brig: unexpected argument "extra"; `brig stop` takes a profile and nothing more
```

The warning fires inside `split`, before `rejectTail` runs.

A related case on a verb that does forward a tail: `brig sh claude ls -q` says
"Put -q before the profile for brig to read it", but `--quiet` is documented as
"with run or create", so the advice does not apply to `sh` either.

## Environment

- Host OS: macOS 26.3.1
- Architecture: arm64
- Installed how: `make build`

Found while reviewing #115, which introduced `agentTail`. Not a blocker for
that PR: every line that works today still works, and the token is still
forwarded. Filed separately so the PR can land.

pmoust:
`agentTail` warned about brig's own flags in the agent tail on every verb, including the ones that forward no tail. `rejectTail` then refused the same line, so `brig stop claude extra -q` printed two messages that disagreed: one said `-q` belongs to the agent, the next said `stop` has no agent to give it to.

The set of verbs that forward a tail now lives in one predicate, `forwardsTail`, and both `rejectTail` and `agentTail` read it. A verb that forwards nothing says nothing in `agentTail` and leaves the one correct message to `rejectTail`.

The advice on the verbs that do forward a tail was also wrong for `sh`. `brig sh claude ls -q` said to put `-q` before the profile, but `sh` reads `-q` in no run-line position. The advice now names the position the typed verb really reads the flag in, and when there is none, says so instead of sending the user to a second refusal.

Fixes: #123


cmd/brig/main.go
```diff
@@ -950,43 +970,105 @@ func warnPositionalMeaning(word string) {
...
+// run reads every run-line flag it has. sh, shell and exec continue a session
+// rather than shaping a fresh run, so the flags a run alone acts on are read in
+// no run-line position on them: -q suppresses the execution envelope, which is
+// documented for run and create; --no-project and --detach are run's own.
+func honorsRunLine(verb, long string) bool {
+	if verb == "run" {
+		return true
+	}
+	switch long {
+	case "quiet", "no-project", "detach":
```
ananos: This is a second hand-maintained list of what a verb reads, sitting beside `brigFlags`, which already carries a position per flag. It is the same shape `forwardsTail` just removed above, so it would be a shame to reintroduce it here.

A new run-only flag, or `--no-project` gaining a meaning on `sh`, changes the right answer and nothing fails. Could the table carry it instead -- a `runOnly` field on the row, say? The default branch has no test either, so only `-q` is covered today.

Not blocking, just noting it while we are in here.

ananos (commented):
Good fix -- `forwardsTail` as the one predicate that both `rejectTail` and `agentTail` read is exactly right, and it kills the contradiction cleanly.

One small thought inline, no need to block on it.

cmd/brig/main.go
pmoust: Agreed, and done in 4c9a224. The row carries it now: `runOnly` on `--no-project`, `--detach` and the run-line `-q`, and `honorsRunLine` reads the table. A test walks every run-line row and checks that sh's answer matches the row, so a flag added without the decision fails rather than drifts. The verb-list half of the same pattern is filed as #145.

pmoust:
Folded to one commit, per the commit-shape note on #140. Same tree, one message carrying the whole change.

ananos (approved):
The table carries it now and a test walks every run-line row, so a flag added without the decision fails rather than drifts. That is better than what I asked for. Verified `runOnly` on the three rows and `honorsRunLine` reading the table.

PR #134, opened 2026-09-04
fix(session): scope the reserved-label refusal to the asking agent

### LINKED ISSUE #125: Scope the reserved-label refusal to the agent whose workspace could collide
## Describe the bug

`profile.Reserved` matches a label against the trailing word of any reserved
profile name, without checking whether the agent asking could actually collide
with it. So a label is refused for an agent whose workspace it would never
touch.

## To reproduce

1. `brig run codex@desktop`

## Expected behavior

Accepted. The workspace would be `~/brig/codex-desktop`, and no profile named
`codex-desktop` exists, so nothing collides.

## Actual behavior

```
brig: session label "desktop" is the workspace the claude-desktop profile
      already uses. Pick another label
```

The refusal is correct for `claude@desktop`, which would land on
`~/brig/claude-desktop`. It is wrong for every other agent, because the check
reads the label alone rather than `<agent>-<label>`.

## Environment

- Host OS: macOS 26.3.1
- Architecture: arm64
- Profile: codex
- Installed how: `make build`

Pre-existing in `session.Resolve`; #115 carries it into the new `agent@label`
form, which is how I ran into it. Not a blocker for that PR, since it
over-refuses rather than under-refuses. I checked all eight shipped profiles for
the opposite case, a `<agentA>-<label>` that equals another profile'"'"'s name, and
there is none today.

pmoust:
`profile.Reserved` matched a label against the trailing word of any reserved profile name without asking which agent wanted it. So `codex@desktop` was refused because `claude-desktop` exists, even though its workspace would be `~/brig/codex-desktop`, a directory no profile owns.

A session's workspace is `~/brig/<profile>-<slug>`, where the profile is the name as it resolved, not the word typed. `claude` is an alias for `claude-code`, so `claude@desktop` lands on `~/brig/claude-code-desktop` and collides with nothing either. The issue's own text has that pair wrong, and the correction is on the issue.

Two commits. The first passes the agent to `Reserved` and compares `<agent>-<slug>` against the reserved names. The second resolves the agent through `profile.Lookup` first, so both call sites, `ParseRef` for `agent@label` and `Resolve` for `--name`, compare the same canonical name and give the same answer for the same workspace. They disagreed between the two commits, which the new test now pins.

With an agent given, only the pair rule applies: a slug is never in front on its own, so a slug equal to a reserved profile's whole name is not a session workspace. Without an agent, which is what import asks, the whole-name and trailing-word rules stand as before.

The guard still fires where it should. A user profile actually named `claude`, which `Lookup` prefers over the alias, makes `claude@desktop` land on `~/brig/claude-desktop` and it is refused, naming `claude-desktop` as the owner.

Checked on a build of this branch: `claude@desktop`, `codex@desktop`, `codex@claude-desktop` and `claude --name desktop` are all accepted; `nosuch@desktop` exits 3. Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`.

Fixes: #125


internal/profile/registry.go
```diff
@@ -287,23 +287,48 @@ func Names() []string {
...
+// no agent and the workspace is the name on its own. With none given, a slug
+// that is a reserved profile's whole name is refused, and so is the trailing
+// word a slug of the reserved name reads as -- which is how a profile named
+// "desktop" is turned away for reading as claude-desktop.
 //
 // It reads the merged set rather than the built-ins alone, so a profile of
 // your own can declare itself reserved. That is the honest reading of a field
 // a file can now set.
-func Reserved(slug string) (string, bool) {
+func Reserved(slug string, agent ...string) (string, bool) {
```
ananos: The variadic reads as an optional argument, so a caller that forgets it silently gets the old unscoped behaviour. That fails safe -- it refuses more, never less -- but it fails quietly, and the compiler will not help us.

Two separate functions would say it plainly: `Reserved(slug)` for the profile-name case and `ReservedFor(slug, agent)` for the scoped one. Up to you, it is a small thing.

Worth a line in the doc comment either way: the scoped form wants the canonical profile name. `Reserved("desktop", "claude")` refuses and `Reserved("desktop", "claude-code")` does not, which is right, but only obvious once you know the call site resolves the alias first.

ananos (commented):
Good catch on the scoping -- `codex@desktop` had no business being refused, and the ref-parse cases read well.

One note on the signature, inline.

internal/profile/registry.go
pmoust: Done in 82e5304: `Reserved(slug)` for the profile-name question import asks, `ReservedFor(slug, agent)` for a session. The doc comment on `ReservedFor` now says `agent` is the profile as it resolved, with the `claude` versus `claude-code` example, and that both call sites look the alias up first.

pmoust:
Folded to one commit, per the commit-shape note on #140. Same tree; the message is now one account of the finished change rather than three steps, including the resolved-name fact and the two-function API.

ananos (approved):
`Reserved`/`ReservedFor` split verified, and the doc comment carries the claude vs claude-code example plus the note that both call sites resolve the alias first. A caller can no longer reach the unscoped rule by forgetting an argument, which was the whole worry.

PR #135, opened 2026-09-04
fix(agent): refuse a trailing argument in every subcommand

### LINKED ISSUE #126: Refuse trailing arguments in brig agent subcommands instead of ignoring them
## Describe the bug

The `brig agent` subcommands accept and silently ignore trailing arguments,
where the top-level verbs refuse them. `brig ls` and `brig reset` refuse a
stray argument, the run verbs refuse one through `rejectTail`, and
`brig agent bogus` is refused as an unknown subcommand, so this is the one
place a stray word is dropped.

## To reproduce

1. `brig agent ls extra`
2. `brig agent edit a b`
3. `brig agent import a b`

## Expected behavior

A usage error naming the argument that has no use, matching `brig ls extra`:

```
brig: unexpected argument "extra"; `brig ls` takes no arguments
```

## Actual behavior

`brig agent ls extra` prints the listing and ignores `extra`. `brig agent edit
a b` and `brig agent import a b` act on the first word and drop the second.

## Environment

- Host OS: macOS 26.3.1
- Architecture: arm64
- Installed how: `make build`

Found while reviewing #118, which introduces the `agent` noun group. Not a
blocker for that PR. Filed separately since #118'"'"'s aim is one grammar across
the management commands, and this is the remaining inconsistency in it.

pmoust:
The `brig agent` subcommands accepted and silently ignored a trailing argument, where every other verb refuses one. `brig agent ls extra` printed the listing and dropped `extra`; `brig agent edit a b` and `brig agent import a b` acted on the first word and dropped the second.

They now refuse it with the same message shape `brig ls` uses, exit 2:

```
brig: unexpected argument "extra"; `brig agent ls` takes no arguments
brig: unexpected argument "b"; `brig agent edit` takes one agent and nothing more
```

`show`, `new`, `export` and `rm` already refused through `parseExportLine` and `nameAndYes`; the comment on the new helper says why they are not routed through it. The deprecated `list` and `load` behave like the verbs they map to and still print their notice.

Checked on a build of this branch: the three lines above and the deprecated spelling. Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`.

Fixes: #126

### DIFF
--- a/cmd/brig/main.go
+++ b/cmd/brig/main.go
@@ -1761,6 +1761,9 @@ func agentCmd(args []string) error {
 		fmt.Print(agentUsage)
 		return nil
 	case "ls":
+		if err := rejectAgentTail("brig agent ls", "takes no arguments", args[1:]); err != nil {
+			return err
+		}
 		return listProfiles()
 	case "show":
 		err = showAgent(args[1:])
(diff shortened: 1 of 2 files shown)


ananos (approved):
Thanks -- naming `brig agent ls` for the retired `list` spelling reads odd at first, but the deprecation notice above it already points the reader there, so the two lines agree. Good to have the whole subcommand set covered in one test.

PR #137, opened 2026-09-04
feat(cli): add brig logs <ref> to read a sandbox's log

### LINKED ISSUE #25: Add `brig logs <session>`
### Problem

After `brig run --detach`, or after a boot that never became ready, there is
no brig command that shows what happened. brig's own advice is to leave brig:
when a sandbox will not come up it prints the runtime's logs command,
`hull logs <sandbox>` or `nerdctl logs <sandbox>`. That asks the user for two
facts brig has and they do not: the sandbox name, which is not the session
name, and which runtime is underneath.

The user-mode gateway on `hvi` writes its own log next to its socket. No
command points at it, so a network failure at boot is invisible unless someone
already knows the file is there.

### Proposed solution

Add `brig logs <session>`:

```
brig logs api
brig logs api --follow
brig logs api --since 10m
brig logs --gateway
```

- Resolve the session to its sandbox and its runtime, then stream the
  runtime's logs. The interface already carries a per-runtime logs command,
  which is the seam this uses.
- `--gateway` prints the gateway log instead, for the network path.
- Keep the existing hint on a failed boot, and make it name `brig logs`
  first.

Filter terminal control sequences out of the stream by default, and add
`--raw` for the user who wants the bytes. brig stays in the middle of this
stream, unlike `brig run`, so this is one of the few places it can filter at
all. hull already filters in its own `logs` for the same reason.

### Alternatives considered

Keep pointing at the runtime. It costs nothing to maintain, and it asks the
user to learn the sandbox naming scheme and the runtime layer to read one log,
which is what brig exists to hide.

None. It reads output that already exists on the host.

### LINKED ISSUE #136: Class a missing sandbox as exit 3 on rm and logs, not the runtime's exit 1
## Describe the bug

`rm` and `logs` hand the runtime's own error back when the sandbox does not
exist, and exit 1. README documents 3 as "no such profile or sandbox -- the
name resolves to nothing", which is this case.

## To reproduce

With no sandbox for `claude`:

```
$ brig rm claude
brig: /opt/homebrew/bin/hull rm brig-claude-code: exit status 1: error: instance not found: brig-claude-code
$ echo $?
1

$ brig logs claude
error: instance not found: brig-claude-code
brig: exit status 1
$ echo $?
1
```

## Expected behavior

```
brig: no sandbox for claude. `brig ls` lists them
$ echo $?
3
```

Both verbs know the sandbox name they are about to use and can ask the
runtime's `List` whether it exists before handing over. `stop` already treats a
missing sandbox as not a failure and exits 0, which is right for `stop` and
wrong here: there is nothing to remove and nothing to read.

## Environment

- Host OS: macOS 15.6
- Architecture: arm64
- brig version: `brig dev`, built from source at the tip of #25's branch
- Installed how: `go build`

Found while landing #25. `logs` copies what `rm` already did, so this is one
fix across two verbs rather than a defect in #25.

pmoust:
## Summary

After `brig run -d`, or a boot that never became ready, no brig command showed what happened. brig's only advice was to leave brig for `hull logs <sandbox>` or `nerdctl logs <sandbox>`, which asks the reader for two facts brig has and they do not: the sandbox name, which is not the ref, and which runtime is underneath. The gateway wrote its own log beside its socket and nothing pointed at it.

`brig logs <ref>` resolves the ref to its sandbox and runtime the way every other ref'd verb does, then streams. `--gateway` reads the shared gateway's log; `--gateway <ref>` reads the gateway serving that sandbox alone, which since #130 is the case this is most for. Control sequences are filtered by default, `--raw` hands back the bytes. Three commits, each describing one change and each building and testing on its own.

## Related issues

Closes #25, closes #136.

## Changes

- `Runtime` gains `Logs(LogsSpec)` beside `LogsHint`, in both adapters. The runtime's own stderr goes to a separate `Err` writer, filtered alike, so `brig logs claude > out.txt` leaves a diagnostic on the terminal rather than in the file.
- `brig logs <ref> [--follow] [--tail N] [--raw]`. `--since` from #25 is dropped: hull offers `--follow` and `--tail` only (checked against `hull 0.1.0-rc21`), and brig promises only what both runtimes have.
- `brig logs --gateway [<ref>]`. The isolated log does not survive a stop; `clearGatewayRecord` is untouched. A sandbox on the shared gateway is told so from its resolved network rather than offered a guess.
- The boot-failure hint names `brig logs <ref>` first and the runtime's own command second.
- **`brig rm` and `brig logs` exit 3 on a sandbox that does not exist**, asking the runtime's `List` first, naming the ref by its slug. A `List` that fails stays exit 4. `stop` is untouched. `rm` does not prune the index on that path: nothing was removed, so nothing about the index has been settled.
- Docs: README's command table, `docs/runtimes.md` (no longer says `hull logs` is never run; argv listing gains it), `docs/troubleshooting.md` (the hint, verbatim), `docs/security.md` (`brig logs` stays in the middle of the stream and filters; `--raw` re-opens the surface that page describes).

## Checklist

- [x] `script/smoke.sh` passes, with new cases for the argv brig hands the runtime, the missing-sandbox exits, and the ref round-trip
- [x] I have run `go test ./... -race`
- [x] I have exercised the change against a real runtime: `brig logs --gateway`, `brig logs nosuch@x` (3), `brig logs claude` on a missing sandbox (3), `brig rm claude -n Refactor` naming the slug, all on hull 0.1.0-rc21
- [x] I have updated the affected docs (README, `docs/security.md`, `docs/runtimes.md`, `docs/troubleshooting.md`)

Touches `internal/wrap` only in `Remove`, which keeps its index bookkeeping; the not-found path now leaves the index alone, pinned by `TestRemoveSandboxKeepsTheIndexOnMissing`. Nothing here reads or forwards a credential, and `brig logs` reads host-side files the runtime already writes.

### DIFF
--- a/cmd/brig/logs.go
+++ b/cmd/brig/logs.go
@@ -103,7 +103,7 @@ func logsCmd(args []string) error {
 	if err != nil {
 		return err
 	}
-	return streamLogs(rt, name, o, os.Stdout)
+	return logsFor(rt, name, ref, o, os.Stdout)
 }
 
 // resolveSandbox turns a ref into the runtime it runs on and the sandbox name
@@ -136,6 +136,28 @@ func resolveSandbox(ref string) (*wrap.Config, runtime.Runtime, string, error) {
 		return nil, nil, "", err
 	}
 	return cfg, rt, cfg.VMName, nil
+}
+
+// logsFor confirms the runtime has the sandbox before streaming its log.
+//
+// There is nothing to read from a sandbox that is not there, which the exit
+// table calls a not-found (3) rather than a log stream that started and failed
+// (1). So the runtime is asked before the stream is opened, and the ref the
+// reader typed -- not the sandbox name -- is what the not-found names. A List
+// that itself fails comes back as is (exit 4, the runtime unavailable), not as
+// absence: the two are different facts. See sandboxPresent.
+//
+// Split from logsCmd so a test can drive it with a runtime double, the way
+// streamLogs is split for the same reason.
+func logsFor(rt runtime.Runtime, name, ref string, o logsOptions, out io.Writer) error {
+	present, err := sandboxPresent(rt, name)
+	if err != nil {
+		return err
+	}
+	if !present {
+		return noSandboxf(ref)
+	}
+	return streamLogs(rt, name, o, out)
 }
 
 // streamLogs hands the runtime the log request, filtering terminal control
--- a/cmd/brig/logs.go
+++ b/cmd/brig/logs.go
@@ -1,0 +1,313 @@
+package main
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/brig-sh/brig/internal/profile"
+	"github.com/brig-sh/brig/internal/runtime"
+	"github.com/brig-sh/brig/internal/session"
+	"github.com/brig-sh/brig/internal/wrap"
+)
+
+// logsOptions is what `brig logs` was asked for.
+type logsOptions struct {
+	follow bool
+	// tail is how many trailing lines to show, or -1 for all -- the default,
+	// and what both runtimes read as "all".
+	tail int
+	raw  bool
+}
+
+// logsCmd streams a sandbox's log, the one thing no brig command pointed at
+// before: after `brig run -d`, or a boot that never became ready, brig's only
+// advice was to leave brig for `hull logs <sandbox>` or `nerdctl logs
+// <sandbox>` -- which asks the reader for the sandbox name, which is not the
+// ref, and which runtime is underneath, both of which brig has and they do not.
+//
+// It parses its own flags rather than going through the run line, because
+// --gateway takes no ref: the gateway is one per host, and the run line's
+// parser requires a ref to resolve a profile at all.
+func logsCmd(args []string) error {
+	o := logsOptions{tail: -1}
+	var gateway bool
+	ref := ""
+	for i := 0; i < len(args); i++ {
+		a := args[i]
+		switch {
+		case a == "--follow":
+			o.follow = true
+		case a == "--raw":
+			o.raw = true
+		case a == "--gateway":
+			gateway = true
+		case a == "--tail":
+			i++
+			if i >= len(args) {
+				return usagef("--tail needs a number, for example `--tail 100`")
+			}
+			n, err := strconv.Atoi(args[i])
+			if err != nil {
+				return usagef("--tail needs a number, not %q", args[i])
+			}
+			o.tail = n
+		case strings.HasPrefix(a, "--tail="):
+			raw := a[len("--tail="):]
+			n, err := strconv.Atoi(raw)
+			if err != nil {
+				return usagef("--tail needs a number, not %q", raw)
+			}
+			o.tail = n
+		case strings.HasPrefix(a, "-"):
+			return usagef("unknown flag %q for `brig logs`", a)
+		default:
+			if ref != "" {
+				return usagef("unexpected argument %q; `brig logs` takes one ref", a)
+			}
+			ref = a
+		}
+	}
+
+	// The gateway log is per host, so it takes no ref -- and naming one is a
+	// mistake to say, not a session to go and resolve.
+	if gateway {
+		if ref != "" {
+			return usagef("`brig logs --gateway` names no session; the gateway is per host, "+
+				"not per sandbox. Drop %q", ref)
+		}
+		return gatewayLogs(os.Stdout, o)
+	}
+	if ref == "" {
+		return usagef("brig logs needs a ref, for example `brig logs claude`. " +
+			"`brig logs --gateway` reads the host's gateway log instead")
+	}
+
+	// Resolved the way every other ref'd verb resolves: the ref names the agent
+	// and the session, the profile picks the runtime, and Load derives the
+	// sandbox name from the two.
+	r, err := session.ParseRef(ref)
+	if err != nil {
+		// Classed as a usage error, the same way split classes the identical
+		// ParseRef failure on the run line: nothing ran, so this is a token in
+		// the wrong place rather than a run that started and failed.
+		return usagef("%s", err)
+	}
+	t, ok := profile.Lookup(r.Agent)
+	if !ok {
+		return notFoundf("unknown profile %q. `brig agent ls` lists them", r.Agent)
+	}
+	rt, err := runtime.DetectFor(runtime.Preference{Bin: t.RuntimeBin})
+	if err != nil {
+		return err
+	}
+	cfg, err := wrap.Load(t, wrap.Options{Name: r.Label, Verbosity: verbosity}, rt)
+	if err != nil {
+		return err
+	}
+	return streamLogs(rt, cfg.VMName, o, os.Stdout)
+}
+
+// streamLogs hands the runtime the log request, filtering terminal control
+// sequences on the way out unless the raw bytes were asked for.
+//
+// A function of its own, taking the runtime rather than reaching for it, so a
+// test can hand it a double and assert that --follow and --tail reach the
+// adapter with the values the line carried.
+func streamLogs(rt runtime.Runtime, name string, o logsOptions, out io.Writer) error {
+	return rt.Logs(runtime.LogsSpec{
+		Name:   name,
+		Follow: o.follow,
+		Tail:   o.tail,
+		Out:    filtered(out, o.raw),
+		Err:    filtered(os.Stderr, o.raw),
+	})
+}
+
+// gatewayLogs prints the shared gateway's own log. brig starts that gateway for
+// the hvi backend and writes its output to a file beside the socket; a network
+// failure at boot lands there and nothing pointed at it until now.
+func gatewayLogs(out io.Writer, o logsOptions) error {
+	path, err := runtime.GatewayLogPath()
+	if err != nil {
+		return err
+	}
+	return streamGatewayFile(filtered(out, o.raw), path, o.follow, o.tail)
+}
+
+// gatewayFollowPoll is how often --follow re-reads the gateway log for bytes
+// appended since. It is a file brig does not own the writer of, so there is
+// nothing to wait on but the file itself.
+const gatewayFollowPoll = 500 * time.Millisecond
+
+// streamGatewayFile writes the gateway log to out, tailing it under --follow.
+//
+// A missing file is not a fault to fail loudly over: the gateway log appears
+// only once a sandbox has needed the host network gateway, so on a machine that
+// has booted only vz sandboxes there is simply nothing yet.
+func streamGatewayFile(out io.Writer, path string, follow bool, tail int) error {
+	f, err := os.Open(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return notFoundf("no gateway log at %s yet -- it appears once a sandbox "+
+				"first needs the host network gateway", path)
+		}
+		return err
+	}
+	defer func() { _ = f.Close() }()
+
+	data, err := io.ReadAll(f)
+	if err != nil {
+		return err
+	}
+	if tail >= 0 {
+		data = lastLines(data, tail)
+	}
+	if _, err := out.Write(data); err != nil {
+		return err
+	}
+	if !follow {
+		return nil
+	}
+	// Interrupting this leaves the gateway running: it serves every sandbox on
+	// the host, and reading its log is not owning it. Ctrl-C reaches this
+	// process and ends it, which is the whole of what stopping should do.
+	for {
+		time.Sleep(gatewayFollowPoll)
+		more, err := io.ReadAll(f)
+		if err != nil {
+			return err
+		}
+		if len(more) > 0 {
+			if _, err := out.Write(more); err != nil {
+				return err
+			}
+		}
+	}
+}
+
+// lastLines keeps only the final n lines of data, for --tail on a file the
+// runtime is not the one reading. n <= 0 keeps nothing; fewer lines than asked
+// keeps them all.
+func lastLines(data []byte, n int) []byte {
+	if n <= 0 {
+		return nil
+	}
+	lines := bytes.SplitAfter(data, []byte("\n"))
+	// SplitAfter leaves a trailing empty piece after a final newline; it is not
+	// a line, so it does not count toward n.
+	if len(lines) > 0 && len(lines[len(lines)-1]) == 0 {
+		lines = lines[:len(lines)-1]
+	}
+	if len(lines) <= n {
+		return data
+	}
+	return bytes.Join(lines[len(lines)-n:], nil)
+}
+
+// filtered wraps a writer to strip terminal control sequences, or hands it back
+// untouched when the raw bytes were asked for.
+func filtered(out io.Writer, raw bool) io.Writer {
+	if raw {
+		return out
+	}
+	return &controlFilter{w: out}
+}
+
+// filterState is where the control-sequence filter is in a sequence it has
+// started to recognise. Held on the writer rather than reset per Write so a
+// sequence split across two writes -- which streaming makes ordinary -- is
+// still filtered whole.
+type filterState int
+
+const (
+	fText    filterState = iota // ordinary text
+	fEsc                        // just saw ESC
+	fCSI                        // in ESC [ ... , waiting for the final byte
+	fOSC                        // in ESC ] ... , waiting for BEL or ST
+	fOSCEsc                     // in an OSC and saw ESC, waiting for the \ of ST
+	fCharset                    // ESC ( or ) , which consumes one more byte
+)
+
+// controlFilter strips terminal control sequences from a log as it passes
+// through. brig sits in the middle of this stream -- unlike `brig run`, which
+// hands the terminal straight to the agent -- which is what makes filtering
+// possible here at all. A runtime replays into its log whatever a program
+// wrote, cursor moves and colour and title-setting included, and a log meant to
+// be read back is text, not a terminal to drive. --raw skips this.
+//
+// Newline, tab and carriage return are kept: they are the layout of the text,
+// not control of a terminal. Every other C0 control and DEL is dropped.
+type controlFilter struct {
+	w  io.Writer
+	st filterState
+}
+
+func (f *controlFilter) Write(p []byte) (int, error) {
+	out := make([]byte, 0, len(p))
+	for _, b := range p {
+		switch f.st {
+		case fText:
+			switch {
+			case b == 0x1b:
+				f.st = fEsc
+			case b == '\n' || b == '\t' || b == '\r':
+				out = append(out, b)
+			case b < 0x20 || b == 0x7f:
+				// A C0 control or DEL with no place in read-back text.
+			default:
+				out = append(out, b)
+			}
+		case fEsc:
+			switch b {
+			case '[':
+				f.st = fCSI
+			case ']':
+				f.st = fOSC
+			case '(', ')':
+				f.st = fCharset
+			default:
+				// A lone ESC-something, e.g. ESC c reset: drop the pair.
+				f.st = fText
+			}
+		case fCSI:
+			// The final byte of a CSI is in 0x40-0x7e; the bytes before it are
+			// parameters and intermediates.
+			if b >= 0x40 && b <= 0x7e {
+				f.st = fText
+			}
+		case fOSC:
+			switch b {
+			case 0x07: // BEL terminates an OSC
+				f.st = fText
+			case 0x1b: // ESC may begin the ST that terminates it
+				f.st = fOSCEsc
+			}
+		case fOSCEsc:
+			// ESC \ is the string terminator; anything else was ESC inside the
+			// OSC body, and either way the sequence is over as far as text goes.
+			f.st = fText
+		case fCharset:
+			f.st = fText
+		}
+	}
+	if _, err := f.w.Write(out); err != nil {
+		return 0, err
+	}
+	// The whole input was consumed, sequences included, so report it fully
+	// written: a short count with no error is what makes io.Copy report
+	// ErrShortWrite, and this writer drops bytes by design.
+	return len(p), nil
+}
+
+// stripControl runs a complete byte slice through the filter, for a caller --
+// a test above all -- that has all the bytes at once.
+func stripControl(in []byte) []byte {
+	var buf bytes.Buffer
+	f := &controlFilter{w: &buf}
+	_, _ = f.Write(in)
+	return buf.Bytes()
+}
--- a/cmd/brig/logs.go
+++ b/cmd/brig/logs.go
@@ -30,8 +30,9 @@ type logsOptions struct {
 // ref, and which runtime is underneath, both of which brig has and they do not.
 //
 // It parses its own flags rather than going through the run line, because
-// --gateway takes no ref: the gateway is one per host, and the run line's
-// parser requires a ref to resolve a profile at all.
+// --gateway takes an optional ref: the shared gateway is one per host and is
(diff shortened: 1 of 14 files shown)


ananos (commented):
Really glad to see this -- the gateway writing a boot's network failure to a file nothing named was a genuine hole.

One wrinkle from what landed yesterday in #130: the gateway is no longer one per host. A sandbox on `--network isolated`, or any sandbox carrying a policy, gets its own gateway writing to `sandbox-<name>.log`. Two notes inline.

cmd/brig/logs.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+			ref = a
+		}
+	}
+
+	// The gateway log is per host, so it takes no ref -- and naming one is a
+	// mistake to say, not a session to go and resolve.
+	if gateway {
+		if ref != "" {
+			return usagef("`brig logs --gateway` names no session; the gateway is per host, "+
```
ananos: This message is no longer true after #130: an isolated sandbox has a gateway of its own, so "the gateway is per host" holds only for the shared one.

Awkwardly, the isolated case is the one this feature is most for -- if that gateway fails to start, its log is the only evidence, and `--gateway` cannot reach it today. Would it make sense for `brig logs --gateway <ref>` to read that sandbox's gateway log, and keep the bare `--gateway` for the shared one?

internal/runtime/gateway.go
```diff
@@ -131,6 +131,20 @@ func gatewaySocket() (string, error) {
 	return filepath.Join(dir, "gateway-"+socketTag(gatewaySubnet)+".sock"), nil
 }
 
+// GatewayLogPath is where the shared gateway's own output is written, beside
+// its socket. ensureGateway opens it, and `brig logs --gateway` points at it:
+// the gateway logs a boot's network failure to a file no command named until
+// now, so a network that never came up was invisible unless you already knew
+// the file was there. Derived from the socket by the same rule startGateway
+// uses, so the writer and the reader cannot drift onto two paths.
+func GatewayLogPath() (string, error) {
```
ananos, line 149: This resolves the shared socket only, so it answers for the shared gateway and nothing else. `isolatedSocket(name)` plus the same `gatewayLogPath` would give the per-sandbox one, if we go that way.

One thing to know before we do: `clearGatewayRecord` deletes the isolated log when the gateway is confirmed stopped, which is what keeps `~/.brig` from growing a file per sandbox ever isolated. So a reader for those logs and that deletion need to agree -- either the log survives a stop, or `brig logs --gateway <ref>` only works while the sandbox is up. Happy to take the cleanup side if you prefer the log to stick around.

cmd/brig/logs.go
pmoust: Done in 81e31a3, both forms: bare `--gateway` reads the shared gateway's log, `--gateway <ref>` reads that sandbox's own. The ref resolves to its sandbox the same way the non-gateway path does, through one `resolveSandbox`, and the usage error that refused the ref is gone. The comments that said the gateway is per host now say per host for the shared one and per sandbox for an isolated one.

internal/runtime/gateway.go
pmoust: `IsolatedGatewayLogPath(name)` beside `GatewayLogPath`, both through `gatewayLogPath`, in 81e31a3. I took the side where the log does not survive a stop: `clearGatewayRecord` is untouched, since a file per sandbox ever isolated is the growth you named, and the case this is for, a gateway that failed to start, leaves its log in place because nothing confirmed it stopped. When the file is absent the message says the sandbox's gateway is not running and its log goes with it, names the ref, and is a not-found so a script can tell it from a failure.

pmoust:
Folded, per the commit-shape note on #140: the bad-ref class fix is inside the `logs` commit now, and the isolated-gateway log stays as its own second commit since it builds and tests alone. Each of the two checked out and gated on its own.

ananos (dismissed):
Both forms of `--gateway` is the right answer, and I am glad you took the side where the log does not survive a stop -- a file per sandbox ever isolated was the growth I was worried about. The failed-to-start case still leaves its log in place because nothing confirmed the stop, which is exactly the case the feature is for.

`IsolatedGatewayLogPath` beside `GatewayLogPath`, both through `gatewayLogPath`, verified.

ananos:
Approved, and #138 is merged into this branch, so this PR now carries both commits -- that part is as intended for the stack.

It has gone conflicting against `main` though. Everything that landed this morning went in as rebase-merges, so #133, #134 and #146 are on `main` under new SHAs and this branch predates them; `main.go` and `gateway.go` are touched on both sides.

Needs a rebase onto the current `main` and then it can go in as it stands. I have left the branch alone rather than rebasing it for you.

internal/wrap/run.go
```diff
@@ -630,9 +648,23 @@ func (c *Config) Stop() error {
...
 	err := c.Runtime.Remove(c.VMName)
+	c.Forget()
+	return err
+}
+
+// Forget clears the index entries that name this sandbox: the workspace record
+// and the slug claim. Both are idempotent, so it is safe whether or not the
+// sandbox still exists.
+//
+// A method of its own, so that rm's not-found path can prune too. When the
```
asapranidis: This says the extraction exists so rm's not-found path can prune too, and that clearing it there leaves no stale record. `removeSandbox` refuses to do that, at length, and `TestRemoveSandboxKeepsTheIndexOnMissing` pins the refusal. `grep -rn Forget` finds one call, line 651 inside `Remove`.

Leftover of #138's earlier design, whose body still describes the pruning version. Either inline it back into `Remove` or rewrite the comment. No runtime effect, but it is a false comment sitting on the exact decision commit 3 reversed.

cmd/brig/main.go
```diff
@@ -1609,6 +1616,62 @@ func takeAll(args []string) (rest []string, all bool) {
...
+func removeSandbox(cfg *wrap.Config, ref string) error {
+	present, err := sandboxPresent(cfg.Runtime, cfg.VMName)
+	if err != nil {
+		return err
+	}
+	if !present {
+		// Not pruned here. "Not in the list" is not always "gone": hull.List
+		// falls back to a plain `ps` on a hull without `ps -a`, and that listing
+		// carries only the running instances, so a stopped sandbox reads as
+		// absent while it is still in hull's store. Forgetting its index entry on
```
asapranidis: The premise does not hold for hull. `hull ps -a` is not "not universal", it does not exist in any version:

```
$ hull ps -a
error: flag provided but not defined: -a     (exit 1)
```

So the fallback is the only path, and plain `hull ps` lists everything: `cmd/hull/ps.go` -> `store.ListInstances()` reads every instance's `state.json` with no status filter and renders stopped ones as `stopped`. hull's own `Usage: "list running instances"` string is wrong, which is likely where this came from.

Good news is that the gate therefore does not refuse a stopped sandbox. But not-pruning now rests on something untrue, and this comment is what the next change will be measured against. Same wording in the test at `notfound_test.go` and in the commit message.

internal/runtime/hull.go
```diff
@@ -780,6 +780,25 @@ func firstLines(s string, n int) string {
...
+	args := []string{"logs"}
+	if spec.Follow {
+		args = append(args, "--follow")
+	}
+	// hull's own default is -1 (all), so spec.Tail passes straight through.
+	args = append(args, "--tail", strconv.Itoa(spec.Tail), spec.Name)
+	cmd := exec.Command(h.bin, args...)
+	cmd.Env = mergeEnv(telemetryEnv(false))
+	cmd.Stdout = spec.Out
+	cmd.Stderr = spec.Out
```
asapranidis: For hull this puts diagnostics on stdout. `hull logs <missing>` writes `error: instance not found: ...` to stderr and exits 1; the log itself goes to stdout. So this stream carries only hull's own errors, never guest output, and `brig logs claude > out.txt` on a hull-side failure (a race past the List gate, a store permission error) writes the cause into `out.txt`, prints nothing to the terminal, and exits 1 on a bare `exit status 1` -- the shape #136 was filed against. `os.Stderr` here keeps the split.

internal/runtime/hull.go
```diff
@@ -780,6 +780,25 @@ func firstLines(s string, n int) string {
 	return strings.Join(lines, "; ")
 }
 
+// Logs streams the instance's log. hull writes it to its own stdout, so the
+// stream is handed straight to spec.Out; --follow and --tail are the two flags
+// hull offers, and brig offers only what both runtimes share. The instance is
+// the positional. Not counted: reading a log is not the user action telemetry
+// attributes, the same reason ps and rm read false.
+func (h *hull) Logs(spec LogsSpec) error {
```
asapranidis, line 796: `docs/runtimes.md:96` says `hull logs <name>` "appears in brig's error output as a suggestion and is never run". True until this method. The argv listing on that page also does not carry `logs`.

cmd/brig/main.go
```diff
@@ -434,9 +437,13 @@ func run(args []string) error {
...
 	switch verb {
 	case "stop":
+		// Deliberately not gated on the sandbox existing: stopping one that was
+		// never running is the end state stop asks for, not a failure. See
+		// Config.Stop. rm and logs below are the opposite -- there is nothing to
+		// remove and nothing to read -- which is why only they check first.
 		return cfg.Stop()
 	case "rm":
-		return cfg.Remove()
+		return removeSandbox(cfg, session.Ref{Agent: profileName, Label: cfg.RawName}.String())
```
asapranidis: `RawName` is the name as typed, `Slug` is the canonical form, and they differ for any non-slug `-n`. `ParseRef` refuses a label that is not already its slug, so `brig rm claude -n Refactor` on a missing sandbox prints `no sandbox for claude@Refactor` -- a ref brig itself rejects, which is what `noSandboxf`'s comment says it avoids. The index key is `sessionKey(Profile.Name, Slug)`, and `logHint` below does it canonically through `RefOfSandbox`. `cfg.Slug` fixes it.

cmd/brig/logs.go
```diff
@@ -0,0 +1,385 @@
...
+// ~/.brig from growing a file for every sandbox ever isolated. A gateway that
+// never came up was never confirmed stopped, so the failure this is for leaves
+// its log in place. Absent means the gateway is not running, and the message
+// says so rather than implying the log was somewhere else.
+func isolatedGatewayLogs(out io.Writer, o logsOptions, ref, name string) error {
+	path, err := runtime.IsolatedGatewayLogPath(name)
+	if err != nil {
+		return err
+	}
+	missing := notFoundf("no gateway log at %s: the gateway serving %s is not running, "+
```
asapranidis, line 213: For a running sandbox on the shared gateway -- the common case for this flag -- the leading sentence is false, and the true reading ("was never isolated") is third in a list. `resolveSandbox` already loads the whole Config and throws away `cfg.Network`, which says whether this run resolved to isolated. With it the message can say `claude is on the shared gateway; brig logs --gateway reads that one`.

internal/wrap/run.go
```diff
@@ -382,8 +382,7 @@ func (c *Config) EnsureRunning(set creds.Set) error {
 	// to find out.
 	c.rememberSession()
 	if !c.waitReady() {
-		return fmt.Errorf("sandbox did not become ready; check '%s'",
-			c.Runtime.LogsHint(c.VMName))
+		return fmt.Errorf("sandbox did not become ready; check '%s'", c.logHint())
```
asapranidis, line 385: `docs/troubleshooting.md:14` and `:42` quote this message verbatim in its old form, and line 24 ("Check the runtime's own logs, which is what the message points at") is now the second half of the advice. That page is where a reader lands for exactly this feature.

cmd/brig/main.go
```diff
@@ -49,6 +49,7 @@ usage:
   brig rm   <ref>                                stop and remove the sandbox
   brig rm   --all                                stop and remove every brig sandbox
   brig ls   [-q]                                 list sandboxes; -q prints the refs
+  brig logs <ref> [--follow] [--tail N] [--raw]  stream the sandbox log (--gateway: that sandbox's gateway log; alone: the shared one)
```
asapranidis: 134 columns. Every other row here is <= 85, and `brig info`, `brig rm --all` and `brig policy attach` wrap their descriptions onto continuation lines. Also missing from README's Commands table, which lists every other verb down to `brig version`.

script/smoke.sh
```diff
@@ -804,6 +804,27 @@ grep -q '^argv: stop brig-claude-code' "$STUB_LOG" \
...
+"$WORK/brig" run claude -d > /dev/null 2>&1
+: > "$STUB_LOG"
+"$WORK/brig" logs claude > /dev/null 2>&1
+grep -q '^argv: logs --tail -1 brig-claude-code' "$STUB_LOG" \
+  && ok "logs streams the sandbox log through the runtime" \
+  || bad "logs streams the sandbox log -- got: $(grep '^argv:' "$STUB_LOG")"
+: > "$STUB_LOG"
+"$WORK/brig" logs claude --follow > /dev/null 2>&1
+grep -q '^argv: logs --follow --tail -1 brig-claude-code' "$STUB_LOG" \
+  && ok "logs --follow passes --follow to the runtime" \
```
asapranidis, line 831: The round-trip loop below ("every ref the listing prints is a word every verb takes") drives run, sh, info, stop, rm and the verbless form. `logs` is this PR's new ref'd verb and is not in it; one more `takes_ref` line.

asapranidis (commented):
Gates on `7a40a9b`: `go build`, `go vet`, `gofmt -l`, `go test -race ./...` all clean (12 packages ok, `internal/secret` included). `script/smoke.sh` PASS here and PASS on `origin/main`, no regression. Rebase is done: merge-base with `origin/main` is main's tip.

Two tests proved to bite, not tautologies: reverting `IsolatedGatewayLogPath` to the shared socket fails `TestIsolatedGatewayLogsReadsTheFile` (reads `shared gateway up`), and adding `cfg.Forget()` to rm's not-found path fails `TestRemoveSandboxKeepsTheIndexOnMissing`.

Verified holding: `IsolatedGatewayLogPath` goes through the same `isolatedSocket(name)` the writer uses, hash fallback included, so reader and writer cannot drift; `clearGatewayRecord` does remove the log with the pid and spec; and the filter is load-bearing rather than duplicated, since hull's own `guestTerminalWriter` skips filtering when stdout is not a tty, which under brig it never is, so `--raw` does deliver raw bytes. hull rc21 offers exactly `--follow` and `--tail int` (-1 = all) and no `--since`, and brig's exact argv `logs --follow --tail -1 <name>` parses and fails only at instance lookup.

**Description is stale.** "Second commit: a bad ref on `logs` exited 1..." is folded into commit 1, and the real second commit (the isolated gateway log) is never mentioned. "**Known gap, not this PR's** ... Filed as #136" is fixed by commit 3 on this branch, and the trailer still says only `Fixes: #25`. The body also never says `brig rm` changed behaviour or that `internal/wrap` gained an exported `Config.Forget`, and the template's sections and checklist are absent.

**Docs untouched, four pages now wrong:** `README.md`'s Commands table has no `brig logs` row; `docs/troubleshooting.md:14,42` quote the old boot message; `docs/runtimes.md:96,119` still say the runtime's logs command is only ever suggested; `docs/security.md:654-664` says brig never stays in the middle of that stream, which `brig logs` now does. `--raw` re-opens the OSC 52 / DCS surface that page names, and hull gates its own escape hatch behind a warning.

Notes inline. Not blocking, happy to approve once you have picked what to take.

ananos:
Two notes on where this stands.

**The rebase went stale underneath you, through no fault of the rebase.** @asapranidis was right that the merge-base was main's tip when he checked at 12:40. #158 landed after that, so main is now `a3ce070` and the merge-base here is `4bf5409`, nine commits back. The overlap is only `cmd/brig/main.go` and `script/smoke.sh`, both touched by the completion work, so it should be a mechanical rebase rather than a real disagreement.

**I want to second the `docs/security.md` point, because it lands on a paragraph I asked to be changed on another PR.**

That page currently says:

> `hull exec` and `hull logs` do filter, because hull stays in the middle of that stream; brig deliberately does not stay.

After this PR that is wrong twice over, and the more interesting half is that it undersells what you built. `brig logs` *does* stay in the middle, and `streamLogs` filters by default -- so the honest sentence is stronger than the one on the page, not weaker. `--raw` is then the deliberate opt-out, and it re-opens exactly the OSC 52 / DCS surface the paragraph spends a column describing, so it deserves naming there rather than only in the flag's help.

Worth knowing while you edit it: #141 rewrites that same paragraph. It carries #142's commit -- that merged into #141's branch rather than into main -- which scopes the `syscall.Exec` claim to "except under `--json`". So the two of you are editing the same six lines from different directions. No textual conflict today, since this PR does not touch the file, but whoever lands second needs a paragraph that is true about `--json`, about `brig logs` filtering, and about `--raw`. Might be simplest for whichever lands first to leave the other a clean seam.

The rest of @asapranidis's review is fair too, particularly the stale description -- the trailer still says only `Fixes: #25` while commit 3 closes #136.

internal/wrap/run.go
pmoust: Right, and it is the comment that is wrong rather than the code. `Forget` had exactly one caller, so it is inlined back into `Remove` in 8b48c4c and the method is gone. The two `ForgetSandbox`/`ForgetSlugClaim` calls now sit where the removal happens, under a comment that says removal is the only thing that clears them and that rm's not-found path leaves them alone because it removed nothing.

cmd/brig/main.go
pmoust: Confirmed on the installed hull: `hull ps -a` gives `flag provided but not defined: -a`, and plain `hull ps` reads every instance's state with no status filter. So the premise was untrue and the fallback is the only path.

Rewritten in 8b48c4c to the reason that actually holds: nothing was removed, so nothing about the index has been settled. The entry goes when a removal happens, in `Remove`, or when `ls` prunes against a real listing. Dropping it on one absence would put this path in the business of deciding a sandbox is gone, which the verb that removes it is better placed to judge. The test comment moved with it.

Worth noting `docs/runtimes.md:67` still documents the `ps -a` attempt and its fallback, which is accurate about brig's argv even though the flag never lands. Left as is.

internal/runtime/hull.go
pmoust: Fixed in 8b48c4c, and it is the same shape #136 was filed against, which makes it worse than a nit.

`LogsSpec` gains `Err`, kept apart from `Out` rather than pointed at `os.Stderr` directly: on a container runtime that stream carries the guest's stderr, which is log content a reader wants filtered too, so the caller wraps both halves in the same control-sequence filter. On hull it carries diagnostics and nothing else, so `brig logs claude > out.txt` now leaves the cause on the terminal and the log in the file.

internal/runtime/hull.go
pmoust: Fixed in 8b48c4c. That sentence now says `hull logs <name>` is what `brig logs <ref>` runs, and is also named in the error output when a sandbox will not come up. Added `hull logs [--follow] [--tail <n>] <name>` to the argv listing above it, which was missing the verb entirely.

cmd/brig/main.go
pmoust: Fixed in 8b48c4c: `cfg.Slug`, with a comment saying why. Checked on a build of the branch, `brig rm claude -n Refactor` on a missing sandbox now prints `no sandbox for claude@refactor`, which is a ref `ParseRef` accepts and what the index is keyed by.

cmd/brig/logs.go
pmoust: Fixed in 8b48c4c. `resolveSandbox` returns the `Config` now, so the shared case is answered from what the run resolved rather than guessed at: `claude is on the shared gateway, which has no log of its own per sandbox. \`brig logs --gateway\` reads the shared gateway's log`. The absent-log message keeps the isolated case and drops the "was never isolated" reading, which that branch can no longer reach.

internal/wrap/run.go
pmoust: Fixed in 8b48c4c. Both quotes carry the new hint verbatim, and the line about checking the runtime's own logs now reads `brig logs claude` with a sentence saying that is `hull logs` underneath and why the message still names that spelling.

cmd/brig/main.go
pmoust: Fixed in 8b48c4c: split across three rows, the widest 85, with `brig logs --gateway [<ref>]` given its own line since it takes no ref in one form. Added to README's Commands table too.

One nearby: `brig doctor [<agent>]` is 103 columns and predates this branch, so I left it. Worth a one-line follow-up.

script/smoke.sh
pmoust: Fixed in 8b48c4c: `takes_ref "brig logs" logs "$ref"`, placed after `info` while the sandbox from `run -d` is still up and before `stop` takes it down. Passing on the stub.

ananos:
Good progress -- README, `docs/runtimes.md` and `docs/troubleshooting.md` are all updated, and the branch is green here (`go vet`, `go test -race`, linux cross-build).

**`docs/security.md` is still untouched, though.** I checked against the branch's own base rather than against main, so this is not the diff being confused by main moving: `git diff <merge-base> pr137 -- docs/security.md` is empty. The page still says

> `hull exec` and `hull logs` do filter, because hull stays in the middle of that stream; brig deliberately does not stay.

which is now wrong in the direction that undersells you -- `brig logs` does stay, and `streamLogs` filters by default. `--raw` then turns that filtering off, which re-opens the OSC 52 / DCS surface the paragraph describes, and it is not named there.

Note that paragraph changed on main since your base: #141 landed and scoped the `syscall.Exec` sentence to "except under `--json`". Your base predates it, so the rebase will bring in a paragraph that already has one exception in it, and the edit is to add the second rather than to rewrite what you last saw.

While you are in there: main is 17 commits ahead of this branch's base, and the repo requires a branch to be up to date before it merges, so it needs the rebase anyway.

One small thing on `8b48c4c fix(cli): correct what review found in the logs verb` -- that is the shape we agreed to avoid on #140, and you have been folding everywhere else today, including on #154 an hour ago where the comment fix went into the commit that caused it. Same treatment here would leave three commits that each describe a change rather than a round of review. Entirely a history point, the content is right.

pmoust:
All three taken, and the branch is at `3645012` on current `main`.

**`docs/security.md`.** The paragraph now says `brig run` deliberately does not stay in the stream, `brig logs` does and filters by default, and `--raw` turns that off and hands you the bytes with the OSC 52 / DCS surface intact. It sits beside the `--json` exception the rebase brought in, as you said it would.

**The commit shape.** `8b48c4c` is gone. Its nine changes are folded into the commits they correct: the stderr split, the docs and the round-trip into the `logs` commit; the shared-gateway answer into the isolated-gateway commit; the slug, the not-pruning reasoning and the inlined `Forget` into the exit-3 commit. Each message describes its finished change. I checked the split by diffing the PR against its merge-base before and after: identical except the new `security.md` paragraph. Each commit checked out, built and vetted alone.

**Rebase.** On `731a3cf`, which now carries #154, #157 and #160, so CI is green rather than red on `main`'s old completion test.

The description is rewritten to the template, with `Closes #136` added, since the exit-3 commit fixes it here.

ananos (approved):
All three points closed, and the security.md paragraph is now the best version of itself:

> `hull exec` and `hull logs` do filter, because hull stays in the middle of that stream. `brig run` deliberately does not stay. `brig logs` does: it reads a log back rather than driving a terminal, so it filters control sequences by default, and `--raw` turns that off and hands you the bytes with the surface above intact.

Scoping the old claim to `brig run`, stating that `brig logs` filters, and tying `--raw` back to the OSC 52 / DCS paragraph directly above it is exactly right -- and it reads as the stronger promise it actually is, rather than the weaker one the page used to make. It also sits cleanly beside the `--json` exception #141 added, which is the reconciliation I was worried about.

Verified on this head: rebased and up to date with main, three semantic commits, all four doc pages touched, and `go vet` + `go test -race` (12 packages) + linux cross-build green.

Thanks for folding the history back too.

PR #138, opened 2026-09-04
fix(cli): make rm and logs exit 3 when the sandbox is absent

pmoust:
`brig rm claude` and `brig logs claude` with no sandbox for `claude` handed back the runtime's own error and exited 1. README documents 3 as "no such profile or sandbox", which is this case.

Both verbs now ask the runtime's `List` for the sandbox name before handing over. Absent:

```
brig: no sandbox for claude. `brig ls` lists them
```

exit 3, naming the ref typed rather than the sandbox name. Present: `Remove` and `Logs` as before. A `List` that itself fails comes back as is, exit 4, not folded into not-found; the two are different facts and the exit table keeps them apart.

`stop` is deliberately untouched and still exits 0 on a missing sandbox: that is the end state it asks for. A comment at the dispatch says so, and a test pins it.

`rm` on the not-found path still prunes the index entry, since it names a sandbox the user asked to be rid of. `Forget` is pulled out of `Remove` so both paths share it.

Stacked on #137 because it touches `logs.go`; GitHub retargets to `main` when that merges.

Checked on a build of this branch: `rm claude`, `logs claude` and `rm claude@refactor` exit 3 with the message above; `stop claude` exits 0. Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`.

Fixes: #136

### DIFF
--- a/cmd/brig/main.go
+++ b/cmd/brig/main.go
@@ -428,9 +428,13 @@ func run(args []string) error {
 	// keychain prompt that reading the host login may bring up.
 	switch verb {
 	case "stop":
+		// Deliberately not gated on the sandbox existing: stopping one that was
+		// never running is the end state stop asks for, not a failure. See
+		// Config.Stop. rm and logs below are the opposite -- there is nothing to
+		// remove and nothing to read -- which is why only they check first.
 		return cfg.Stop()
 	case "rm":
-		return cfg.Remove()
+		return removeSandbox(cfg, session.Ref{Agent: profileName, Label: cfg.RawName}.String())
 	}
 
 	set, err := cfg.BuildEnv()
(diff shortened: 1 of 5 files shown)


cmd/brig/main.go
```diff
@@ -1514,6 +1518,58 @@ func takeAll(args []string) (rest []string, all bool) {
...
+
+// sandboxPresent reports whether the runtime has a sandbox of this name at all,
+// running or stopped.
+//
+// It is how rm and logs tell "there is nothing here" -- a not-found, exit 3 --
+// from a runtime that could not be asked -- its own error, exit 4. The two are
+// different facts and the README's exit table keeps them apart, so a List that
+// fails is returned as is rather than folded into absence.
+func sandboxPresent(rt runtime.Runtime, name string) (bool, error) {
+	list, err := rt.List()
```
ananos, line 1557: `hull.List()` falls back to a plain `hull ps` when `ps -a` is not supported, and that listing carries only the running instances. On such a hull, `brig rm <a stopped sandbox>` takes the not-present branch: it calls `cfg.Forget()` and returns exit 3, while the instance is still in hull's store -- so we drop the index entry for a sandbox that is really there.

The exit code is not the worry, the `Forget()` is. Two options I can see: gate the forget on the removal having actually happened, or have `sandboxPresent` report "cannot tell" when the listing came from the fallback, and keep today's behaviour there. Slight preference for the first, it needs no new plumbing. What do you think?

ananos (commented):
Thanks -- exit 3 for an absent sandbox is the right call, and `noSandboxf` giving both verbs one message is good.

One thing worth sorting before this lands, inline below.

cmd/brig/main.go
pmoust: Took the first option, in cfde3f0. The not-found branch no longer forgets; the entry goes when a removal actually happens in `Remove`, or when `ls` prunes against a full listing. The test that asserted the prune now asserts the entry is kept, with the `ps -a` fallback as the reason in its comment.

pmoust:
Folded to one commit and rebased onto #137's rewritten tip, per the commit-shape note on #140. The message no longer says rm prunes on the not-found path; it says why it does not.

ananos (approved):
The not-found branch no longer forgets, and the reason is written into the code where the next reader will meet it. That was the one real bug in this batch -- dropping the index entry for a sandbox still in hull's store, on a hull whose `ps -a` falls back -- and it is properly closed. The test asserting the entry is kept, with the fallback named as the reason, is the right regression.

Issue #139, opened 2026-09-05
Check BRIG_RUNTIME_BIN the way a profile's runtimeBin is checked


pmoust:
## Describe the bug

`DetectFor` in `internal/runtime/runtime.go` takes `BRIG_RUNTIME_BIN` on
trust, while a profile's `runtimeBin` goes through `runtimeBinFromProfile`,
which checks that the path is on disk and executable and wraps
`ErrBadRuntime` when it is not. The sentinel's own comment says both cases
should reach it:

> An unknown BRIG_RUNTIME, or a runtimeBin that is not on disk or is not
> executable, wrap this.

Only the profile path does.

## To reproduce

```
$ BRIG_RUNTIME_BIN=/nonexistent brig run codex
brig: cannot tell whether the sandbox brig-codex is already running, so refusing to start a second one over it: /nonexistent ps: fork/exec /nonexistent: no such file or directory
$ echo $?
1
```

## Expected behavior

The same refusal a profile's bad `runtimeBin` gets, before anything is run,
exit 4:

```
brig: runtime unavailable: BRIG_RUNTIME_BIN is /nonexistent, which is not there
$ echo $?
4
```

## Environment

- Host OS: macOS 15.6
- Architecture: arm64
- brig version: `brig dev`, built from source
- Installed how: `go build`

## Logs and additional context

Found while landing #9. `brig doctor` looks the binary up itself so it does
not vouch for a path nothing can run, but every other verb still discovers it
at the first `ps`. The check `runtimeBinFromProfile` already makes is the one
to apply to the environment value too.

PR #140, opened 2026-09-05
feat(cli): add brig doctor to name the failing prerequisite

### LINKED ISSUE #9: Add `brig doctor`
### Problem

A first run can fail at the host OS version, at virtualization, at the runtime
binary, at the image pull, or at the secret store. The error arrives from the
layer underneath, in the words of that layer. #4 is one example: a macOS 14.5
host, and the failure surfaces as `dyld: missing symbol called` from the VMM.

There is no command that reports which prerequisite is missing.

### Proposed solution

Add `brig doctor`. It checks the chain in order and prints one line per check,
with the exact fix when a check fails:

- host OS and architecture, against the supported matrix
- hardware virtualization
- the runtime binary and its version (`hull` or `nerdctl`), plus containerd
- boot assets
- registry reachability, and the image architecture for the selected profile
- signature verification tooling, and the current `BRIG_VERIFY` mode
- profile files: each one parses, and the errors name the file
- secret store availability
- workspace filesystem compatibility
- the `brigd` socket, and its permissions

Add `--json` for the same report, with the field rules from #7. Exit
with code 5 when a required check fails, per #8.

### Additional context

`brig doctor` output is what a bug report needs. The bug template asks the
reporter to assemble the same facts by hand today.

pmoust:
A first run can fail at the host OS, at virtualization, at the runtime binary, at the image pull, or at the secret store, and the error arrives in the words of the layer underneath. `brig doctor` checks the chain in boot order and prints one line per check, with the exact fix under a failure.

```
$ brig doctor
  ok  host      macOS 26.5 on arm64
  ok  virtual   Hypervisor.framework available
  ok  runtime   hull 0.1.0-rc21 at /opt/homebrew/bin/hull
  !!  boot      assets missing at /Users/pmoust/.hull/assets
          run any agent once to fetch them, or set BRIG_BOOT_ASSETS to a directory that has them
  ok  verify    cosign at /opt/homebrew/bin/cosign, BRIG_VERIFY=warn
  ok  profiles  8 built in, 0 in /Users/pmoust/.config/brig
  ok  secrets   keychain reachable
  --  brigd     not running (no socket at /Users/pmoust/.brig/brigd.sock)
  --  image     pass an agent to check its image: brig doctor claude
```

`brig doctor <agent>` adds the image check: the reference resolves against the trust policy and the digest is named. A check whose prerequisite failed says `not reached` rather than failing again in the missing layer's words.

The report is data first and rendered second, the shape `envelope` already uses, so `--json` and the text carry the same facts. `--json` introduces the constants and the field rule #7 will reuse: `apiVersion` is `brig.sh/v1alpha1`, one `kind` per report, fields are added and never renamed or removed inside one version, no credential value anywhere.

**Exit code.** #9 asked for a flat 5. The table that landed in #104 makes 5 mean "verification refused", which a diagnostic that boots nothing never does. Only two checks gate the status, and each returns the error class a real run would: a missing or broken runtime exits 4, a secret store that will not open exits 6. Everything else prints its finding and its fix and leaves the status alone.

**Left out, on purpose.** Image architecture: nothing in the tree reads a manifest platform. Workspace filesystem: nothing exists and it is not clear what it would check. Both are named in #9 and can be filed on their own.

**Moved.** The brigd socket path logic was `package main` in `cmd/brigd`; it now lives in `internal/brigsock`, verbatim, so doctor can ask where the daemon listens. Virtualization is a new probe: `kern.hv_support` on darwin, `/dev/kvm` on linux.

**Second commit.** The fleet executor that produced the first was in disk-wait for four hours and committed without a passing build: three names collided with declarations `cmd/brig` already had. Fixed. It also took `Detect`'s answer on trust, so `BRIG_RUNTIME_BIN=/nonexistent brig doctor` vouched for a binary that was not there. It now looks the binary up and exits 4 on a miss. The underlying gap in `DetectFor` is #139.

Checked on a build of this branch: the output above, `brig doctor codex` resolving the image, `brig doctor nosuch` exiting 3, and the bad-binary case exiting 4. Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`, which has two new cases.

Fixes: #9

### DIFF
--- a/cmd/brig/doctor.go
+++ b/cmd/brig/doctor.go
@@ -1,0 +1,476 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	goruntime "runtime"
+	"strings"
+	"syscall"
+
+	"github.com/brig-sh/brig/internal/brigsock"
+	"github.com/brig-sh/brig/internal/creds"
+	"github.com/brig-sh/brig/internal/profile"
+	"github.com/brig-sh/brig/internal/runtime"
+	"github.com/brig-sh/brig/internal/secret"
+	"github.com/brig-sh/brig/internal/verify"
+	"github.com/brig-sh/brig/internal/wrap"
+)
+
+// brig doctor reports, one line each and in the order a boot hits them, the
+// prerequisites a first run can fail on: the host OS, hardware virtualization,
+// the runtime binary, its boot assets, the signature tooling, the profiles, the
+// secret store, the daemon, and -- when an agent is named -- its image. A first
+// run that dies at any of these surfaces the failure in the words of the layer
+// underneath (a macOS host too old shows up as `dyld: missing symbol` from the
+// VMM); this is the command that says which prerequisite it was, in brig's own
+// words, before anything boots.
+//
+// The report is data first, rendered second: runDoctor returns a slice of
+// checks and prints nothing, so the text form and the --json form say the same
+// facts and a check is testable without capturing stdout. This is the shape
+// (*Config).envelope already uses for the execution envelope; a new check
+// appends to the slice and nothing downstream moves.
+
+// checkState is the two-letter state of one check. Three, and no more: a reader
+// scanning a column wants to find the failures at a glance, and two letters is
+// enough to carry pass, fail and "did not run" without a legend.
+type checkState string
+
+const (
+	// statePass: the prerequisite is there.
+	statePass checkState = "ok"
+	// stateFail: it is not, and the fix is on the next line.
+	stateFail checkState = "!!"
+	// stateInfo is skipped, informational, or not reached because a check it
+	// depended on failed. All three are "nothing to act on here", which is why
+	// they share a mark: a not-reached check must not read as a second failure
+	// in the words of the missing layer, which is the whole thing doctor exists
+	// to stop.
+	stateInfo checkState = "--"
+)
+
+// check is one line of the report as data. It is rendered and serialised, never
+// printed from inside the probe that produced it.
+type check struct {
+	Name    string     `json:"name"`
+	State   checkState `json:"state"`
+	Finding string     `json:"finding"`
+	// Fix is what to type when the check failed, printed indented under the
+	// finding. Empty on a check that did not fail.
+	Fix string `json:"fix,omitempty"`
+	// err is the typed error a failed check contributes to the exit status, or
+	// nil. It is not serialised -- an exit code is not a fact about the host --
+	// and only two checks set it: see the note on doctorExit.
+	err error
+}
+
+// virtualization is the seam a test replaces to force a hostile host: the probe
+// has no environment escape hatch of its own. The other two seams doctor reads
+// already exist for the same reason -- detectRuntime in telemetry.go and
+// openStore in secret.go -- and the BRIG_* variables cover the rest.
+var virtualization = probeVirtualization
+
+// doctorCmd is `brig doctor [<agent>]`, with an optional --json flag of its own.
+func doctorCmd(out io.Writer, args []string) error {
+	jsonOut, agentName, err := parseDoctorArgs(args)
+	if err != nil {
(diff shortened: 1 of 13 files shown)


cmd/brig/doctor.go
```diff
@@ -0,0 +1,476 @@
...
+// them. Everything else -- the host, virtualization, the signature tooling, the
+// profiles, the secret store, the daemon -- stands on its own.
+func runDoctor(agent *profile.Profile, loadErr error) []check {
+	rtCheck := runtimeCheck()
+	runtimeOK := rtCheck.State == statePass
+	return []check{
+		hostCheck(),
+		virtualCheck(),
+		rtCheck,
+		bootCheck(runtimeOK),
```
ananos, line 176: Since #130 landed, there is one more prerequisite worth a row here: whether the runtime's gateway can enforce an egress policy at all.

The flags arrived after hull 0.1.0-rc25, so every released hull today refuses a policied boot with "upgrade the runtime". That is a prerequisite failure a reader would want named before the boot, which is what this command is for.

`runtime.gatewayEnforces(bin)` already answers it -- it asks the binary rather than parsing a version -- so it would want exporting. Adding the check for whether a bound policy would be refused on the resolved backend would cover the other half. Happy to do this as a follow-up if you would rather keep this PR as it is.

ananos (commented):
This is a good addition -- naming the failing prerequisite before a boot fails on it is exactly the right shape, and the `--` informational state keeps the report honest rather than alarming.

One suggestion inline, and a small consistency note: `doctor` parses its own `--json`, and #137 parses its own `--follow`/`--tail`, so neither is covered by `TestBrigFlagsDeclareAPosition` or the overlap check in #132. Not a problem today, just worth deciding whether new verbs join `brigFlags` or stay hand-rolled.

ananos (commented):
One small process note, and it is about a few of the open PRs rather than this one alone -- no need to act on it until merge time.

CONTRIBUTING asks that every commit stands on its own ("Each commit should build and pass tests, so that `git bisect` remains usable"), and that fixup commits from a review round are squashed before merge. Since we rebase-merge, whatever sits on the branch lands on main word for word.

`fix(cli): build doctor against its package, and look up the binary` is the one worth catching here: it says the commit before it did not build, so a bisect that stops there cannot be tested. #137, #138, #134 and #142 have the same shape -- a second commit fixing the first, inside a branch that has never shipped.

Could we fold each of those into the commit that introduced the code before merging? Same content, one semantic commit per change.

I am not throwing stones -- I had exactly this on #130, three commits where two were fixing the first, and folding them at the end took about ten minutes. Happy to help if any of them turn out fiddly to split.

cmd/brig/doctor.go
pmoust: Yes, and as a follow-up, since #140 is based before #130 and the function is not on its base. Filed as #144 with what you wrote here: export `gatewayEnforces`, one diagnostic row, plus whether a bound policy would be refused on the resolved backend.

pmoust:
On the commit-shape note in your review: agreed, and doing it now rather than at merge time. Every fixup from a review round is being folded into the commit that introduced the code, so each commit on each branch builds and passes tests on its own: #133, #134, #137, #140 first, then #138, #141 and #142 once their bases are rewritten, since they stack. Content stays byte for byte; only the commits and their messages change. I will note on each PR when its branch is folded.

pmoust:
Folded to one commit. The message now carries the binary lookup as part of the design, with no mention of an earlier commit that did not build, since that history is gone.

ananos (approved):
Follow-up rather than in-scope is right, since #140 is based before #130 and the function is not on its base. #144 captures it.

One correction to what I wrote in that comment, for whoever picks #144 up: the flags are not in any released hull. rc25 is the newest tag and does not carry them; they sit 17 commits ahead on hull's main. So the doctor row would today report every released hull as unable to enforce -- which is the honest answer, and the reason the row is worth having.

Issue #43, opened 2026-08-25
Let brig env and brig ls work without a runtime on PATH


pmoust:
### Problem

Both commands fail with "no microVM runtime found on PATH" when hull or nerdctl is absent. `brig env`
touches the runtime for one line of its report, the runtime kind and binary. The rest comes from the
profile, the environment and host git config. `brig ls` needs the runtime to list live sandboxes, and can
still say so rather than failing.

This is the wrong failure at the wrong moment. The person most likely to run `brig env` is the one whose
runtime is not working, and it is the command that would tell them what brig intends to do.

### Proposed solution

Report what is knowable without a runtime, and mark the runtime line as unavailable. Exit 0. `brig ls` with
no runtime answers that there is nothing to list.

### Acceptance criteria

- `brig env <profile>` with no runtime on PATH exits 0 and prints every line that does not need one.
- `brig ls` with no runtime exits 0.

### Additional context

#9 is the command that diagnoses a missing runtime. This is about the two commands that should not need
one.

Issue #12, opened 2026-08-25
Add shell completions


pmoust:
### Problem

Session names, profile names and flags are all typed by hand. A wrong profile
name is only reported after brig has loaded the profile directory.

### Proposed solution

Add `brig completion bash|zsh|fish`. Complete the verbs, the profile names
from `profile.Names()`, the session names from the runtime list, and the flags
of the verb that is being typed. Install the zsh completion with the Homebrew
cask.

### Additional context

This depends on #5. Completing session names is only useful once a
session name is what the lifecycle verbs accept.

asapranidis:
**Rescope note, from the #47 planning pass.** Body left as filed.

Two additions, both consequences of #47 rather than changes of intent.

**1. Complete refs, not two separate name lists.** "The profile names from `profile.Names()`, the
session names from the runtime list" becomes: complete `<agent>@<label>`. The `@` is what makes
this tractable — completing `claude` offers the agents, and completing `claude@` offers that
agent's existing labels from the session index. #47 names "what completion offers" as one of the
reasons the separator has to carry the boundary, so this is the issue that proves it.

**2. Completion has to be position-aware.** #47 gives brig's flags three positions with two closed
sets: global flags left of the verb (`--json`, `--isolation`, `-q`, `--verbose`, `--no-color`,
`-y`, `--force`), run-line flags between the verb and the ref (`--home`, `--image`, `--mem`,
`--cpus`, `-d`, `--skills`, `--workspace-mode`), and the agent's own flags to the right of the ref
— which brig must not complete at all, since it does not own them.

"The flags of the verb that is being typed" therefore becomes "the flags legal at the cursor's
position", which is a different and slightly harder job. It is also the payoff: the closed sets
only help a user who can discover them, and completion is how they do.

### Dependencies

Behind #5, as already noted in the body, and behind the ref-and-parser issue to be filed under
#47 — the position table is what completion reads.

PR #141, opened 2026-09-05
feat(cli): add --json to the read verbs

### LINKED ISSUE #7: Add `--json` to ls, env, profiles and secret ls

pmoust:
Every list command printed aligned text, so a script parsed column positions and any change to the text broke it.

`--json` is now a global flag, left of the verb, and the read verbs also take it right after the verb since that is what people type. It works on `ls`, `info`, `agent ls`, `secret ls` and `doctor`. A verb with no JSON form refuses it with exit 2 and names the verbs that have one; #110 lifts that for `run` next.

All five use the envelope #9 introduced in `jsonout.go`: `{apiVersion, kind, data}`. The rule between the two JSON shapes brig now has is written beside the constants: a document verb (`agent show`, `agent export`, `agent new`, `policy show`) prints the bare document because the output is a file; a list or report verb prints the envelope. The four document verbs are unchanged.

Shapes, all through `writeJSONDocument`:

- `SandboxList`: `data.runtime` once, `data.sandboxes` as `{ref, agent, label, sandbox, state, workspace}`. #7's example put `image` per row; it is not in hand for a listing and would need every profile loaded, so `info` carries it instead.
- `Info`: built from the same data `envelope` and `Status` already assemble, through a new data form in `internal/wrap/infojson.go`. Credentials are name and source only, never a value; the test seeds a known value and asserts it never appears. With no runtime on PATH it still exits 0 with `runtime: {available: false}`, keeping #43.
- `AgentList`, `SecretList`: the fields the text output already has. An empty store is `data: []`, not the prose hint.
- `Doctor`: the verb's own `--json` and the global one set the same bool.

The global-position test that asserted `--json` was refused is rewritten to assert it is accepted, with `--nope` and `-y` still refused. The `ls -q` round-trip property is extended to `ls --json`: every ref it prints is accepted by every verb.

One nit left for review: `info`'s `credentials` is `null` rather than `[]` when nothing is forwarded.

Stacked on #140 for the `jsonout.go` constants; GitHub retargets to `main` when that merges.

Checked on a build of this branch: every shape above, both flag positions, `-q --json`, the three refusals, and `info --json` with hull off PATH. Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`, which has one new case.

Fixes: #7

### DIFF
--- a/cmd/brig/main.go
+++ b/cmd/brig/main.go
@@ -19,6 +19,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/brig-sh/brig/internal/creds"
 	"github.com/brig-sh/brig/internal/profile"
@@ -85,6 +86,9 @@ global flags (left of the command, as in: brig -q run claude):
       --json             machine-readable output, for the read verbs: ls, info,
                          agent ls, secret ls and doctor. Also accepted after the
                          verb (brig ls --json). Every other verb refuses it
+      --json (with run)  run the agent as a child and, after it exits, print one
+                         JSON line with its exit status -- so a script can tell
+                         "brig refused" from "the agent failed"
 Within one apiVersion the JSON only ever gains fields, never renames or drops
 one, and no field carries a credential value -- so a script written against it
 keeps parsing, and a machine-readable dump stays a place a secret cannot leak.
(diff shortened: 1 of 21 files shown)


cmd/brig/main.go
```diff
@@ -2351,6 +2555,41 @@ func deprecated(old, replacement string) {
...
+}
+
+// jsonUnsupportedf is the refusal a verb with no JSON form gives to --json: a
+// usage error (exit 2) that names the verbs that do have one, so the reader
+// moves the flag rather than guessing which command it belongs on. #110 lifts
+// this for run, which is why the message is about this release rather than a
+// permanent absence.
+func jsonUnsupportedf(verb string) error {
+	return usagef("`brig %s` has no --json output. --json is for the read verbs: "+
+		"ls, info, agent ls, secret ls, doctor", verb)
```
ananos: `verbTakesGlobalJSON` accepts `env`, but this list does not mention it, so a reader who lands here is told `brig env --json` is unsupported when it works.

I suspect that is deliberate, since `env` is deprecated toward `info` and we would rather not advertise it. If so, could we say so -- "(`env` takes it too, but it is deprecated; prefer `info`)" -- or drop `env` from the predicate and let it fall through to this message? Either is fine, they just should not disagree.

cmd/brig/main.go
```diff
@@ -2351,6 +2555,41 @@ func deprecated(old, replacement string) {
...
+// be a parameter most of them ignore. run sets it every invocation, so a test
+// driving run twice does not inherit the first one's setting.
+var globalJSON bool
+
+// verbTakesGlobalJSON reports whether the verb in the global position has a
+// --json form to ask for. The five read verbs do; agent and secret only in
+// their ls subcommand, which is the one that lists. Everything else -- run, sh,
+// stop, rm, and the verbs that manage rather than list -- has none, so --json
+// left of it is a usage error rather than a flag dropped on the floor.
+func verbTakesGlobalJSON(verb string, rest []string) bool {
```
ananos, line 2812: Minor, and the same note I left on #133: this is a third hand-maintained verb list, after `forwardsTail` and `honorsRunLine`. Each one is reasonable alone. Together they are a little drift surface, and none of them fails when a new verb forgets to join.

No change needed here, just flagging the pattern while several of these are open at once.

ananos (commented):
Good shape overall. Pinning `apiVersion` and promising fields are only ever added within one version is the part that makes this safe to depend on, and it is nice to see it stated in the docs rather than implied.

One small inconsistency inline.

cmd/brig/main.go
pmoust: Deliberate, as you guessed, and now said in the message (6801b5f): `env` is named as taking it and as deprecated, prefer `info`. The predicate is unchanged, so a deprecated spelling keeps working until it is removed, and the two no longer disagree.

cmd/brig/main.go
pmoust: Agreed. Filed as #145: one verb table read by dispatch, the three predicates, `usage` and the completion scripts, with a test that every dispatched verb has a row. #12 needs the same list, which is the moment to build it rather than a fourth copy.

pmoust:
Folded to one commit and rebased onto #140's rewritten tip. Same tree.

ananos (dismissed):
The message names `env` as deprecated now, so the predicate and the text agree while a deprecated spelling keeps working. And #145 is the right home for the verb-table point -- building it when #12 needs the same list, rather than as a fourth copy, is better timing than doing it here.

ananos:
Heads up, and an apology: I closed this one by accident.

I merged #140 with `--delete-branch`, and deleting `feat/9-brig-doctor` -- this PR's base at the time -- made GitHub close this PR two seconds later. I have reopened it and the base is back to `main`. Nothing was lost: `feat/7-json-read-verbs` is intact at `ace3b193` and no commit was dropped.

It does need a rebase before it can go in. Everything merged this morning went in as rebase-merges, so #140's commit is on `main` under a new SHA (`352349c6`) while this branch still carries the old one -- hence the conflict. A `git rebase --onto main <old-140-tip> feat/7-json-read-verbs` should be all it takes, and #142 will want the same afterwards.

The review itself stands approved -- this is only about the SHAs.

ananos (approved):
Re-approving: my earlier approval was dismissed automatically when the branch was rebased, not withdrawn.

Rebased cleanly onto main and it now carries #142's commit too, since that merged into this branch rather than into main -- so this one PR brings both the read-verb `--json` and the run-as-child change, including the `docs/security.md` paragraph scoping the exec handover to everything except `--json`. I checked that paragraph is present here; main does not have it yet, which is correct, it arrives with this.

`go vet` and `go test -race` green on the rebased head.

PR #142, opened 2026-09-05
feat(run): run the agent as a child under --json

pmoust:
`brig run` ends by replacing its own process with the runtime. That is the right default and it stays: the agent owns the terminal, gets signals directly, and its exit status is the command's. It also means brig can never report its own failure in a machine-readable way on a run line, so a script cannot tell "brig refused to start this" from "the agent ran and exited 1".

Under `--json`, and only then, the agent runs as a child. `Attach` joins `Replace` on the runtime interface; both adapters build the child from the same function `Replace` uses, and a test pins that the argv and env are equal. Without `--json` nothing changes, and the first test in the PR is the one that proves `run` still reaches `Replace` and never `Attach`.

**Signals.** The child inherits the tty, so it sits in brig's process group and a Ctrl-C reaches both from the kernel. brig catches SIGINT and waits, forwarding nothing, or the child would get it twice. SIGTERM and SIGHUP reach brig alone and are forwarded. The reasoning is in the comment above the handler because it is the part most likely to be "corrected" wrong later.

**The object.** One compact line on stdout after the agent exits, or instead of running it when brig refuses, so a script's rule is "the last line of stdout is brig's". Kind `Run`, with `stage` (`brig`, `agent`, `detached`, `gui`), `exit`, `error` on a brig refusal, `signal` when the kernel killed the agent, and `ref` and `sandbox` always. brig's exit status keeps its meaning: the agent's own when it ran, the class from the exit table when brig refused. The object is what tells them apart.

`sh` takes the same path on the same mechanism. `-d`, a GUI profile and a shell profile each print their object in place of today's text.

"Always run as a child" was considered and rejected: it changes TTY ownership and signal delivery for every interactive run for the benefit of a flag.

**Second commit.** Found by running the build here: a ref that will not parse is refused inside `parse`, before the JSON context existed, so `brig --json run claude@BAD` printed no line at all. The context is now made before `parse` and refreshed after, in both flag positions.

Stacked on #141 for the global `--json`; retargets to `main` when that merges.

Checked on a build of this branch: an unknown agent, a bad ref in both positions, and a bad `BRIG_RUNTIME_BIN` each print one `Run` line with `stage: brig` and the right exit; `stop` still refuses `--json` naming `run` and `sh`. Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`, which has two new cases driving the stub's exit status through `--json run`.

Fixes: #110

### DIFF
--- a/internal/runtime/attach.go
+++ b/internal/runtime/attach.go
@@ -1,0 +1,114 @@
+package runtime
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"os/signal"
+	"syscall"
+)
+
+// execHandover and attachHandover are the two ways a built command is handed
+// down, kept behind package vars so a test can capture the argv and env each
+// receives without a syscall.Exec that never returns. Replace and Attach share
+// one builder already -- replaceCmd in each adapter -- and stubbing these is how
+// a test proves the two paths cannot drift.
+var (
+	execHandover   = syscall.Exec
+	attachHandover = RunAttached
+)
+
+// RunAttached runs argv as a child of brig, inheriting brig's own stdin, stdout
+// and stderr, waits for it, and returns its exit status. argv[0] is the binary;
+// env is the full child environment. Both adapters build the pair from the very
+// function their Replace uses, so the child and the process replacement cannot
+// drift.
+//
+// It is the child-process counterpart of syscall.Exec. Replace hands the
+// terminal, the signals and the exit status straight to the runtime and never
+// returns, which is the right default: the agent's TUI wants a real tty with
+// nothing between it and the keyboard. Attach keeps brig alive across the exec
+// so brig can report the outcome on a machine-readable line afterwards, which
+// Replace by definition cannot -- there is no brig left once the exec has
+// happened.
+//
+// A child killed by a signal is reported as 128 plus the signal number, the
+// convention a shell follows, so a caller mapping the status onto its own exit
+// returns the same number a shell would for the same death.
+func RunAttached(argv, env []string) (int, error) {
+	cmd := exec.Command(argv[0], argv[1:]...)
+	cmd.Env = env
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+
+	// Registered before Start so a signal that arrives in the gap between fork
+	// and the forwarding goroutine is buffered rather than taking brig down
+	// under the default disposition.
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+	defer signal.Stop(sigs)
+
+	if err := cmd.Start(); err != nil {
+		return 0, err
+	}
+
+	// The child holds the terminal, so it sits in brig's own process group and a
+	// Ctrl-C from the tty is delivered by the kernel to BOTH brig and the child
+	// at the same instant. brig must therefore neither exit on SIGINT nor
+	// forward it: the child already has it, and a second copy -- or brig dying
+	// and orphaning the child mid-keystroke -- is exactly the bug this guards
+	// against. So SIGINT is caught and dropped, and brig goes on waiting.
+	//
+	// SIGTERM and SIGHUP are different in kind: they reach brig alone (a `kill`,
+	// a hung-up session), not through the tty, so the child never sees them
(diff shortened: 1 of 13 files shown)


ananos (commented):
The hard part here is done well. `attachedStatus` handling both the ordinary code and `128+signal`, and `main` reading `agentExit` before it ever prints, means the agent's status still comes back exactly as its own.

Two things below: one documentation promise this quietly changes, and one signal I would rather see spelled out.

internal/runtime/attach.go
```diff
@@ -0,0 +1,106 @@
...
+	//
+	// This is the reasoning most likely to be "corrected" wrongly later: do not
+	// add SIGINT to the forwarded set, and do not make SIGINT return early.
+	done := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case s := <-sigs:
+				switch s {
+				case syscall.SIGTERM, syscall.SIGHUP:
```
ananos, line 84: `os.Interrupt` is registered above but not forwarded here, so brig no longer dies on `^C` and the agent gets it only because the terminal delivers to the whole foreground process group.

That works today and is probably what we want -- brig should wait for the child rather than leave it orphaned. It is the kind of thing that breaks silently though: if the child ever lands in its own process group, or ignores SIGINT, `^C` stops reaching anything and there is no second-`^C` escape.

Could we say that in a comment, and forward SIGINT alongside the other two? Forwarding it to a child already receiving it costs nothing and removes the dependency on the group.

internal/runtime/attach.go
```diff
@@ -0,0 +1,106 @@
...
+func RunAttached(argv, env []string) (int, error) {
+	cmd := exec.Command(argv[0], argv[1:]...)
+	cmd.Env = env
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+
+	// Registered before Start so a signal that arrives in the gap between fork
+	// and the forwarding goroutine is buffered rather than taking brig down
+	// under the default disposition.
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
```
ananos, line 47: This changes a promise `docs/security.md` makes today:

> `brig` hands the tty over with `syscall.Exec` and is gone before the agent produces a byte, which is what buys correct `^C` handling and a truthful exit status

Under `--json` brig is a live parent for the agent's whole life, so that sentence is no longer true on this path. The exit status half you have kept, which is the important one.

Could we update that paragraph in the same PR? CONTRIBUTING asks for it, and it is the page people read before deciding what to trust brig with. Two sentences saying the exec handover holds except under `--json`, and why, should cover it.

internal/runtime/attach.go
pmoust: Kept as is, with the reasoning you asked for written into the comment (9aa6977), including the dependency on the shared process group and that nothing here calls Setpgid. I did not add SIGINT to the forwarded set: the child already receives it from the tty, and a second copy per keystroke is a quit to most TUIs, so forwarding would break Ctrl-C in every `--json` run to guard against a case the code does not create. If the child ever gets its own process group, the comment says the handler must start forwarding it then.

internal/runtime/attach.go
pmoust: Done in 9aa6977. `docs/security.md` now says the exec handover holds except under `--json`, where brig stays as the agent's parent to print one status line after it exits; the tty, the bytes and the exit status are still the agent's.

pmoust:
Folded to one commit and rebased onto #141's rewritten tip. The message now carries the bad-ref line, the process-group dependency and the security.md change as parts of the design.

ananos (approved):
Taking your point on SIGINT -- you are right and I was wrong to ask for it. The child already gets it from the tty, and a second copy per keystroke is a quit to most TUIs, so forwarding would break Ctrl-C in every `--json` run to defend against a case the code does not create. The comment recording the process-group dependency, and saying what must change if the child ever gets its own group, is the durable half of that anyway.

`docs/security.md` verified: the exec handover is stated as holding except under `--json`, with the tty, the bytes and the exit status still the agent's. Thanks for doing that in the same PR.

ananos:
Approved above. Holding the merge until #141 is in: this is still based on `feat/7-json-read-verbs`, and #141 needs a rebase first (see the note there -- I closed it by accident deleting #140's branch, and have reopened it).

Once #141 lands on `main`, this will want the same rebase onto the new SHAs and then it can go straight in.

PR #146, opened 2026-09-05
fix(runtime): check BRIG_RUNTIME_BIN before driving it

### LINKED ISSUE #139: Check BRIG_RUNTIME_BIN the way a profile's runtimeBin is checked

pmoust:
`DetectFor` took `BRIG_RUNTIME_BIN` on trust, where a profile's `runtimeBin` was checked for being on disk and executable and wrapped `ErrBadRuntime` when it was not. The sentinel's own comment said both should reach it; only the profile path did. So a bad value was discovered at the first `ps`, with the runtime's own error and exit 1.

Both paths now share one check. A bare name is looked up on PATH the way `exec.LookPath` does, so `BRIG_RUNTIME_BIN=hull` keeps working; a path that is not there, not a file, or not executable is refused before anything runs, naming the variable, exit 4.

`info` and `ls` still work with no runtime installed: they swallow `ErrNoRuntime`, not `ErrBadRuntime`, and a runtime that was named and is broken is something to fail over.

Checked on a build of this branch: `/nonexistent` and a bare name not on PATH exit 4 naming `BRIG_RUNTIME_BIN`; `BRIG_RUNTIME_BIN=hull` resolves; `info` with hull off PATH still exits 0. Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`, which has one new case.

Fixes: #139

### DIFF
--- a/docs/runtimes.md
+++ b/docs/runtimes.md
@@ -144,7 +144,9 @@ the command line even then.
 
 1. `BRIG_RUNTIME` names the backend, `hull` or `nerdctl`. Anything else is
    refused by name. Unset, it is `hull` on macOS and `nerdctl` everywhere else.
-2. `BRIG_RUNTIME_BIN` is the executable to run, and skips the PATH lookup.
+2. `BRIG_RUNTIME_BIN` is the executable to run. A path is taken as it stands
+   and a bare name is looked up on PATH, and either way one that is missing or
+   not executable is reported against the variable before anything runs.
 3. A profile's `runtimeBin` does the same thing without a variable per shell,
    and loses to `BRIG_RUNTIME_BIN`. A leading `~` is expanded, and a path that
    is missing or not executable is reported against the profile that named it.
--- a/internal/runtime/preference_test.go
+++ b/internal/runtime/preference_test.go
@@ -116,6 +116,95 @@ func TestDetectForSentinelMarksOnlyTheMissingRuntime(t *testing.T) {
 	}
 }
 
+// A path in BRIG_RUNTIME_BIN is checked the way a profile's runtimeBin is.
+// Taken on trust, a value that is not there reached the first exec instead,
+// where it surfaced as a sandbox whose state brig could not read and named the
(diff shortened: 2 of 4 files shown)


ananos (approved):
Nice catch on the asymmetry -- the sentinel's comment said both paths reached it and only one did. `runnable` as one test phrased as the clause after whichever setting named the path is the right shape, and keeping LookPath's semantics means `BRIG_RUNTIME_BIN=hull` still means what everyone assumes it means.

Checked that resolving the bare name to an absolute path does not disturb the gateway records from #130: `ownsGateway` matches on the socket path, not the binary, and `gatewaySpec` carries no binary at all. No interaction.

PR #149, opened 2026-09-05
fix(runtime): refuse a GUI profile on the container path

### LINKED ISSUE #48: Refuse a GUI profile on the container path instead of booting it headless
### Problem

The hull adapter checks whether the chosen backend can show a window, refuses a GUI profile when it
cannot, and names the setting that caused it. The nerdctl adapter never reads the GUI field at all, so
the same profile on the container path issues an ordinary run with no display handling and no message.

The sandbox starts. The window never appears. Nothing says why.

### Proposed solution

Refuse a GUI profile on any path that cannot show a window, with the same message shape the hull adapter
already uses. If a container path can carry a display, implement it. Doing neither is the only option that
should be off the table.

### Acceptance criteria

- A GUI profile on the nerdctl path either shows a window or refuses with a message naming the reason.
- A test covers the refusal.

pmoust:
The hull adapter refuses a GUI profile when its backend cannot show a window and names the setting. The nerdctl adapter never read the GUI field, so the same profile on the container path issued an ordinary run: the sandbox started, the window never appeared, nothing said why.

`nerdctl.Run` now refuses a GUI profile before any command is built, in the same shape hull uses. The reason is the path rather than a setting: the container runtime has no display on either driver, so the message says to run it on macOS, where hull's vz backend can show it. A plain error, as hull's is, so the two paths exit alike.

Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`; the new test asserts the refusal happens before any command runs and that a non-GUI spec is unaffected.

Fixes: #48

### DIFF
--- a/internal/runtime/nerdctl.go
+++ b/internal/runtime/nerdctl.go
@@ -178,15 +178,33 @@ func (n *nerdctl) isDocker() bool { return filepath.Base(n.bin) == "docker" }
 // nerdctl has a flag for. GenericBoot is not ignored -- urunc reads the same
 // two annotations from the container's OCI spec that hull takes on its command
 // line, so the Linux path is a pass-through.
-// CanRun refuses a policy this runtime cannot enforce, rather than dropping it
-// in silence.
+// CanRun refuses what this runtime cannot honour, rather than dropping it in
+// silence. It guards both paths, which is the whole point of living here rather
+// than in Run: a sandbox that is already up is joined rather than booted, so Run
+// is never called on the second `brig run`, and a check that only guarded Run
+// would refuse a profile the first time and wave it through every time after.
+// That is the failure RunChecker describes; hull refuses in supports, and its
+// CanRun is supports, so both paths are covered the same way here. Run calls
+// CanRun first, so the direct path is unchanged.
 //
-// Nothing here reads spec.Egress: this runtime hands the sandbox to the
-// container network, which brig does not filter. A boot that carried on would
-// print a POLICY row and an isolated NETWORK row over a sandbox with
-// unrestricted egress, which is the exact outcome the refusals on the other
-// backend exist to prevent. See RunChecker for why it is not inside Run.
+// A GUI profile wants a graphical window, and there is nowhere on this path to
+// put one: the container runtime has no display on either driver, so the refusal
+// is about the path rather than a setting to flip. hull refuses the same profile
+// when its backend cannot show a window; refuse it here too, before anything is
+// built, so the two paths exit alike instead of booting headless with the window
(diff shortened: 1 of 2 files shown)


internal/runtime/nerdctl.go
```diff
@@ -196,6 +196,17 @@ func (n *nerdctl) CanRun(spec RunSpec) error {
 }
 
 func (n *nerdctl) Run(spec RunSpec) error {
+	// A GUI profile wants a graphical window, and there is nowhere on this path
```
ananos: The refusal is inside `Run`, and this is the one path where that is not enough.

A sandbox that is already up is joined rather than booted, so `Run` is never called on the second `brig run` -- which means a GUI profile is refused the first time and waved through every time after, with the window silently missing exactly as it is today. It is the failure mode the `RunChecker` doc comment describes a few files over, in almost these words.

hull does not have this problem: its GUI refusal lives in `supports`, and `CanRun` *is* `supports`, so it is reachable from both paths. The PR body says this refuses "in the same shape hull uses" -- moving it into `CanRun` is what would make that true.

Cheap to do: `backendSpec` already fills in `GUI: c.Profile.IsGUI()`, so `checkBackend` is passing the field through and nothing new needs plumbing. Moving the block into `nerdctl.CanRun` above -- alongside the policy refusal, which is there for the same reason -- covers the join path for free, and `Run` still calls `CanRun` first so the direct path is unchanged.

The two new tests should keep passing as they are, since they go through `Run`. Worth one more asserting `CanRun` refuses it, to pin the property that the join path is covered.

ananos (commented):
Right problem, and the message is good -- naming the path rather than a setting to flip is the honest framing, since there is nothing to toggle here.

One placement point inline, which I think is worth fixing before this goes in.

internal/runtime/nerdctl.go
pmoust: You were right, and it was the join path exactly as you described. Moved into `nerdctl.CanRun` beside the policy refusal in 2f19598, message unchanged, with the doc comment saying why it lives there. `Run` still calls `CanRun` first so the direct path is identical, and the two existing tests pass unchanged. Added one asserting `CanRun` itself refuses, which is the property that pins the join path.

ananos (approved):
Exactly right, and better than what I asked for: the refusal sits ahead of the policy one in `CanRun`, the doc comment explains why it lives there rather than in `Run`, and `Run` still calls `CanRun` first so the direct path is untouched. Verified the new test pins `CanRun` itself, which is the property that covers the join path.

`go vet` and `go test -race` green here.

PR #151, opened 2026-09-05
docs: add AI policy

pmoust:
Adds `AI_POLICY.md` to brig.

It states what AI-assisted work is welcome here, that disclosure is optional,
and that whoever submits a change stays responsible for it. The same text goes
into every repository, with this repository's name in place.

`CONTRIBUTING.md` links to it. Its rule requiring AI use to be disclosed is
removed, along with the "LLM usage" section of the pull request template. The
policy replaces both.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -53,11 +53,3 @@ change. A negative test is worth more here than a positive one: "the denied
 variable was not forwarded" and "the planted symlink was refused" are the
 assertions that catch a regression.
 -->
-
-## LLM usage
-
-<!--
-For transparency: note any AI assistance used in this change, e.g.
-"Authored with assistance from Claude Code; I have reviewed every line and am
-accountable for the change."
--->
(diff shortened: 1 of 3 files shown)


ananos:
Read it through -- it is a good document, and the closing summary earns its place. The bar it sets on evidence ("do not claim that tests passed... unless that actually happened") and on maintainer attention is the part worth keeping strict.

Not merging this one with the others: it removes the disclosure requirement from CONTRIBUTING and the LLM section from the PR template, which is a policy call for @ananos rather than something to carry in on a code review. Flagged for a decision. No objection from me on the text itself.

ananos (approved):
@ananos has confirmed the policy call, so this is good to go.

The document itself reads well, and the two parts that will do the most work are the evidence bar ("do not claim that tests passed, a bug was reproduced, a benchmark improved, or a behavior was verified unless that actually happened") and the section on maintainer attention. Dropping the disclosure requirement is coherent with those: what is asked for is verification, not a label.

Issue #45, opened 2026-08-25
Publish the cask from the release workflow, and add a release checklist


pmoust:
### Problem

The release config skips publishing a cask for a prerelease, and every tag so far is a prerelease, so the
tap works only because the casks are maintained by hand. Both carry a header saying they were written by
hand against a specific release, with the checksums written in.

They already drift: brig's cask and hull's cask are pinned to different releases. Meanwhile the tap's README
still tells visitors that the advertised install path does not exist and points them at a private tap.

Other public surfaces a release leaves behind: two GitHub releases carry the same tag, one draft and one
prerelease; pkg.go.dev serves a much older version than the repository; and the bug report template asks
reporters to run `brig status`, which is not a command.

### Proposed solution

Publish the cask from the release workflow for prereleases too, or cut a stable tag so the existing config
applies. Add a release checklist covering the tap README, the issue template, one release per tag, and the
documentation a release changes. Fix the bug report template to name a command that exists, which is
`brig env` today and `brig inspect` after #10.

### Acceptance criteria

- A tagged release updates the cask with no hand edit.
- One release exists per tag.
- The tap README describes the install path that works.
- The bug report template names commands that exist.

pmoust:
Partly landed. The bug template no longer asks for \`brig status\`: \`.github/ISSUE_TEMPLATE/bug_report.md:22\` now says \`brig info <agent>\`.

Still open, and what the title describes: \`.goreleaser.yaml:194\` has \`skip_upload: auto\` under \`homebrew_casks\`, and the release workflow marks every tag so far as a prerelease, so the cask has never actually been published.

PR #152, opened 2026-09-05
docs: add a release checklist, and reuse an existing draft on re-release

pmoust:
Two GitHub releases exist for `v0.1.0-rc16`, a Draft and a Pre-release with the same name: the workflow opened its draft, and a second release was created beside it instead of publishing the draft. Nothing in the repo said how a release is meant to be cut.

Two commits:

- `ci`: `release.use_existing_draft: true` in `.goreleaser.yaml`, so a re-run of the workflow for a tag that already has a draft targets that draft rather than opening a second release. A goreleaser v2.5 option; the workflow pins `~> v2`. `draft: true` stays, for the reason its comment gives. `goreleaser check` validates the result.
- `docs`: `docs/releasing.md`, the checklist in order: bump `VERSION` and why the tag must match it, tag and push, publish the one draft and never a second, the tap and why the cask ships only on a stable tag, the proxy request that gets pkg.go.dev to index the version, the docs a release touches, and a final check. Linked from `CONTRIBUTING.md` under a new Releasing heading.

Not in this PR, per #45's other points: publishing the cask on a prerelease, which needs a decision about hull's cask first; the tap's own README, which is another repository. The bug template already names `brig info`.

Refs: #45

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -153,6 +153,13 @@ release:
   # now, which makes the review the only thing standing between a tag and
   # something the world can download.
   draft: true
+  # Re-run the workflow for a tag that already has a draft -- a
+  # workflow_dispatch retry, or a second push -- and target that draft
+  # instead of opening a second release beside it. Without this, tag
+  # v0.1.0-rc16 ended up with two releases of the same name, a Draft and a
+  # Pre-release; the fix is to publish the draft, not create another. (GitHub
+  # only, goreleaser >= v2.5; the workflow pins `~> v2`.)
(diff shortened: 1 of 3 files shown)


ananos (approved):
Good to have the checklist written down, and reusing the draft is the right fix for the rc16 duplicate -- publish the one draft, never create a second.

Verified `use_existing_draft` needs goreleaser >= v2.5 and the workflow pins `~> v2`, so the option resolves; the release-config check passing is `goreleaser check` accepting it rather than just YAML being well-formed.

PR #153, opened 2026-09-05
feat(secret): add a Secret Service backend for Linux

### LINKED ISSUE #16: Add a secret store for Linux
### Problem

`brig secret` works on macOS only. On every other platform the store refuses
to open, with "brig secret needs the macOS keychain so far". The README says
the same in its command table, on both secret rows.

A Linux user therefore keeps credentials in environment variables, which is
the path `brig secret import` was built to replace.

### Proposed solution

Implement the store for Linux against the Secret Service API, which is what
GNOME Keyring and KWallet both speak. Keep the `Store` interface as it is. The
non-darwin file already says a second backend is an addition at that seam.

For a host with no keyring, add one documented fallback that holds no
plaintext at rest: a command binding that brig runs to fetch the value, in the
shape `brig secret import <profile> <name> --from-command`, which already
exists.

Say in the error which of the two is missing, and how to install it.

### Alternatives considered

- An encrypted file under `~/.brig`, unlocked by a passphrase. It works
  everywhere and it makes brig the owner of a key, which the current design
  avoids on purpose.
- Leave Linux to environment variables. It is what happens now, and it means
  the platform where brig has the container runtime story has the weaker
  credential story.

None. The store is on the host, and what reaches the guest is unchanged.

pmoust:
`brig secret` worked on macOS only, so a Linux user kept credentials in environment variables, the path `brig secret import` exists to replace, on the platform where brig has the container runtime story.

Linux now stores into a Secret Service keyring on the session bus, which is what `gnome-keyring` and KWallet both speak. One item per secret in the default collection, tagged with a `service` attribute of brig's own so `List` searches by attribute and touches nothing else in the keyring. Provenance goes in item attributes with the encoding `provenance.go` already defines; `Modified` comes from the item. The collection is unlocked through the keyring's own prompt when locked. `Kind()` is `secret-service`. The `Store` interface is unchanged and `Annotator` is implemented; `keychain_other.go` keeps every other non-darwin platform.

When there is no session bus, or no service answers on it, `open()` says which of the two is missing and both ways out: install a keyring and log in to a session that starts it, or bind the secret to a command with `brig secret import <profile> --from-command`, which holds no plaintext at rest. No file fallback, on purpose.

New dependency: `github.com/godbus/dbus/v5`, and nothing else.

Two commits: the backend, then the two README rows and the two `docs/secrets.md` passages that said "macOS only".

**Testing.** Unit tests cover the attribute mapping both ways, name validation and the two error shapes. The integration test runs create, read, update, delete, list, provenance round-trip and `ErrExists` under a test-only `service` value, and skips unless `DBUS_SESSION_BUS_ADDRESS` names a bus with a service on it. Neither the fleet executor that wrote this nor this Mac has one, so **the backend has not yet been exercised against a live keyring**; a reviewer on a Linux desktop running `go test ./internal/secret/` would close that gap. Builds and vets on darwin and linux; `go vet ./...`, `go test -race ./...` and `./script/smoke.sh` pass here.

Fixes: #16

### DIFF
--- a/internal/secret/secretservice_linux.go
+++ b/internal/secret/secretservice_linux.go
@@ -53,13 +53,10 @@ const (
 	propModified   = itemIface + ".Modified"
 )
 
-// defaultAlias is the collection brig stores into: whichever one the keyring
-// has aliased as the user's default (gnome-keyring calls it "login"). The alias
-// path is where ReadAlias points when a default is set.
-const (
-	defaultAlias     = "default"
-	defaultAliasPath = dbus.ObjectPath("/org/freedesktop/secrets/aliases/default")
-)
+// defaultAlias is the alias brig reads to find the collection it stores into:
+// whichever one the keyring has aliased as the user's default (gnome-keyring
+// calls it "login").
+const defaultAlias = "default"
 
 // noObject is the object path the Secret Service returns to mean "none": no
 // prompt is needed, or no default collection exists. It is a bare "/".
@@ -91,9 +88,10 @@ type dbusSecret struct {
 
 // secretService is the Store backed by the freedesktop Secret Service.
 type secretService struct {
-	service string
-	conn    *dbus.Conn
-	session dbus.ObjectPath
+	service    string
+	conn       *dbus.Conn
+	session    dbus.ObjectPath
+	collection dbus.ObjectPath
 }
 
 // Caught at build time rather than wherever a secretService first gets assigned
@@ -131,6 +129,18 @@ var (
 		}
 		return slices.Contains(names, secretsName)
 	}
+
+	// readDefaultAlias resolves the "default" alias to the collection brig
+	// stores into, held as a variable for the same reason as the two above: the
+	// executor has no bus, so a test forces the answer -- in particular the
+	// noObject a keyring with no default collection returns.
+	readDefaultAlias = func(s *secretService) (dbus.ObjectPath, error) {
+		var path dbus.ObjectPath
+		if err := s.service0().Call(methodReadAlias, 0, defaultAlias).Store(&path); err != nil {
+			return noObject, fmt.Errorf("finding the default keyring collection: %w", err)
+		}
+		return path, nil
+	}
 )
 
 func open() (Store, error) { return openService(service) }
@@ -148,6 +158,9 @@ func openService(service string) (*secretService, error) {
 	}
 	s := &secretService{service: service, conn: conn}
 	if err := s.openSession(); err != nil {
+		return nil, err
+	}
+	if err := s.resolveCollection(); err != nil {
 		return nil, err
 	}
 	return s, nil
@@ -171,6 +184,22 @@ func errNoService() error {
 		"Install a keyring (gnome-keyring or KWallet, both speak the Secret Service API) and log in "+
 		"to a session that starts it, or bind the secret to a command instead with "+
 		"`brig secret import <profile> --from-command '<sh>'`, which holds no plaintext at rest",
+		ErrUnsupported)
+}
+
+// errNoDefaultCollection is the third shape of the same refusal: a keyring is
+// running and answering, but ReadAlias("default") resolves to noObject, which
+// is the service saying plainly that no default collection exists yet -- not
+// that the default sits at some well-known path. Nothing has created it: a
+// headless or freshly provisioned session never unlocked a login keyring, so
+// brig has nowhere to store into. Both ways out again, in the voice of the two
+// above: an unlock from a desktop session is what creates the default, or bind
+// the secret to a command that holds no plaintext at rest.
+func errNoDefaultCollection() error {
+	return fmt.Errorf("%w: a keyring is running but has no default collection to store into. "+
+		"Unlock your keyring once from a desktop session, which is what creates it, or bind the "+
+		"secret to a command instead with `brig secret import <profile> --from-command '<sh>'`, "+
+		"which holds no plaintext at rest",
 		ErrUnsupported)
 }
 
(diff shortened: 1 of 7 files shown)


internal/secret/secretservice_linux.go
```diff
@@ -0,0 +1,531 @@
...
+// defaultCollection is the path of the collection brig stores into: the alias
+// the keyring points at, or the well-known alias path when ReadAlias reports
+// none, which is the path the keyring will have created the default under.
+func (s *secretService) defaultCollection() (dbus.ObjectPath, error) {
+	var path dbus.ObjectPath
+	if err := s.service0().Call(methodReadAlias, 0, defaultAlias).Store(&path); err != nil {
+		return noObject, fmt.Errorf("finding the default keyring collection: %w", err)
+	}
+	if path == noObject {
+		return defaultAliasPath, nil
```
ananos: This fallback is where the headless case lands, and the assumption in the comment above -- "the path the keyring will have created the default under" -- is the one that does not hold.

What I measured: `ReadAlias("default")` returns `/`, meaning there is genuinely no default collection. brig then returns `defaultAliasPath`, the keyring resolves that alias to `/collection/login`, and since nothing ever created it, `CreateItem` comes back with the raw D-Bus string `Object does not exist at path /org/freedesktop/secrets/collection/login`.

So a `ReadAlias` of `/` is not "the default is at the well-known path", it is "there is no default collection" -- the service telling us the answer plainly. Falling back turns a knowable state into the least helpful message in the package, which is a shame next to `errNoBus` and `errNoService`, both of which are excellent.

This is reachable outside my container: a keyring installed on a headless or freshly-provisioned Linux box has no login collection until something creates it, and Linux is precisely where brig's container path lives, so a server is a normal place to be running this.

Suggestion: treat `path == noObject` as its own refusal, in the shape of the other two -- a keyring is running but has no default collection, unlock it once from a desktop session to create it, or use `--from-command`, which holds no plaintext at rest. Creating the collection ourselves via `CreateCollection` is the other option, though I would lean against brig silently creating a keyring the user did not ask for.

Worth a test too: it is a plain path, since it needs only `ReadAlias` returning `/`, which is what a bus with a keyring and no default collection gives you for free.

ananos (commented):
This is careful work. Searching by a `service` attribute so brig never reaches outside its own namespace, and being explicit in the comment that the namespace buys containment of brig rather than provenance of what it finds, is the right honesty about what a label is worth. The create/update split with the reasoning about the whole-attribute-set match is exactly the trap I would have expected someone to fall into, and `search` merging the locked array with the unlocked one -- so a locked keyring does not read as empty -- is the other one.

You flagged that the backend had never met a live keyring, so I ran it against one. Recipe, in case it is useful for CI:

```
docker run --rm -v $PWD:/src -w /src golang:1.25 bash -c '
  apt-get update -qq && apt-get install -y -qq gnome-keyring dbus-x11
  dbus-run-session -- bash -c "
    echo -n \"\" | gnome-keyring-daemon --unlock --components=secrets
    go test ./internal/secret/ -v -run Integration"'
```

The Secret Service does come up and answer (`NameHasOwner org.freedesktop.secrets` returns true). Results:

- All unit tests pass, including both `open()` error shapes.
- `TestIntegrationReadAndDeleteReportAbsence` and `TestIntegrationUpdateNeverCreates` **pass** against the real service -- so search, the attribute filter, `ErrNotFound` and update-never-creates are all confirmed on live D-Bus.
- Every test that **writes** fails: `Object does not exist at path /org/freedesktop/secrets/collection/login`.

Before you go looking for the bug: I do not think it is yours. `secret-tool store` -- libsecret's own client -- fails identically in the same session, so brig is behaving exactly as the reference implementation does. The container has no default collection, because creating one needs the gcr prompter and it cannot open a display.

That said, it did surface a real gap, inline.

internal/secret/secretservice_linux.go
pmoust: Thank you for running it against a real bus; that is the gap I flagged in the PR body and could not close here. Fixed in c3b0151: `ReadAlias` returning `/` is now its own refusal in the voice of `errNoBus` and `errNoService` -- a keyring is running but has no default collection, unlock it once from a desktop session, or bind the secret with `--from-command`. It surfaces from `open()` with the other two, so it lands before a value is read from stdin rather than as a `CreateItem` failure. `defaultAliasPath` is gone, there is a unit test driving `ReadAlias` to `/`, and the error table in `docs/secrets.md` has the row. I did not add `CreateCollection`, for your reason: brig should not silently make a keyring the user did not ask for.

internal/secret/secretservice_linux_test.go
```diff
@@ -0,0 +1,332 @@
...
+		t.Skip("no session bus: set DBUS_SESSION_BUS_ADDRESS and run a Secret Service keyring to exercise the backend")
+	}
+	var b [8]byte
+	if _, err := rand.Read(b[:]); err != nil {
+		t.Fatal(err)
+	}
+	s, err := openService(testService + hex.EncodeToString(b[:]))
+	if err != nil {
+		if errors.Is(err, ErrUnsupported) {
+			t.Skipf("no Secret Service on the session bus: %v", err)
```
ananos, line 188: This skip line now says something that is not true in the case it most often fires on.

`ErrUnsupported` covers three causes, and for the new one a Secret Service *is* on the session bus and answering -- it is the default collection that does not exist. That is exactly what my container hits, so the run prints "no Secret Service on the session bus" six times over an error whose own text correctly says a keyring is running.

The error already carries the whole story, so `t.Skipf("%v", err)` would read better, or `t.Skipf("no usable keyring: %v", err)` if you want a prefix. Test output only, no behaviour attached.

ananos (approved):
Closed properly, and I re-ran the same container to check rather than reading it: `gnome-keyring` under `dbus-run-session`, the identical setup that produced the raw D-Bus string last time.

Before: four integration tests **failed** with `Object does not exist at path /org/freedesktop/secrets/collection/login`.

Now: they **skip**, carrying your message -- "a keyring is running but has no default collection to store into. Unlock your keyring once from a desktop session, which is what creates it, or bind the secret to a command instead with `brig secret import <profile> --from-command`". `TestResolveCollectionReportsNoDefaultCollection` passes, and the package is green.

Two things I am glad you did rather than the obvious alternatives: raising it from `open()` via `resolveCollection`, so `brig secret create` refuses before it reads the value from stdin instead of failing halfway through a `CreateItem`; and not adding `CreateCollection`. `readDefaultAlias` as a seam is what makes the unit test possible without a bus at all.

One cosmetic thing left over, below. Not worth holding the PR for -- approving.

PR #154, opened 2026-09-06
feat(brigd): version the protocol, return stable codes, and check the peer

### LINKED ISSUE #21: Version the brigd protocol, and harden the socket
### Problem

`brigd` speaks line-delimited JSON, one request per line. A request is
`{"op":"ensure","agent":"claude-code","name":"refactor"}` and a failure is
`{"ok":false,"error":"..."}`. The protocol has no version field, no request id
and no error code, so a client matches on error text, and a client written
against 0.1 cannot tell it is talking to 0.3.

The socket is created 0600 and carries lifecycle control over sandboxes
holding live credentials. The mode is right. Nothing above it checks who is on
the other end.

The protocol is documented in public, so people will write against it whether
or not it is called internal.

### Proposed solution

1. Add a version field to every request and response, and refuse a version
   brigd does not know.
2. Add an optional request id, echoed in the response, so a client can
   pipeline.
3. Add a stable `code` next to `error`, from the set in #8.
4. Check the peer credential on accept, and refuse a connection from another
   uid.
5. Write down what happens after a restart: which sessions are reconciled from
   the runtime, and which are dropped.
6. State in the brigd doc whether this is a supported interface or an internal
   one. Either answer is fine. The absence of an answer is not.

### Alternatives considered

Mark the protocol internal and remove it from the docs. It costs nothing to
maintain, and it removes the only way to drive brig from a long-lived process.

pmoust:
`brigd` spoke line-delimited JSON with no version field, no request id and no error code, so a client matched on error text and one written against 0.1 could not tell it was talking to 0.3. The socket is 0600 and carries lifecycle control over sandboxes holding live credentials; nothing above the mode checked who was on the other end.

Four commits, each building and passing tests on its own:

1. `internal/exitcode`, a shared classifier producing the six codes in README's exit table from the same internal error types `cmd/brig/exit.go` reads. Its package comment records that `exit.go` is the other producer and should move onto this package once the CLI branches in flight have landed; that move is not in this PR.
2. Protocol `v`, `id` and `code`. A request with no `v` is read as version 1, since no older client exists; an unknown `v` is refused with the usage code and names the versions brigd speaks. `id` is echoed unchanged when given. `code` sits beside `error` so a script driving brigd branches the way one driving brig does.
3. The peer check: the connection's uid is read (`SO_PEERCRED` on Linux, `LOCAL_PEERCRED` on darwin, one file per platform) and a uid other than brigd's own gets one well-formed refusal line and a close.
4. `docs/brigd.md`: the fields, examples carrying `v`, `id` and `code`, what a restart reconciles, and the statement the issue asked for: the protocol is internal until 0.3, a field may be added inside `v: 1` but none renamed or removed, and a breaking change is `v: 2` — the same rule `--json` output follows.

New dependency `golang.org/x/sys`, direct. It forces `go 1.25` to `go 1.25.0` in `go.mod`: `x/sys v0.47.0` declares `go 1.25.0` itself, and Go refuses to build when the main module's directive is lower. Note for whoever merges second: #153 also touches `go.mod` for `x/sys`, so one of the two will need a `go mod tidy` after the other lands.

Tested with `go vet ./...`, `go test -race ./...` and `./script/smoke.sh`; each of the four commits was checked out and built and vetted on its own; cross-builds for linux.

Fixes: #21


internal/exitcode/exitcode.go
```diff
@@ -0,0 +1,87 @@
...
+// general failure, 2 a usage error, 3 no such profile or sandbox, 4 a runtime
+// that is missing or broken, 5 a boot verification refused, 6 a credential that
+// could not be resolved. A caller that only checked "zero or not" keeps working;
+// one that wants to branch on the reason now can.
+//
+// cmd/brig/exit.go is the other producer of these numbers -- it classifies the
+// same internal error types for the CLI. It is package main and cannot be
+// imported, and an open PR is editing it, so it is left where it is; once the
+// branches touching it land it is to be moved onto this package so there is one
+// classifier rather than two that can drift. Do not do that move here.
```
ananos, line 14: This paragraph is honest about the duplication and I have no argument with deferring the move. It does leave one thing untrue elsewhere, though.

`cmd/brig/exit.go` currently opens with:

> the mapping below is **the only thing that produces them**, so the table and the behaviour cannot drift

After this PR that is no longer the case -- there are two producers, which is precisely what this comment acknowledges. So the file that promises drift is impossible is the one file that stops being right. Could we amend that sentence in this PR, pointing at `internal/exitcode` and at the planned move? One line, and it keeps the pair honest until they are merged.

I compared the two classifiers and they agree today: same order, same classes, same numbers, most-specific-first. Worth naming the concrete hazard for whoever does the move, though -- the error types are *not* shared. `usageError`/`notFoundError` in the CLI and `UsageError`/`NotFoundError` here are distinct types, so `Of` would not classify the CLI's and `exitCode` would not classify these. They agree now only because each is used in its own binary. Unifying the types is the actual work of that follow-up, not moving the function.

Worth an issue so it does not rest solely on a comment -- happy to file it if you would rather keep this PR moving.

ananos (commented):
Strong PR, and the peer check is the part I went at hardest. Reading the uid from the socket rather than from anything the client says is the right call, both platform files take it from the kernel, and doing it once at accept -- with the reasoning about why it is not in `handle` -- is correct.

What I checked, since an authorization change deserves more than a green tick:

- `go d.accept(conn)` is genuinely on the `ln.Accept()` path, so the check is not dead code that a passing unit test would hide.
- `startDaemon` runs the real `serve()`, and `ask()` dials a real unix socket, so **every socket test in the file goes through `peerUID` for real** on both platforms. If `LOCAL_PEERCRED` were wrong on darwin, the whole suite would fail with "could not read the peer's identity" rather than passing quietly. That is much better positive coverage than it looks like at first glance.
- The negative half is `peerAllowed(own+1, own)`, a pure function rather than a second uid on the socket. That is the honest limit and I think it is the right trade -- a real cross-uid test needs root -- but it is worth being clear that what is proven is the decision and the refusal's shape, not the wiring for the deny path. The wiring is proven for allow, which is most of the value.

Also ran it here: `go vet` clean, `go test -race ./cmd/brigd/... ./internal/exitcode/...` green, the full suite green, and `GOOS=linux` cross-builds. Four commits that each stand on their own, which is exactly the shape we asked for on #140.

One thing to fix before merge, inline.

ananos:
Heads up on the ordering you called in the PR body: #153 has just landed, so this is the one that needs the `go mod tidy`. It brought `github.com/godbus/dbus/v5` in, and this PR adds `golang.org/x/sys` and moves the directive to `go 1.25.0`.

The review finding above still stands as the one thing to fix before merge -- `cmd/brig/exit.go`'s comment claiming its mapping is the only producer of those codes.

internal/exitcode/exitcode.go
pmoust: Good catch, and the sentence you quote is the one that stops being true. Fixing it in this PR.

The hazard you name is the real one for whoever does the move: `usageError` and `notFoundError` are `cmd/brig` types, so `internal/exitcode` cannot see them and classifies its own request errors instead. That is why the two agree today by construction rather than by sharing code, and why the move needs those types lifted first rather than the classifier simply deleted.

pmoust:
Rebased onto current `main` (`4ef1832`), four commits, each checked out and built and vetted on its own. The `exit.go` header fix from review is folded into the commit that made that header untrue, the `exitcode` one, rather than sitting as a fifth commit.

One thing to know when CI runs: `TestCompletePositions` fails on this branch exactly as it fails on `main` today. That is the #141/#158 collision, fixed by #160 and unrelated to this change; every other test and the smoke script pass. Once #160 lands this goes green with no further push.

internal/exitcode/exitcode.go
pmoust: Done, in the `exitcode` commit itself (23ca0f3) rather than as a follow-up, since that is the commit that makes the old sentence untrue. The header now says there are two producers, that they agree by construction rather than by sharing code, and that collapsing them means lifting `usageError` and `notFoundError` somewhere both can see.

ananos (approved):
The comment fix is right, and folding it into 23ca0f3 -- the commit that makes the old sentence untrue -- is the right home for it. It now says there are two producers, that they agree by construction rather than by sharing code, and that collapsing them means lifting `usageError` and `notFoundError` somewhere both can see. The added rule that a change here belongs in `internal/exitcode` too is better than what I asked for.

Re-verified the peer check on this head: `go vet` clean, `cmd/brigd` green, linux cross-build fine.

One thing before it can go in: it is one commit behind main and `TestCompletePositions` fails on this head -- not your change. It is the table-derived fix `6ed914b`, which landed after your base. I rebased it locally onto current main to be sure: rebases clean, and the test passes. So it needs the branch updating and nothing more.

PR #155, opened 2026-09-06
fix: restore the work two folded commits reverted

pmoust:
**`main` is missing merged work and needs this before anything else lands.** The commits for #130, #131, #132 and #135 are still in the history; their content is not in the tree.

Gone from `main` today:

| | |
| --- | --- |
| #130 | the per-sandbox gateway and egress enforcement: `Egress`, `isolatedSocket`, `gatewayEnforces`, `ensureIsolatedGateway`, and their tests. `internal/runtime/gateway.go` is 168 lines where it should be 546 |
| #131 | a verbless bad ref exits 1 again instead of 2 |
| #132 | the flag-overlap test knows one agent again instead of eight |
| #135 | `brig agent ls extra` silently drops the argument again |

**Cause, and it is mine.** I rewrote two branches with `git reset --soft main` while their working trees were still the ones they had been written against, several merges earlier. A commit made that way carries the old tree, so its diff against the new parent reverts everything the branch had not seen, and a rebase-merge replays exactly that diff. The two commits are `1fb3036` and `7da1279`. I checked each fold by comparing its tree to the branch's own previous tip, which is the one check that cannot catch this.

Nothing was lost permanently: every commit is intact, only `main`'s tree regressed.

**How this was rebuilt.** From `817f3e6`, the last unaffected commit, replaying in order the intended changes of the two commits at fault and every commit merged after them, all nine cleanly. The result is recorded here as one change against `main`, and its tree is byte-identical to that rebuild. The two fixes keep their behaviour: `runOnly` still lives on the `brigFlags` row, `Reserved`/`ReservedFor` are still two functions. No new work is in this diff.

**Verification.** A marker from each of the four restored changes is present again; `gofmt`, `go vet ./...`, `go test ./...` and `./script/smoke.sh` pass, plus a linux cross-build. Two tests failed on a first run and passed alone and in a second full run: `TestHullPinsDigestAsksTheBinary` and `TestARequestThatWouldPromptIsRefusedRatherThanAsked`, both timeout-bounded, both squeezed by my own parallel suites on this machine.

**Worth knowing for review order:** #146 and #140 merged while `main` was missing #130, so their CI never saw the two together. The suite is green here with them combined, but that is a combination CI has not run until this PR.

Refs: #130
Refs: #131
Refs: #132
Refs: #135


pmoust:
The guard that would have stopped this is now #156, targeting `main` on its own: it compares the test names at the merge base with the names at the head and fails on any that vanished. Run against the two commits at fault it reports 78 and 12 removed tests and exits 1.

It flags this PR too, for three tests, and correctly: #130 and #132 renamed them, and all three successors are present here.

| flagged | renamed to |
| --- | --- |
| `TestBrigFlagsOverlapWithClaudeCodeOnlyWhereKnown` | `TestBrigFlagsOverlapWithShippedAgentsOnlyWhereKnown` |
| `TestAttachSaysThePolicyIsNotEnforcedYet` | `TestAttachSaysWhereThePolicyIsEnforced` |
| `TestCheckSaysThePolicyIsNotEnforcedYet` | `TestCheckSaysWhereThePolicyIsEnforced` |

I have labelled this PR `removes-tests`, which is the escape hatch #156 documents.

ananos (approved):
Verified this independently before approving. The rebuilt branch lands on tree `aaa90bc9`, byte-identical to the version I checked before the force-push, so the content is the same and the history now reads properly -- a revert that names the mistake, then the real fix.

On my side: no test missing against `817f3e6`, `gateway.go` back to 546 lines, `go vet ./...` clean, `go test ./...` green on 11 of 12 packages. `internal/secret` fails on the keychain (`User interaction is not allowed`), which needs a GUI session and is unrelated to this change.

Good to merge. Let's take #156 right after, so the guard is armed for everything that follows.

ananos:
One thing to pick up after this lands: the force-push orphaned the stack. #137, #141 and #154 are still children of the old tip `33eebff`, so GitHub shows them swallowing the whole restoration (+4965, +5802 and +4450 respectively). #149 is fine, it was never parented here.

Nothing is wrong with those branches. They just need replaying onto the restored main:

```
git rebase --onto origin/main 33eebff feat/25-brig-logs        # #137
git rebase --onto origin/main 33eebff feat/7-json-read-verbs   # #141
git rebase --onto origin/main 33eebff feat/21-brigd-protocol   # #154
```

All three are clean here, no conflicts, and the diffs drop back to their real size: +1133/-7, +1974/-75 and +618/-22.

@pmoust worth keeping `33eebff` around locally until you are through those three. It is unreferenced now, so gc will take it eventually.

pmoust:
Agreed

Issue #32, opened 2026-08-25
Correct the claim that ~/.claude is memory-only


pmoust:
### Problem

The README and `docs/secrets.md` both say `~/.claude` is memory-only so that no credential reaches host
disk. Seven paths under it are declared as volumes in the `claude-code` profile and are hostmounted into
the workspace, so they live on host disk: the settings file, the instructions file, sessions, projects,
plugins, skills and the history file.

The credential file itself is on a memory-backed mount and does not reach disk. The directory named in the
claim is partly host disk by design, and the profile's own comment states the correct rule: anything under
`.claude` that is not named in `volumes` is ephemeral.

The narrower claim is the strong one and is worth keeping. The sentence as written is not accurate, and it
is the kind of sentence a careful reader checks first.

### Proposed solution

Replace the sentence with the scoped one: the credential file is written to a memory-backed mount and never
reaches host disk. Then say what does reach disk, and why: transcripts, settings and history are hostmounted
into the workspace on purpose, so a session's contents persist even though its credential does not.

### Acceptance criteria

- No published sentence claims `~/.claude` as a whole is memory-only.
- `docs/security.md` names the hostmounted paths.
- The claims suite in #22 gains a test that the credential file is on a memory-backed mount and that the
  hostmounted paths are on disk, so the two cannot drift apart again.

PR #157, opened 2026-09-06
docs: add release, CI, coverage, Go and license badges

pmoust:
The README opened with the logo and went straight into prose. A reader had to leave the page to learn whether `main` is green, which release is current, or what Go version the module needs.

Five badges under the lockup: latest release, CI on `main`, coverage, Go version, license.

Two details worth the review:

- The release badge passes `include_prereleases`. Every tag so far is a prerelease, so the default query reports no releases at all. With it, the badge reads `v0.1.0-rc17` today.
- The Go badge reads `go.mod` through shields rather than hardcoding a number, so it cannot go stale when the directive moves. It reads `v1.25` today.

A coverage badge needs a number, so this also measures coverage in CI and uploads it. The lane runs **on a push to `main` only**: the figure describes the branch other work builds on, and measuring it per pull request would spend a second full `go test` run to move a badge that only `main` publishes. One run writes the profile; the step summary is rendered from that same profile rather than from a second run. The upload is `fail_ci_if_error: false`, because a coverage service that cannot be reached is not a reason to call a green commit broken.

Local coverage today is 81.1% of statements. I verified all four external badge URLs return 200 and render the values above; the codecov one renders `unknown`, as expected until the first upload.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,24 @@ jobs:
       - name: go vet
         run: go vet ./...
 
+      # The profile costs almost nothing to write on a run that happens anyway,
+      # so the coverage number comes from the same run that gates the merge
+      # rather than from a second suite that could disagree with it. atomic is
+      # what -race requires, and what a parallel package run needs regardless.
       - name: go test
-        run: go test -race ./...
+        run: go test -race -covermode=atomic -coverprofile=coverage.out ./...
+
+      # Only the upload is gated: the badge follows main, and a pull request
+      # pays nothing for a step it skips. Never fails the run, since a coverage
+      # service that cannot be reached is not a reason to call a green commit
+      # broken.
+      - name: Upload coverage to Codecov
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        uses: codecov/codecov-action@0fb7174895f61a3b6b78fc075e0cd60383518dac # v5.5.5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage.out
+          fail_ci_if_error: false
 
       # The macOS path is the one that ships first, and it is also the one no
       # Linux runner can execute. Compiling it is the part CI can do, and it
--- a/README.md
+++ b/README.md
@@ -5,6 +5,14 @@
   </picture>
 </p>
 
+<p align="center">
+  <a href="https://github.com/brig-sh/brig/releases/latest"><img alt="Release" src="https://img.shields.io/github/v/release/brig-sh/brig?include_prereleases"></a>
+  <a href="https://github.com/brig-sh/brig/actions/workflows/ci.yml"><img alt="CI" src="https://github.com/brig-sh/brig/actions/workflows/ci.yml/badge.svg"></a>
+  <a href="https://codecov.io/gh/brig-sh/brig"><img alt="Coverage" src="https://codecov.io/gh/brig-sh/brig/graph/badge.svg"></a>
+  <a href="go.mod"><img alt="Go" src="https://img.shields.io/github/go-mod/go-version/brig-sh/brig"></a>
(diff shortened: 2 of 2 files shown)


ananos:
Useful, and the two details you flagged are the right two. `include_prereleases` is genuinely load-bearing -- I checked, every brig release so far is a prerelease, so without it the badge reads "no releases". Reading the Go version out of `go.mod` through shields rather than hardcoding it is the other good call.

Confirmed the rest as described: the badge points at `v0.1.0-rc17`, which is current; the codecov action is pinned by SHA; `fail_ci_if_error: false` with the reasoning is right, and so is measuring on `main` only rather than spending a second full suite per PR.

Incidentally the release list still shows the `v0.1.0-rc16` Draft sitting beside the `v0.1.0-rc16` Pre-release -- the duplicate #152 just fixed. Nice to see the bug that motivated it still visible.

One thing before this goes in: the branch carries a merge commit, `Merge branch 'main' into docs/readme-badges`. We rebase-merge here, so it will be flattened on the way in and no history harm is done, but it is worth rebasing rather than merging while a branch is in review -- it keeps the diff readable and matches what CONTRIBUTING asks for. Same note that went round on #140 the other day, no more than that.

Happy to approve once it is rebased onto the current `main`.

ananos:
One more, now that hull's equivalent (brig-sh/hull#32, merged) is next to it: hull gets the same badge without a second suite.

It adds `-covermode=atomic -coverprofile=coverage.out` to the unit-test step that already runs, then gates only the *upload* on `push` + `refs/heads/main`. This PR instead adds a whole `coverage` job that runs `go test ./...` again on every push to main.

The reasoning in the comment here -- not spending a second run per pull request -- is right, and hull's shape honours it more cheaply: the profile costs almost nothing to write on a run that happens anyway, so there is no second suite on any event at all. It also means the coverage number comes from the same run that gates the merge, rather than a separate one that could in principle disagree.

Worth folding in while this is being rebased. Not a blocker if you would rather keep the job separate -- there is a fair argument that an isolated job is easier to reason about.

pmoust:
yeah - fixing

pmoust:
Both done, and the branch is now one commit on current `main` (`a94e4b6`).

**The merge commit is gone.** It came from a UI "Update branch"; rebased instead, which also picked up #156, so the `tests kept` job is on this branch's base now.

**Coverage folded into the existing run.** The separate `coverage` job is deleted. `go test -race ./...` becomes `go test -race -covermode=atomic -coverprofile=coverage.out ./...`, and only the upload is gated on `push` + `refs/heads/main`. Your reasoning carried it: the profile costs almost nothing on a run that happens anyway, there is now no second suite on any event, and the number comes from the same run that gates the merge rather than from one that could disagree with it. `atomic` is what `-race` requires regardless.

On the `v0.1.0-rc16` duplicate still sitting in the release list: that is #152's fix working forward, not backward. `use_existing_draft` stops the next tag from doing it again; the existing pair needs someone to delete the stale draft by hand. Happy to do that if you want it tidy before the next release.

ananos (approved):
Both done and verified: one commit on current main with no merge commit, and the separate `coverage` job is gone -- the profile now comes off the `go test -race` run that already gates the merge, with only the upload gated on a push to main. That is the hull shape, and it is the better one.

`go vet`, `go test -race` and the linux cross-build all green here.

Fair point on the rc16 duplicate too: #152 stops a second release being created from here on, it does not tidy the one already there. That wants deleting by hand whenever someone is next in the releases page.

ananos:
Approved, but the merge is refused: "the head branch is not up to date with the base branch". Main moved to `6ed914b` after you rebased, so this needs one more update before it can go in. Nothing wrong with the branch itself.

Worth knowing that rule is on, incidentally -- it is the thing that would have caught the red main earlier today, when #141 and #158 were each green against a base that did not have the other.

PR #158, opened 2026-09-07
feat(cli): add shell completion for bash, zsh and fish

### LINKED ISSUE #12: Add shell completions

asapranidis:
## Summary

Refs, agent names and flags are all typed by hand, and a name that is wrong is
only reported once brig has resolved it. That cost falls hardest on the closed
flag sets: three positions with two closed vocabularies only help a reader who
can discover them, and completion is how they discover them.

`brig completion bash|zsh|fish` prints a script to stdout and installs nothing --
where a script belongs differs per shell and per host, and a command that edits
a startup file has a blast radius no other verb here has. The docs carry the
line per shell, and the Homebrew cask now installs all three itself.

The scripts carry none of brig's vocabulary. Each one collects the words left of
the cursor, hands them to a hidden `brig __complete`, and renders what comes
back: a directive line, then the candidates. So the verbs, the refs and every
position rule are declared once, in Go, beside the table that already decides
them, and the three shells cannot come to disagree about what brig accepts
because none of them knows.

Completion stops exactly where brig's own parsing stops. Global flags left of
the verb; brig's own flags on **either** side of the ref, because `split` reads
them wherever they stand on the run line; and then nothing, from the first word
or flag brig does not own -- `--`, an unrecognised flag, or a bare word past
run's project slot, which is itself offered as directories. What brig does not
own it does not complete, and that boundary is walked once, in the order `split`
walks it, so the engine and the parser cannot disagree about where it is.

Refs complete as one grammar, `claude` and `claude@refactor` together: the agents
from the profile registry, the labels from the session index. The index is a file
read rather than a question put to the runtime, because detecting a runtime and
listing its sandboxes costs hundreds of milliseconds on a keystroke and answers
nothing at all on a host that has none. A sandbox removed outside brig therefore
lingers as a candidate until the next `brig ls` prunes it -- a stale candidate is
a name that fails when run, not a wrong action.

## Related issues

Closes #12
Refs #47 (the position table this reads is what makes "the flags legal at the
cursor" answerable)

## Changes

- `brig completion bash|zsh|fish` prints an embedded script; nothing is installed.
- `brig __complete` is the engine: candidates on stdout, nothing on stderr, exit
  0 always, and dispatched before the profile-load notices for that reason -- a
  warning printed on a TAB press lands in the middle of the line being typed.
- Four directives are the wire format. `:names` carries matched candidates;
  `:dirs` and `:files` hand the slot to the shell, which is the one thing a
  candidate list cannot say; `:none` means brig owns nothing here, and is
  distinct from a failure so a shell does not fall back to filenames where brig
  asked for silence.
- `walkRunLine` decides the boundary the way `split` does, including the `=` that
  bash breaks out of `--flag=value` (it is consumed with its value, so it is not
  read as the ref or shifted into an operand slot) and `rm --all`, which names no
  session and refuses every argument, so it completes nothing further.
- Run-line flags come from `brigFlags` directly, so a flag added there is
  completed with no further change. The noun groups build their arguments with a
  `flag.FlagSet` inside the function that runs them, so there is no table to read
  them off; `groups` is that table for completion only, and a subcommand missing
  from it completes nothing.
- Only what the help text teaches is offered. The retired spellings brig still
  answers to (`exec`, `shell`, `env`, `create`, `reset`, `profiles`, `policies`,
  `--name`, `--workspace`, `agent list`) are absent. `--memory` needed a new
  `undocumented` field on `brigFlags`: it is not retiring, but the table could
  not otherwise say "still works, do not teach it".
- `stop` and `rm` offer the sessions there are rather than every agent there
  could be. `run`, `sh` and `info` take an agent that has never run, so they
  offer all of them. `agent edit` and `agent rm` offer only agents with a file of
  their own, which is what both commands require.
- `wrap.SessionRefs` reads the session index's keys. The index was already keyed
  by ref, so nothing new is recorded.
- The archive carries the three scripts verbatim and `homebrew_casks.completions`
  installs them, so a cask install has completion without a second step.

## Checklist

Testing beyond the suite: **all three scripts have now had their own shell run
over them.** bash through a real completion call on `/bin/bash` 3.2.57 (no
`compopt`, which is why every `compopt` call is guarded) under `set -u`; zsh
through real TAB presses in a `zpty` pseudo-terminal; fish through `complete -C`
on fish 4.9.2. All three answer as the table expects, `:dirs` and `:files`
included.

`script/smoke.sh` now carries a completion section for that, since these are
shell bugs the Go tests cannot reach. It prefers `/bin/bash` over whatever `bash`
resolves to, because on macOS that is 3.2 -- the version where the `set -u` bug
lives -- while `bash` on PATH is often a Homebrew 5.x that cannot reproduce it.
Both new bash cases were confirmed to fail against the previous scripts.

`cmd/brig/completion_test.go` is a 43-case position table plus tests for the wire
format, for the engine's silence when a profile will not parse, for the dispatch
path including the `--` guard, and `TestGroupsTableCoversEveryGroupFlag`, which
reads the noun groups' own flag registrations through `go/ast` and requires each
to appear in the completion table.

Neither promise changes, and one of them shaped a decision worth flagging:
**secret names are deliberately not completed.** Listing them means opening the
store -- on macOS that is `security dump-keychain`, and a secret-service backend
can raise an unlock prompt -- and a keystroke must do neither. `brig secret <TAB>`
completes the subcommands; the operand stays empty. `TestCompletePositions`
asserts that (`secret names are not offered`), so a later change that reaches for
the store on a keystroke fails the test rather than surprising someone at a
prompt.

What the engine reads on a keystroke: the profile registry, the session index,
and the policy directory. What it does not: the runtime, the keyring, or any
credential. The one new export, `wrap.SessionRefs`, returns session names, which
`brig ls` already prints.

---
Written with AI assistance (Claude Code). I have reviewed every line and am
accountable for it.

### DIFF
--- a/cmd/brig/completion.go
+++ b/cmd/brig/completion.go
@@ -1,0 +1,694 @@
+package main
+
+import (
+	"embed"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+
+	"github.com/brig-sh/brig/internal/policy"
+	"github.com/brig-sh/brig/internal/profile"
+	"github.com/brig-sh/brig/internal/wrap"
+)
+
+// Shell completion, in two commands.
+//
+// `brig completion <shell>` writes a script to stdout and installs nothing.
+// Where that script belongs differs per shell and per host, and a command that
+// edits a startup file has a blast radius nothing else here has; the docs say
+// the line to add.
+//
+// The scripts carry none of brig's vocabulary. Each one collects the words left
+// of the cursor, hands them to `brig __complete`, and renders what comes back.
+// So the verbs, the flags and where each flag is legal are declared once, in
+// Go, beside the table that already decides them -- and the three shells cannot
+// come to disagree about what brig accepts, because none of them knows.
+//
+// __complete answers on stdout and says nothing anywhere else. It is an ABI
+// between a script and the binary that generated it, so it is in no help text.
+
+//go:embed completions/brig.bash completions/brig.zsh completions/brig.fish
+var completionScripts embed.FS
+
+// completeVerb is the hidden engine's name. Two underscores, because it is not
+// a word anyone types and it must not read like one.
+const completeVerb = "__complete"
+
+const completionUsage = `brig completion -- print a shell completion script
+
+usage:
+  brig completion bash    print the bash script
+  brig completion zsh     print the zsh script
+  brig completion fish    print the fish script
+
+Nothing is installed for you. Write the script where your shell reads
+completions from, or source it from your startup file:
+
+  bash    brig completion bash >/usr/local/etc/bash_completion.d/brig
+          (or: eval "$(brig completion bash)" in ~/.bashrc)
+  zsh     brig completion zsh >"${fpath[1]}/_brig"
+          then start a new shell, or run: compinit
+  fish    brig completion fish >~/.config/fish/completions/brig.fish
+
+Completion offers the verbs, the refs ` + "`brig ls`" + ` prints, and the flags that are
+legal where the cursor is. Right of the ref the arguments are the agent's, so
+brig completes nothing there -- with one exception: on run the first word after
+the ref is the project directory brig mounts, and directories are offered for
+it.
+`
+
+// completionCmd prints the script for one shell.
+func completionCmd(out io.Writer, args []string) error {
+	if len(args) == 0 {
+		return usagef("completion needs a shell: bash, zsh or fish")
+	}
+	switch args[0] {
+	case "-h", "--help", "help":
+		_, err := io.WriteString(out, completionUsage)
+		return err
+	}
+	if len(args) > 1 {
+		return usagef("unexpected argument %q; `brig completion` takes one shell", args[1])
+	}
+	shell := args[0]
+	switch shell {
+	case "bash", "zsh", "fish":
+	default:
+		return usagef("no completion script for %q; brig has bash, zsh and fish", shell)
+	}
+	script, err := completionScripts.ReadFile("completions/brig." + shell)
+	if err != nil {
+		return err
+	}
+	_, err = out.Write(script)
+	return err
+}
+
+// The directive is the first line of an answer, and it says what the lines
+// under it are -- or, where a candidate list cannot say it, what the shell
+// should do instead.
+const (
+	// dirNames: candidates follow, already matched against the current word.
+	dirNames = ":names"
+	// dirDirs: complete directories. No candidates follow.
+	dirDirs = ":dirs"
+	// dirFiles: complete paths. No candidates follow.
+	dirFiles = ":files"
+	// dirNone: brig owns nothing at this position, so offer nothing. It has to
+	// be distinguishable from a failure: a shell that reads "brig had no
+	// answer" falls back to filenames, and the tail right of the ref is the one
+	// place brig has to stay silent rather than guess.
+	dirNone = ":none"
+)
+
+// completeCmd answers one completion request. It writes candidates to out and
+// nothing to stderr, ever, and returns no error: a notice or a failure printed
+// on a keystroke lands in the middle of the line the reader is typing.
+func completeCmd(out io.Writer, words []string) {
+	// The profile directory is read here rather than by the dispatcher, so that
+	// a file that will not parse costs a candidate rather than a line of
+	// diagnostics across the prompt. What loaded is completed; what did not is
+	// not, and the next real command says why.
+	_ = profile.Load(profile.Dir())
+	// Nothing below prints a notice, and this is the belt to that braces: a
+	// warning added anywhere the engine reaches stays silent on a keystroke.
+	verbosity = wrap.Quiet
+
+	directive, candidates := complete(words)
+	fmt.Fprintln(out, directive)
+	for _, c := range candidates {
+		fmt.Fprintln(out, c)
+	}
+}
+
+// verbs are the commands completion offers: the taught vocabulary, and only it.
+//
+// The retired spellings brig still answers to -- exec, shell, env, create,
+// reset, profiles, policies -- are deliberately absent. They keep working and
+// they say what replaced them; completing them would teach a word that is
+// leaving, to the one reader who has not typed it yet.
+var verbs = []string{
+	"agent",
+	"completion",
+	"doctor",
+	"info",
+	"ls",
+	"policy",
+	"rm",
+	"run",
+	"secret",
+	"sh",
+	"stop",
+	"telemetry",
+	"version",
+}
+
+// refVerbs are the verbs whose operand is a session ref.
+var refVerbs = map[string]bool{
+	"run":  true,
+	"sh":   true,
+	"stop": true,
+	"rm":   true,
+	"info": true,
+}
+
+// complete is the whole decision: which words are already on the line, and what
+// may stand where the cursor is.
+//
+// words is the line with `brig` dropped, and the word under the cursor is the
+// last of them -- empty when the cursor sits on fresh whitespace. That the
+// current word is always present, even empty, is what lets one function serve
+// both "finish this token" and "what can come next".
+func complete(words []string) (string, []string) {
+	if len(words) == 0 {
+		words = []string{""}
+	}
+	cur := words[len(words)-1]
+	before := words[:len(words)-1]
+
+	// The verb, and what stands between it and the cursor. Everything left of
+	// the verb is a global flag or a mistake, and both are skipped: a token brig
+	// would refuse is not a verb, so reading on is what keeps the rest of the
+	// line completable.
+	verb := ""
+	var rest []string
+	for i := 0; i < len(before); i++ {
+		a := before[i]
+		if strings.HasPrefix(a, "-") {
+			if mine, takesValue := ours(a, posGlobal); mine && takesValue && !strings.Contains(a, "=") {
+				i++
+			}
+			continue
+		}
+		verb, rest = a, before[i+1:]
+		break
+	}
+
+	if verb == "" {
+		if strings.HasPrefix(cur, "-") {
+			return names(cur, flagSpellings(posGlobal, ""))
+		}
+		// Only the verbs. `brig claude@refactor` with no verb works, and is in
+		// no help text on purpose; completing it would teach the second
+		// spelling that the taught line already covers.
+		return names(cur, verbs)
+	}
+
+	switch {
+	case verb == "completion":
+		if len(rest) > 0 {
+			return dirNone, nil
+		}
+		return names(cur, []string{"bash", "fish", "zsh"})
+	case verb == "ls":
+		// One flag, no operand.
+		if strings.HasPrefix(cur, "-") {
+			return names(cur, []string{"--quiet", "-q"})
+		}
+		return dirNone, nil
+	case verb == "doctor":
+		if strings.HasPrefix(cur, "-") {
+			return names(cur, []string{"--json"})
+		}
+		if len(bareWords(rest, posRun)) > 0 {
+			return dirNone, nil
+		}
+		return names(cur, agentNames())
+	case groups[verb] != nil:
+		return completeGroup(verb, rest, cur)
+	case refVerbs[verb]:
+		return completeRunLine(verb, rest, cur)
+	}
+	return dirNone, nil
+}
+
+// completeRunLine answers between a lifecycle verb and the end of the line.
+//
+// The three positions a brig line has are the whole of the logic here: brig's
+// own flags up to the ref, the ref, and then the agent's own vocabulary, which
+// brig does not complete because it does not own it.
+func completeRunLine(verb string, rest []string, cur string) (string, []string) {
+	// `--` is the reader saying the rest is the agent's. Nothing after it is
+	// brig's to complete, whatever it looks like.
+	for _, a := range rest {
+		if a == "--" {
+			return dirNone, nil
+		}
+	}
+
+	// The value of one of brig's own flags, when that is what the cursor is on.
+	if kind, ok := pendingValue(rest, cur); ok {
+		return operandCandidates(kind, cur)
+	}
+
+	bare := bareWords(rest, posRun)
+	refGiven := len(bare) > 0
+
+	if strings.HasPrefix(cur, "-") {
+		if refGiven {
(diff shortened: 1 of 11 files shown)


panosmaurikos (approved):

ananos (commented):
thanks @asapranidis, seems great! The shape is right -- one `__complete` verb, scripts that know none of brig's vocabulary -- and the silence-on-stderr contract is a nice touch. I ran the engine against the PR head and hit a few real cases we should fix before merging:

- **bash splits `--flag=value` on `=`**, so the engine sees three tokens and only consumes the `=`. For instance:
  - `brig run --network=<TAB>` completes nothing (`complete(["run","--network","="])` returns `:none`)
  - `brig run --mem=4096 <TAB>` returns `:dirs` -- the value counts as the ref, so refs are never offered
  - `brig policy attach --name=x <TAB>` offers *agent* names where the policy name belongs (the operand slot shifts by one)

  The walkers should consume two tokens when the next one is a bare `=`, and normalize `cur="="` to empty (as `_init_completion` does).
- **brig.bash breaks under `set -u`** on bash <= 4.3 (macOS `/bin/bash` is 3.2.57): `brig <TAB>` prints `words[@]: unbound variable` mid-line. We should guard the empty expansion with `${words[@]+"${words[@]}"}`.
- **the fish cache never invalidates**: `__brig_out` is keyed on the commandline text only, so a session started after the first TAB is never offered for the life of the shell. I would drop the memoization -- bash and zsh re-ask on every keystroke and are fine.
- **`brig rm --all <TAB>` offers session refs** that removeAll then refuses. Should be `:none` once `--all` is on the line.
- **`brig secret create gh -f <TAB>` completes nothing** -- the groups table misses the `-f/--file/--stdin` flags that nameAndFile parses. It would be nice to add a small test diffing each `sub` entry against the corresponding FlagSet, so the table cannot drift silently.
- **right of the ref the two halves disagree**: `brig run claude --m<TAB>` is silent for `--mem` (a flag split() does read there), while `brig sh claude npm --network <TAB>` completes postures into the *agent's* argv. pendingValue should stop at the agent boundary, same as the flag-name branch does.

One question: the new `completion` verb now shadows a user profile named `completion` (verb dispatch precedes the ref fallback). Presumably fine, but maybe worth a hint in the usage error?

All of the above reproduced on 42ba406 with a throwaway probe test -- happy to share the vectors. None of this changes the design; it is finetuning the walkers.

asapranidis:
All six reproduced on 42ba406. Fixed in 7deb6c3 (engine), 19edd28 (scripts), 875137c (docs). Table cases added for each. For the two shell bugs I ran the new tests against the old scripts first to confirm they fail.

### 1. `=` splitting

bash splits `--mem=4096` into `--mem`, `=`, `4096`. Both walkers consumed one token after a value-taking flag, leaving `4096` in the stream. It has no `-` prefix, so it was counted as a positional; the first positional on a run line is the ref. Same bug in the group walker shifted every operand index by one.

Fix: consume two tokens when the token after a value-taking flag is `=`. Separately, `cur == "="` is normalised to `""` so the value completes from an empty prefix.

```
brig run --network=<TAB>           :none    → isolated offline shared
brig run --mem=4096 <TAB>          :dirs    → refs
brig policy attach --name=x <TAB>  agents   → allow-gh no-net
```

### 2. Flags right of the ref

`split()` appends to `mine` in its `default:` branch regardless of `passed`, so brig reads its own flags on both sides of the ref. Completion returned `:none` for any flag after the ref, so `--mem` was not offered on a line brig parses.

The condition I had was wrong, not incomplete: I gated on `refGiven`, but the correct gate is where `split()` stops reading — at `--`, at a flag not in `brigFlags`, or at the second positional (first on non-`run` verbs).

Replaced `bareWords()` with `walkRunLine()`, which returns `refGiven`, `projectTaken` and `tailBegun` from one pass in `split()`'s order. `bareWords` returned a count, which cannot distinguish "ref named" from "tail begun". `pendingValue` had the same gap: it matched on the flag alone, ignoring position, so `--network` after `npm` completed brig's postures inside the agent's argv.

```
brig run claude --m<TAB>            :none     → --mem
brig sh claude npm --network <TAB>  postures  → :none
brig run claude ./p npm --m<TAB>              → :none
```

### 3. `rm --all`

`--all` is stripped by `takeAll` before the run line, and `removeAll` rejects any remaining argument — `TestLsAndRemoveAllRefuseArguments` covers that. Completion offered refs. Now `:none` when `--all` is present, and `--all` is offered only while `refGiven` is false.

### 4. `secret create` / `update`

`nameAndFile` registers `-f`, `--file`, `--stdin`; the table had none of them. Added; `-f` returns `:files`.

### 5. bash under `set -u`

`"${words[@]}"` on an empty array is an unbound variable before bash 4.4. Now `${words[@]+"${words[@]}"}`; `cur` reads with a `-` default.

Note on the test: `bash` on my PATH is Homebrew 5.3, which does not reproduce this, so my first version of the smoke case passed against the broken script. `script/smoke.sh` now uses `/bin/bash` (3.2.57 on macOS) and says why. Linux CI is 5.x, so there the case only checks the script runs clean under `set -u`.

### 6. fish cache

`__brig_out` was keyed on `commandline -cp`. The answer also depends on the session index and the profile and policy directories, none of which the line reflects, and there is no cheap way to detect a change in them. Memoization removed; 3 execs per TAB instead of 1.

Two TABs on an unchanged line in one fish process, with a session created between them:

```
old:  claude-code  →  claude-code
new:  claude-code  →  claude-code claude-code@later
```

### Drift test

Not the version you described. Diffing against the real `FlagSet` requires seven parsers (`nameAndFile`, `nameAndYes`, `parseExportLine`, `parseNameAndForce`, `showPolicy`, `parseAttachArgs`/`parseCheckArgs`, `parseImport`) to return theirs, which is a refactor across `secret.go`, `policy.go` and `main.go`. I left that out of a PR that is already under review.

`TestGroupsTableCoversEveryGroupFlag` parses the package with `go/ast`, collects every `*Var(ptr, "name", ...)` registration, and requires each name to appear in the `groups` table. Two limits, both stated in the test:

- One direction, and it does not check which subcommand a flag belongs to. The parsers are shared — `nameAndYes` serves `secret delete` and `agent rm`.
- It reads a name only from a string literal. The run line registers through `func(n string) { fs.StringVar(..., n, ...) }`, so those are invisible; `elsewhere` names `parseGlobal` as the one excluded parser and fails if it stops registering anything.

Reverting the `secret create` entry makes it fail:

```
nameAndFile registers the flag "--file", and the groups table does not offer it.
```

Tell me if you want the `FlagSet` extraction in this PR rather than a follow-up.

### Shadowed profile name

Verb dispatch precedes the ref fallback for every verb, not just this one, so an agent named `ls` is already unreachable as `brig ls`. That is intentional — otherwise the command set depends on which agents are installed. I added the hint to the no-shell error only:

```
brig: completion needs a shell: bash, zsh or fish. The agent of that name is
reached as `brig run completion`
```

### fish is now verified

I had flagged it as unverified in the description. Installed fish 4.9.2 and drove the script with `complete -C`; `script/smoke.sh` covers it when the host has fish and skips otherwise. Description updated — it also still stated the old "nothing right of the ref" rule.

ananos (commented):
Re-checked all six on 875137c, on the real binary rather than by reading. Every one of them is genuinely closed:

- **set -u on bash 3.2** -- built the branch and drove `_brig_completion` under `set -u` in `/bin/bash` 3.2.57. Clean on the empty-array first-word path that used to blow up: 13 verbs at the root, prefix filtering works, profiles after `run`. The `${words[@]+"${words[@]}"}` guard with the comment naming macOS bash 3.2 is exactly right.
- **bash `=`-split** -- `(brig run --network = is)` now returns `isolated`, and the empty-value form returns all three postures. That was the finding and it is fixed.
- **fish memoization** -- `__brig_line`/`__brig_out` are gone and the comment explains why it re-asks. Good call writing down the reason rather than just deleting the cache.
- **`rm --all <TAB>`** -- `:none`, and bare `rm <TAB>` still offers refs.
- **`secret create -<TAB>`** -- `--file`, `--stdin`, `-f` all present.
- **after-ref asymmetry** -- `run claude --<TAB>` offers flag names now.

The three commits are semantic too, not "address review", which I appreciate.

One adjacent thing the fix surfaced rather than caused, inline. Not a regression -- my original report was the bash shape and that is closed -- but it is the same feature not working in the other two shells, so it seemed worth catching now.

cmd/brig/completion.go
```diff
@@ -0,0 +1,775 @@
...
+var sessionVerbs = map[string]bool{
+	"stop": true,
+	"rm":   true,
+}
+
+func refsFor(verb string) []string {
+	if sessionVerbs[verb] {
+		return wrap.SessionRefs()
+	}
+	return refCandidates()
```
ananos, line 396: This bail is what makes inline `--flag=value` complete in bash but nowhere else.

The doc comment above says the `=` is stepped over "so `--network=<cursor>` completes the postures the spaced spelling does". That holds only for a shell that splits on `=`. bash does, because `=` is in `COMP_WORDBREAKS`; zsh and fish do not, so the whole thing arrives as one token and `strings.Contains(last, "=")` returns "" here.

Measured, not inferred:

- zsh: `${(z)'brig run --network=is'}` is **3** words, the last being `--network=is`. Feeding that through the script's own argv shape gives `brig __complete -- run --network=is` -> `:none`. Spaced form -> `isolated offline shared`.
- fish: `commandline -ct` is likewise the single token `--network=is` -> `:none`. Same control passes.
- bash: 5 words, and it works.

So `brig run --network=iso<TAB>` completes nothing in zsh, which is macOS's default shell and so most brig users' shell.

The fix looks small and you already have the piece: `strings.Cut(arg, "=")` is used at line 639 for the spelling. Rather than returning "" when `last` contains `=`, cut it -- left side resolved through `ours`, right side kept as the partial value to filter on -- and the single-token shape lands in the same place the split one already does. The candidate filtering would want the value prefix, which is the one new bit.

Happy for this to be a follow-up issue instead if you would rather not grow the PR; it is a gap in new behaviour rather than a break in existing behaviour, so nothing regresses by shipping as is.

asapranidis: Fixed in 9b28ba5 rather than deferred — zsh being the macOS default makes this most users' shell, and it turned out to be one branch.

Confirmed your measurement first:

```
zsh:   ${(z)"brig run --network=is"}  → 3 words, last = --network=is
fish:  commandline -ct                → --network=is
bash:  "=" in COMP_WORDBREAKS         → yes, so it splits
```

Real zsh TAB on `brig run --network=is` did nothing; `complete -C` in fish returned nothing.

One correction to the diagnosis. The `strings.Contains(last, "=")` bail is not the path this takes, and it is right as it stands: it reads `rest`, where a token containing `=` means the flag already has its value and the next word is the ref. `complete(["run","--network=offline",""])` must return refs, and does. There is now a case pinning that.

The inline token never reaches it, because it is the *current word*, not part of `rest`. `complete(["run","--network=is"])` sets `cur="--network=is"`, `rest=[]`, so `walkRunLine` sees an empty line and `cur` falls through to the flag-name branch, where nothing matches. So the fix went one position earlier, in `completeRunLine` before that branch, and in `completeGroup` for the same shape.

Your suggested `strings.Cut` is what it uses, and the value prefix is handled the way you predicted:

```
run --network=is           :none  → --network=isolated
run --network=             :none  → --network=isolated --network=offline --network=shared
agent new x --from=cla     :none  → --from=claude --from=claude-code --from=claude-desktop
run --network = is                → isolated              (bash's shape, unchanged)
```

Candidates carry the flag because the shell replaces the whole current word — a bare `isolated` would turn `--network=is` into `isolated`. Verified by driving real zsh (`zpty`) and fish (`complete -C`), not just the engine: zsh inserts `brig run --network=isolated`.

Two cases still return `:none`, both deliberate and both with a test:

- **An inline path**, `--home=/pa`. `:dirs` and `:files` delegate to the shell and cannot carry the flag text, so the shell would complete a path against `--home=/pa` literally. Nothing is offered, which is what the inline spelling already did in zsh and fish. Fixable per-shell (`compset -P '*='` in zsh, re-prefixing in fish) if you think it is worth it — say so and I will file it.
- **A flag brig does not own**, which is the agent's word in either shape.

`script/smoke.sh` gained the argv-shape case; it returns `:none` on 69e6026, so it fails without the fix.

ananos: yes, let's file it -- inline path values are the one shape left out, and you already have the per-shell pieces (`compset -P '*='` in zsh, re-prefixing the token in fish). No need to grow this PR; we can iterate over the paths in the follow-up.

Thanks for the correction on the diagnosis -- the current word is indeed the right place, and pinning the completed-inline-then-refs case with a test is exactly what that branch needed. Good catch on the shell replacing the whole word, too; the flag-qualified candidates read well in practice.

ananos (approved):
Re-verified on 9b28ba5 against a build of the branch: the inline form now completes in all three shells (`--network=is` -> `--network=isolated`, `--from=cla` -> the profiles), the completed-inline flag still yields refs, the bash three-token shape is unchanged, and the suite plus `script/smoke.sh` pass. All findings from both rounds are closed.

Thanks @asapranidis, great work -- good to merge! Let's tackle the inline path values as a follow-up issue.

cmd/brig/completion.go
asapranidis: Filed as #159 — the affected set is every flag `flagValue` maps to `opDir`/`opFile` (`--home`, `-w`, `--workspace`, `--image`, `-t`) plus `--file`/`-f` on `secret create` and `secret update`, measured on a3ce070.

It records both per-shell pieces you named, and one thing that came out of writing it up: `qualify` collapsing `:dirs`/`:files` to `:none` means the engine has no way to tell a script which prefix to ignore, so the reply format needs to carry it (`:dirs --home=`). That makes it an ABI change across all three scripts rather than a zsh-only edit, which is worth knowing before someone picks it up.

PR #161, opened 2026-09-07
test(smoke): check that an attached policy reaches the gateway

panosmaurikos:
## Summary

`script/smoke.sh` has one policy case, and it is the refusal: a policy on `vz`,
which cannot enforce one, must stop the boot. That assertion passes even with
the whole feature deleted — an `attach` that writes a file and a boot that
reads nothing would refuse on `vz` just the same. Nothing checks that the rules
actually reach the gateway. `docs/manual-tests/egress-policy.md` measures that
by hand and says CI cannot run it.

This adds the other half. The stub runtime learns to fake `network-gateway`,
which is cheap: brig only wants `--egress-default` in its help and something
answering on the qemu socket. It filters nothing, on purpose — hull tests the
filter itself over its own netstack.

Checked by breaking the code on purpose rather than by watching it pass.
Dropping the rules in `startGateway` fails this case and none of the other 326.
So does making `isolatedNet` always true, and so does booting the guest against
the shared socket while the filtered gateway is up.

## Related issues

Refs #15

## Changes

- The stub runtime answers `network-gateway`: `--egress-default` in its help,
  and a listener on the qemu socket. It keeps `network-gateway` and the socket
  path in its own argv, which is what brig greps for before it records a
  gateway as one it owns — without that, `rm --all` has nothing to stop.
- Every rule of an attached policy has to reach the gateway's command line.
- The sandbox has to be booted against that exact socket, on a `/30`. Naming
  the socket rather than looking for `--gateway-sock`, which hull is passed for
  every networked run on this backend and so could not fail.
- A run with nothing attached has to stay on a shared gateway with no egress
  flag. Four claims, because three of them pass on an empty file.
- Without `python3` the case fails under `CI` and skips elsewhere, so the only
  automated proof of #15 cannot vanish while CI stays green.
- `docs/manual-tests/egress-policy.md` says which half CI now holds.

## Checklist

- [ ] ~~I have exercised the change against a real runtime (`brig run <agent>`), if it touches the run, exec or credential path~~ — this changes `script/smoke.sh` and a doc, not the run path

### DIFF
--- a/docs/manual-tests/egress-policy.md
+++ b/docs/manual-tests/egress-policy.md
@@ -5,7 +5,11 @@ now says a bound policy is enforced at the gateway brig gives the sandbox.
 This file is the measurement behind that sentence.
 
 CI cannot run any of this. `ci.yml` runs on Linux only and never boots a VM,
-and the enforcement point is a macOS backend.
+and the enforcement point is a macOS backend. What CI does hold is the half
+above the guest: `script/smoke.sh` asserts that the rules of an attached
+policy reach the gateway's command line and that the sandbox is put behind
+that gateway. Everything below -- a real kernel, a real resolver, and whether
+a packet actually stops -- is measured here and nowhere else.
 
 ## Result
 
--- a/script/smoke.sh
+++ b/script/smoke.sh
@@ -37,7 +37,8 @@ ok()  { printf '  ok   %s\n' "$1"; }
 bad() { printf '  FAIL %s\n' "$1"; fail=1; }
 
 # --- the stub runtime ---
-# It answers the four questions brig asks (ps, run, exec, stop/rm) and logs
+# It answers the questions brig asks (ps, run, exec, stop/rm, and the network
+# gateway one case below needs) and logs
 # every argument it is given, so the test can assert on what reached argv.
 cat > "$WORK/hull" <<'STUB'
 #!/bin/bash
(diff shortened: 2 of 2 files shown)


ananos (changes requested):
thanks @panosmaurikos, this closes the gap the manual doc describes and the
mutation check in the commit message is exactly what I wanted to see.

A few things though, mostly around the stub gateway. I ran the script on a Mac
three times at b5b3479 to confirm them.

**The stub gateway is never brig's.** After `exec python3 -c ...` the process
argv has no `network-gateway` in it, so `ownsGateway()` (which greps
`ps -o command=` for that string) is always false and `startGateway` never
writes the .pid/.spec record. The comment above the exec says the opposite.
`brig rm --all` then has nothing to stop, and a future case that re-runs the
same sandbox under a changed policy will hit "the gateway serving X is still
running and could not be stopped". One-line fix:

    time.sleep(120)' "$qsock" network-gateway "$@"

python ignores the trailing argv, brig gets the strings it keys on, and
`rm --all` stops the isolated gateway itself.

**The closing `pkill -f "network-gateway --socket $GW"` matches nothing**, for
the same reason. Both python listeners outlive the script by up to 120 s; three
local runs left 33 of them behind. The fix above covers the isolated one. For
the shared one, `pkill -f "$GW/"`, or have python exit when its socket path
disappears.

**"the sandbox is put on the network that gateway serves" cannot fail.** It
greps for `--gateway-sock` and `--gateway-cidr` anywhere in the log, and hull
passes both for every networked hvi run, shared or isolated. A regression that
starts the filtered gateway but boots the guest against the shared socket
passes all four checks. I would pull the `--socket` path out of gw-on.argv and
require that exact value on the `^argv: run` line (and a `/30` cidr).

Smaller ones:

- `BRIG_GATEWAY_DIR=$GW` is set only on the two `run` lines. `rm --all` and
  `policy detach` look in `~/.brig`, so the /30 and address allocated in $GW
  are never released. Export it for the block.
- `$GW` is outside `$WORK` and outside the EXIT trap, so an interrupt leaves it
  and the listeners behind. The length argument in the comment does not hold
  either: `$WORK/sandbox-brig-claude-code.sock.qemu` is 98 bytes on macOS,
  under the 103 limit. `GW="$WORK/gw"` would do, or add `$GW` to the trap.
- Neither hvi `brig run` checks its exit status. The stub logs argv before
  doing anything, so a brig that failed after spawning hull still passes.
  The sibling cases in the file check rc.
- `GATEWAY_SUBNET=198.18.0.0/24` duplicates a Go constant that has already
  moved once. Asserting the socket name starts with `gateway-` and the subnet
  is a `/24` tests the same property without the literal.
- The python3 skip is silent and uncounted, so the only positive test of #15
  can vanish while CI stays green. Maybe `bad` when `$CI` is set.
- The header still says the stub answers "four questions" and that the backend
  is pinned "rather than teach the stub to fake one". Both are now false, and
  docs/manual-tests/egress-policy.md still says CI cannot run this.

Also, could you move the body onto the PR template and tick the boxes? The
"checked by breaking the code on purpose" sentence belongs in the summary.

ananos (approved):
thanks @panosmaurikos, all addressed. Re-ran on a Mac at 8af1cb7: the .pid/.spec
record is now written for the stub gateway, `rm --all` takes the isolated
listener down, nothing outlives the script, and the socket path under $WORK/gw
measures 101 as the comment says. The exact-socket + /30 assertion is the one
that matters and it holds.

Two tiny leftovers, no need for another round: the stub header comment has a
line break mid-sentence ("and logs / every argument"), and the old "exec, so the
process brig started..." sentence now sits above the paragraph that explains
the same thing. Feel free to squash them in if you touch the file again.

Issue #196, opened 2026-09-09
brig --help says a session label reaches the agent, and it reaches one of eight


pmoust:
## Describe the bug

The session paragraph in `brig --help` states the label reaches the agent as its display name, with no qualification. It reaches `claude-code`, on the run path, and nothing else.

## The claim

`cmd/brig/main.go` usage text:

> A `<ref>` is the session. `claude` is that agents default session, and `claude@refactor` is a session of its own -- its own workspace, its own sandbox, and the label reaching the agent as its display name.

## What the code does

`cmd/brig/main.go:724-729`:

```go
func agentArgs(cfg *wrap.Config, t profile.Profile, tail []string) []string {
	if cfg.RawName != "" && t.Name == "claude-code" {
		return append([]string{"--name", cfg.RawName}, tail...)
	}
	return tail
}
```

Two limits, neither stated:

1. **`claude-code` only.** The gate is a literal name comparison. `codex@refactor`, `gemini@refactor` and the rest get their label used for the workspace and the sandbox, and never see it.
2. **Run only.** `agentArgs` has one call site, `main.go:711`, on the run path. `brig sh claude@refactor` does not pass it.

The `@` form does populate `RawName` (`main.go:1415` assigns `ref.Label`), so the helps own example works. It is the generalisation that does not.

## Expected behavior

The sentence says which agents take a display name, or drops the clause. The other two properties in it, its own workspace and its own sandbox, are true for every agent, so the reader has no way to tell that the third is different.

## Actual behavior

A user running `codex@refactor` sees the same session ref in `brig ls` and no display name in the agent, with nothing explaining the difference.

## Environment

brig 0.1.0-rc18 (d11ba32). Read from source, and by tracing `RawName` through `config.go:276-281` and `:416`. Not run end to end: it needs two agents booted and logged in.

## Additional context

Found while writing docs/sessions.md for #111, which now states the narrow rule. The help text and the docs disagree until this is settled, so it is worth deciding which is right rather than leaving them apart.

`--name` is also Claude Codes own flag, which is presumably why the mapping exists for that one agent. If it stays agent-specific, the profile is the natural place to declare it rather than a name comparison in `main.go`.

Issue #180, opened 2026-09-09
Runtime messages point at sections of the README, one by line number


pmoust:
## Describe the bug

Four places in the shipped binary name the README as where a fact is written down. Two of them reach users at runtime. One pins a line number.

## Where

Printed to users:

```
$ brig --help
settings (BRIG_<AGENT>_<KEY> wins over BRIG_<KEY>; see the README for all):
```
`cmd/brig/main.go:136`

```
$ brig telemetry status
The telemetry section of the README lists every field an event carries, who
```
`cmd/brig/telemetry.go:27`

In comments and tests:

- `internal/exitcode/exitcode.go:4` and `:25` name "the exit-code table in README.md beside the command reference"
- `cmd/brig/exit.go:11` the same
- `cmd/brig/doctor.go:152` names "README around line 219"
- `cmd/brig/exit_test.go:14` and `notfound_test.go:89` pin it in tests

## Why it is a problem

The README is the front door and the file most likely to be restructured. Any reorganisation of it silently invalidates strings inside a released binary, and a line-number reference is stale as soon as a paragraph is added above it.

The coupling is backwards. Help text can name a stable reference page, or name nothing.

## Expected behavior

Runtime strings name a page under `docs/` whose job is to hold that fact, or drop the pointer. Comments name a file rather than a line.

## Actual behavior

`brig telemetry status` sends a user to a section of a file that has just been reorganised.

## Environment

brig 0.1.0-rc18 (d11ba32).

## Additional context

This is live now: #111 moves the exit-code table and the environment-variable table out of the README onto reference pages, so those comments need updating in that branch. Filing it separately because the underlying coupling will recur, and because `doctor.go:152`s line number is the shape worth removing rather than re-pointing.

Issue #173, opened 2026-09-09
brig stop promises a restart that is really a fresh boot


pmoust:
## Describe the bug

`Stop()`s own comment says a stopped sandbox starts again as a boot rather than a fresh creation. There is no start path in the code.

## Cause

- The `Runtime` interface has `Run`, `Stop`, `Remove` and `Running`, and no `Start` (`internal/runtime/runtime.go:250-265`).
- Both adapters build `run --detach --name <name>` (`hull.go:433`, `nerdctl.go:315`).
- `EnsureRunning` removes the stopped instance before every boot (`internal/wrap/run.go:278-279`).

So every `brig run` after a `brig stop` creates a new sandbox.

## Effect

Nothing is lost. Everything persistent is in the workspace, so the user gets the outcome they wanted. The cost is that the code and its own comment disagree at `internal/wrap/run.go:601-603`, and the prose repeated it.

## Expected behavior

Either implement a start path, or retire the claim and say what `brig stop` actually keeps: the sandbox name, its row in `brig ls`, and what brig recorded about the session. `brig rm` drops the last of those too.

## Actual behavior

A reversibility that is described and not implemented.

## Environment

brig 0.1.0-rc18 (d11ba32).

## Additional context

Found while documenting the lifecycle for #111. The docs are being corrected there to describe what is kept rather than to characterize the restart. This issue is for the comment and for the decision about whether a real start path is wanted.

Issue #172, opened 2026-09-09
A session named with --name cannot carry a policy, and says nothing


pmoust:
## Describe the bug

A per-session policy attachment is stored under the slug and looked up under the name as typed. The two agree for the `@` form and can differ for `--name`, so a session reached through `--name` silently runs with no policy.

## Cause

Attach refuses any name that is not already slug-clean (`cmd/brig/policy.go:645-663`):

```
brig policy attach locked-down claude-code -n Foo
session "Foo" is not usable as it stands: it would have to become "foo". Type "foo" if that is the session you want
```

So the attachment can only ever be stored under `foo`.

The run looks it up with `rawName`, the name as typed (`internal/wrap/config.go:405`):

```go
egress, policies, err := policy.Resolve(t, rawName, policy.Dir())
```

`EffectivePolicies` then does a plain map lookup with no slugging (`internal/policy/effective.go:36`):

```go
add(a.Sessions[p.Name][session])
```

`brig run claude --name Foo` therefore looks for `Foo` and never finds `foo`.

`--name` is deliberately the lenient spelling: `session.Resolve` sanitizes it and warns which directory it got (`internal/session/ref.go:36-38`). The policy lookup did not follow it.

## Expected behavior

Either the lookup slugs the session name the way the workspace and sandbox names already do, or `--name` refuses a name that would not round-trip, as `@` does.

## Actual behavior

The run proceeds with no policy and no message. `brig policy check claude -n Foo` refuses the question instead of answering it, so the state cannot be inspected either.

The `@` form is unaffected: its labels are slug-clean by construction.

## Environment

brig 0.1.0-rc18 (d11ba32). Read from source. Not reproduced end to end, because that needs writing a policy into the real config directory.

## Additional context

A policy that silently does not apply is worse than one that refuses, since the user has asked for a restriction and been told nothing. Related in shape to #171, where the enforcement gate fails open.

`--name` is retired toward `<agent>@<label>` and is scheduled for removal in 0.3, so one option is to leave the lookup alone and refuse a non-slug `--name` for the remaining window.

PR #197, opened 2026-09-09
Rewrite the README and docs around what the code does

### LINKED ISSUE #111: Rewrite the README and docs around the new nouns

pmoust:
Closes #111.

The docs described a model brig has outgrown, and the README had grown to 9162 words. This rewrites the set against the source at d11ba32 (0.1.0-rc18, which is also the newest tag, so nothing here documents an unreleased command).

## The correction that matters most

Pages still said one host directory is the agent whole world. A run mounts the guest home and, separately, the project named on the command line, read-write at `/work/<name>`. **An agent can change those files.** Every page that described the boundary understated it, including the diagram.

## Other accuracy fixes

- **Verification modes are `off`, `warn` and `require`, and `warn` is the default.** An unverifiable image is reported and booted unless the user asks for `require`. The tag path cannot pin a digest.
- **Three macOS backends, not two.** Six of the eight shipped profiles ask for `hvi`, which needs macOS 15 and is the only backend that enforces egress policy or `isolated` networking. `vz` is the default and the only one that can show a window. doctor `Hypervisor.framework available` line is the `kern.hv_support` probe, not a backend.
- **Persistence is per profile.** The login-in-memory behavior belongs to `claude-code` and `claude-desktop`, the two that declare a tmpfs. The other five write to host disk.
- **Only a required secret fails a run**, and no shipped profile declares one.
- **`brig stop` does not keep a bootable instance.** There is no start path, so the docs now say what it actually keeps.

## Structure

README is 1251 words and links out. New pages carry what it used to hold: `install.md`, `sessions.md`, `authentication.md`, `cli.md`, `telemetry.md`, plus `migration.md`, `stability.md` and a `docs/README.md` index. The manual-test reports are labeled historical, with their dates and results intact.

`docs/stability.md` is also the stability statement #47 notes is missing from the repository. It does not settle #60; it records what is true today.

## One new check

`script/check-retired-spellings.sh` fails when a doc teaches a spelling that 0.3 removes. Every retired spelling still works and prints its replacement, so such a doc is not broken today and becomes wrong at 0.3, which is what #111 asks to automate. It reads commands rather than prose, so "six of the shipped brig profiles" is left alone while `brig profiles` is caught. Not yet wired into CI.

## Go changes

Comments and help text only, no behavior. Moving the exit-code and environment tables out of the README left nine pointers aimed at sections that no longer exist, including a user-facing `brig telemetry status` string and a line-number reference. They now name the pages that hold those facts. Filed separately as #180, since the coupling will recur.

## Defects found, not papered over

Checking these claims turned up 33 issues, filed as #164 to #196 rather than smoothed away in the wording. The docs describe the behavior as it is and link the issue where one exists. The ones that shape this PR text: #167 (verification messages), #169 (`--skills` is a no-op on seven of eight profiles), #170 (`statePaths:` is inert), #172 (a `--name` session cannot carry a policy), #173 (`brig stop`), #196 (the session label reaches one agent).

## Validation

- `go test ./...` passes, 13 packages.
- `go vet ./...`, `gofmt -l` clean.
- `script/smoke.sh` PASS.
- `script/check-retired-spellings.sh` passes, and still catches all 9 planted cases in a self-test.
- Every relative link and anchor resolves, checked by script across all Markdown in the repo. The paths cited from Go source and shell scripts still resolve, including `docs/policies.md` "What this does not do yet" and `docs/runtimes.md#building-hull-from-source-on-macos`.
- No orphan pages.

**Not validated at runtime.** No sandbox was booted, so nothing here is evidence that a documented run works end to end. Claims were checked against the implementation, its tests, and read-only commands against a build of this commit. Untested paths: Linux, Intel Mac, macOS 14, and any flow needing real credentials.

## Review notes

Best read as: README, then `docs/README.md`, then `cli.md` and `sessions.md`, which carry the most new material. The manual-test reports have header-only changes.

Two accuracy passes were still running when this went up; anything they find will land as follow-up commits here.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -55,9 +55,10 @@ info` reports the backend, and `hull ps` / `nerdctl ps` shows the instance.
 ## `brig doctor` output
 
 <!--
-Paste it in full, fenced. `brig doctor` reports every check it can and exits
-0 even when a line reads `!!`, so the whole report matters more than whether
-the command "passed".
+Paste it in full, fenced. `brig doctor` prints a fix beside every line that
+is not `ok`, and still exits 0 for most of them. Only a missing or broken
+runtime and an unreachable secret store set a nonzero status, so the whole
+report matters more than whether the command "passed".
 -->
 
 ## Logs and additional context
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -43,13 +43,22 @@ info` reports the backend, and `hull ps` / `nerdctl ps` shows the instance.
 
 <!-- Fill in what applies; delete the rest. -->
 
+- Command: <!-- the exact `brig ...` line that failed -->
 - Host OS: <!-- e.g. macOS 26.3, or Ubuntu 24.04 -->
 - Architecture: <!-- arm64 | amd64 -->
-- brig version: <!-- brig --version -->
+- brig version: <!-- run `brig version` and paste its output -->
 - Runtime and version: <!-- hull --version, or nerdctl --version -->
 - Profile: <!-- claude-code, codex, ... or a file-backed one -->
 - Installed how: <!-- Homebrew cask | install.sh | make build -->
 - Commit: <!-- git rev-parse HEAD, if built from source -->
+
+## `brig doctor` output
+
+<!--
+Paste it in full, fenced. `brig doctor` reports every check it can and exits
+0 even when a line reads `!!`, so the whole report matters more than whether
+the command "passed".
+-->
 
 ## Logs and additional context
 
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -32,11 +32,12 @@ from the right place.
 <!--
 Delete if there is none.
 
-brig's value is that the guest sees one directory and holds only the
-credentials it was named. A proposal that widens either -- a new mount, a new
-forwarded variable, a new host path the guest can reach -- is still worth
-making, but say so plainly here so the tradeoff is discussed rather than
-discovered. docs/security.md is the place those limits are written down.
+brig's value is that the guest reaches only the host directories brig names
+for it, and holds only the credentials it was named. A proposal that widens
+either -- a new mount, a new forwarded variable, a new host path the guest can
+reach -- is still worth making, but say so plainly here so the tradeoff is
+discussed rather than discovered. docs/security.md is the place those limits
+are written down.
 -->
 
 ## Additional context
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -24,9 +24,10 @@ Check items as you complete them; strike through (~~like this~~) any that do
 not apply, rather than deleting or rewording them. Keep the reasoning in
 Summary or Changes, not here.
 
-`make all` is vet + test + build, which is what CI gates on. `script/smoke.sh`
-drives the real binary end to end against a stub runtime and a stub cosign, so
-it runs anywhere -- no VM and no macOS needed. See CONTRIBUTING.md.
+`make all` is vet + test + build. Run it locally before you push: CI does not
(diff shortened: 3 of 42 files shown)


pmoust:
@claude-fleet review

pmoust:
Two accuracy passes finished after this went up. Both found real errors, and the fixes are in 542fbb0. Recording what they caught, since it changes what is worth re-reading.

**Three that mattered:**

- **`docs/telemetry.md` said two things are counted. It is three.** A sandbox stop is counted (`hull.go:625` passes `counted=true`), including the stop inside a restart; `brig rm` is not (`hull.go:634`). Wrong on the page whose whole job is to say what leaves the machine.
- **`docs/security.md` asserted an absolute that `hostCredential:` falsifies.** "No host credential source is read on the run path" holds for every shipped profile and is false for a profile of your own carrying that key, which reads the named keychain item on every run (#190). Now scoped, with the exception named.
- **`claude-desktop` does carve `.claude/plugins` out of its tmpfs** (`claude-desktop.yaml:65`). The list omitted it, so a reader would conclude anything written there is discarded.

**The other eight:** doctor exits nonzero for two checks rather than never; brig picks the `vz` fallback itself and never leaves it to hull (`hull.go:435` always passes `--hypervisor`); an offline run is exempt from the policy refusal (`hull.go:379`, `spec.Net != "none"`); a stale project is the third restart trigger, not a second; the image row is filled in rather than added by naming an agent; completion skips `-q` because it is retiring, not because it lacks a run-line row; and the `hostConfigDir` row had lost the `--skills` precondition that makes it mean anything.

**One gate defect, worth flagging on its own.** `docs/profiles.md` taught `brig run claude --name desktop` wrapped across a line break inside a single backtick span. `check-retired-spellings.sh` reads line by line, so it could not see it and passed. Fixed the example, and taught the extractor to join a line holding an odd number of backticks to the next one before matching, reported at the first line number. A planted wrapped case now fails the check. That is the one class of error this gate exists to catch, and a reviewer caught it first.

Everything re-run after the fixes: `go test ./...`, `go vet`, `gofmt -l`, `script/smoke.sh`, the retired-spelling gate (still 9/9 on its self-test), and the link and anchor sweep. All pass.

The runtime limitation in the description stands unchanged: no sandbox was booted, so nothing here is evidence a documented run works end to end.

PR #201, opened 2026-09-11
feat(profile)!: remove hostCredential: and its run-path keychain read

### LINKED ISSUE #190: hostCredential still reads the host keychain on every run
## Describe the bug

`hostCredential:` is deprecated and still live. While it parses, the headline credential promise is conditional on which profile is in use.

## What it still does

`internal/profile/profile.go:219` parses it. On every invocation, when the environment supplied nothing for its `targetVar`, brig reads the named macOS keychain generic-password item through `/usr/bin/security find-generic-password -s <service> -w` (`internal/wrap/run.go:62-78`, `:109-154`).

That is a host credential source read on the run path, which is what the current design says does not happen.

`internal/hostsrc/arch_test.go:31-35` names this field as the reason the architecture test cannot yet assert "a run reads no host source".

## Expected behavior

Remove it, and let the architecture test make the unconditional claim.

## Actual behavior

A live hole. Both README.md:123-125 and docs/security.md:78-84 disclose it, so the docs are not wrong, but a field that defeats the promise is worth closing rather than describing.

Only a file-backed profile can trigger it. No built-in declares it, so no shipped profile warns about itself.

## Environment

brig 0.1.0-rc18 (d11ba32), macOS 26.5 arm64. Read from source. Not exercised: reproducing means writing a profile with `hostCredential:` and a real keychain item.

## Additional context

Already scheduled for removal as a deprecated key. Filing it because the deprecation window is the period during which the security claim carries an exception, so the removal date is a security decision rather than a tidiness one.

#22 asks for a test per public security claim, and `internal/hostsrc/arch_test.go` is that test for this claim, waiting on this field.

asapranidis:
## Summary

`hostCredential:` was deprecated in rc16 with a notice that it goes in the next release. Two releases later it was still live: a profile file carrying it made every `brig run` read a macOS keychain item through `security find-generic-password`, which is the one thing docs/security.md says a run does not do. Only a user's own file could trigger it, and the docs disclosed it, but the deprecation window was the period during which the security claim carried an exception, so the removal date is a security decision rather than a tidiness one.

This removes the field and everything that existed only to serve it. A file that still carries the key is refused at load, with an error naming the key and `brig secret import <profile>` as the replacement, rather than parsed with the block dropped: a sandbox that boots without the login its owner configured reports the loss as an authentication prompt inside the guest, the last place anyone connects back to a key in a file. Every other command keeps working, because an unusable profile is reported rather than fatal, like any file that fails to parse.

The architecture test in `internal/hostsrc` drops its caveat and gains a second assertion, so the claim it defends is now "a run reads no host source" rather than "no new host source reaches the run path".

Drafted with Claude Code; every line reviewed by the author.

## Related issues

Closes #190. Refs #22.

## Changes

- `internal/profile`: remove the `HostCredential` type and field. `Parse` refuses a document carrying `hostCredential:` by name, before the strict decoder reports it as an unknown field, so the error can say what to write instead.
- `internal/creds`: remove `ReadHost`, `AdmitHost`, `HostCredential`, `KeychainRead` and the `/usr/bin/security` shell-out. The package no longer imports `os/exec` or `jsonfind`.
- `internal/wrap`: remove the keychain fallback in `BuildEnv`, `admitHostCredential`, the `ReadKeychain`, `HostCred` and `AllowExpired` fields on `Config`, the `hostCredential`-driven "guest login" lines in `Status` and `brig info --json`, and the now-unused `sourceOf`. `BRIG_ALLOW_EXPIRED` is gone with the only path that read it.
- `cmd/brig`: remove `warnDeprecatedProfileKeys`.
- `internal/hostsrc/arch_test.go`: new `TestTheRunPathReadsNoKeychain` asserts no non-test source in `internal/wrap`, `internal/creds` or `cmd/brigd` names the keychain tool. It failed against `main` on `internal/creds/creds.go`.
- Tests: `internal/wrap/hostcred_test.go` deleted with the path it covered. New tests assert the refusal at parse and at CLI load, each naming the replacement command.
- Docs: `security.md`, `secrets.md`, `profiles.md`, `migration.md` and `cli.md` no longer describe the exception, and the security claim is stated without the "on every profile brig ships" qualifier.

## Checklist

The real-runtime box is unchecked on purpose: the change removes a code path rather than altering one that remains, and the run-path behaviour that is left, credentials from brig's own store, is unchanged and covered by the existing suite. `script/smoke.sh` was run on this branch and on `main`, with identical results.

This touches the second promise, that the guest gets the credentials it was named and no others. Before this change a run on a user's own profile could reach a host credential the user had not put in brig's store. After it, no run-path package can: `TestTheRunPathCannotReachTheImporter` holds the importer out, and `TestTheRunPathReadsNoKeychain` holds the keychain tool out. The second is the negative test that fails without this change.


pmoust:
@asapranidis ~LGTM - please take care of conflicts so we can get it in~ . See below:

docs/profiles.md
```diff
@@ -132,219 +125,101 @@ brig agent export x | brig agent import -
...
+| `hostConfigDir`, `projectPaths` | no | Where the user's own agent configuration lives on the host, and which subdirectories of it to seed into the workspace, and only when the run passes `--skills` or sets `BRIG_SKILLS=1`. Both fields are required together, and only `claude-code` declares them, so `--skills` does nothing on the other seven: see [What this costs](#what-this-costs) below |
 | `onboarding` | no | A first-run state file to seed. See below |
-| `hostCredential` | no | **Deprecated, removed next release**, see [migration.md](migration.md#profile-keys). A credential read from the host keychain on every run when the environment carries none. Replaced by `secrets` with `sources`, filled once by `brig secret import`. See below |
-| `reserved` | no | Marks a profile that owns the guest home a session name can otherwise slug onto. See above |
+| `hostCredential` | no | **Removed**, see [migration.md](migration.md#profile-keys). A file that carries it is refused. It read a credential from the host keychain on every run; `secrets` with `sources`, filled once by `brig secret import`, replaces it. See below |
```
pmoust, line 179: I'd rather we remove the line altogether, and mention it in the changelog. It shouldn't be in the docs in perpetuity. Same applies across every doc touched, just commenting on one line to not be pedantic.

pmoust (changes requested):

pmoust:
@claude-fleet review

pmoust (approved):

PR #202, opened 2026-09-11
feat(rm)!: say what rm removes, ask before --all, and add --dry-run

### LINKED ISSUE #11: Say what `reset` and `rm` will remove, and add `--dry-run`
### Problem

`brig reset` stops and removes every brig sandbox. It asks nothing and prints
no list of what it will act on. Workspaces survive,
which is the safe and correct behaviour, and the command line gives the user
no way to know that before pressing return.

### Proposed solution

1. `brig reset` prints the sandboxes it will remove, then asks for
   confirmation. `--force` or `-y` skips the question, for scripts.
2. Both `reset` and `rm` print the line that workspaces stay on the host.
3. Add `--dry-run` to `reset` and `rm`. It prints the same list and exits 0.
4. Accept `brig rm --all` as the spelling of `reset`, and keep `reset` as an
   alias.

asapranidis:
## Summary

`brig rm --all` stopped and removed every brig sandbox without saying which ones, and nothing on the command line previewed the removal. Workspaces were never touched, but the line did not say so until after the fact.

`rm --all` now lists what it is about to remove, one ref and its state per line, says that workspaces stay on the host, and asks. `-y`/`--yes` answers in advance, the spelling `secret delete` and `agent rm` already take. Without a terminal and without `-y` it refuses and names the flag, the same rule `confirmDelete` follows: assuming yes would make the scripted case the one that cannot be stopped. `--dry-run` on both `rm --all` and `rm <ref>` prints the same preview and exits 0 with nothing removed. `brig reset` takes the same flags.

The flags are read in dispatch before the run line, the way `--all` already was, rather than through `brigFlags`. A `--dry-run` in that table would be brig's on every run-line verb, and `brig run opencode --dry-run` has to keep reaching opencode. The `--force` spelling the issue suggested is not taken: in brig it means "replace a file that is already there".

Item 4 of the issue, `rm --all` as the spelling with `reset` as an alias, had already landed.

**Breaking:** `brig rm --all` and `brig reset` without a terminal on stdin now refuse unless `-y` is passed. An unattended script needs `brig rm --all -y`. `script/smoke.sh` is updated the same way.

Drafted and implemented with Claude Code; every line reviewed by the author.

## Related issues

Closes #11

## Changes

- `rm --all` and `reset` print the sandboxes they will remove, say workspaces stay on the host, and ask `[y/N]`; `-y`/`--yes` answers in advance; with nothing to remove nothing is asked
- Without a terminal and without `-y`, `rm --all` refuses and names the flag
- `--dry-run` on `rm --all` and `rm <ref>`: prints the preview, exits 0, removes nothing; a ref with no sandbox is still exit 3
- `rm <ref>` reports the workspace it left behind; `rm <ref> -y` is refused by name
- `takeAll` becomes `takeRemoveFlags`; `removeAll` goes through the injectable `detectRuntime` so tests can drive it with a fake
- Completion offers `--dry-run` on `rm`, and `--yes`/`-y` beside `--all`
- Help text and `docs/cli.md` `brig rm` section updated
- Tests: `cmd/brig/remove_test.go` (12 cases, two through a real pty for the typed yes and the hung-up terminal); the three existing tests that asserted `rm --all --dry-run` is refused now assert the opposite
- `script/smoke.sh`: every `rm --all` gains `-y`; the argument-hygiene section covers the refusal, both dry-runs, `rm <ref> -y`, and an unknown flag beside `--all`

## Checklist


cmd/brig/main.go
```diff
@@ -1910,7 +1963,18 @@ func removeSandbox(cfg *wrap.Config, ref string) error {
 		// entry for a sandbox that is truly gone costs nothing until then.
 		return noSandboxf(ref)
 	}
-	return cfg.Remove()
+	if dryRun {
+		fmt.Printf("would remove %s (sandbox %s). The workspace %s stays on the host\n",
```
ananos: The preview goes to stdout here, but the removal it previews goes to stderr through `warnf`, and `rm --all --dry-run` splits the two (list on stdout, the workspace notice through `warnf`). So this is the one rm line whose stream flips between the preview and the real thing, and the only one `-q` does not take away.

I would split it the way `--all` does:

```go
fmt.Printf("would remove %s (sandbox %s)\n", ref, cfg.VMName)
warnf("The workspace %s stays on the host", cfg.Workspace)
```

Then `brig rm claude --dry-run > list` and `brig -q rm claude --dry-run` behave like their `--all` counterparts. What do you think?

cmd/brig/remove_test.go
```diff
@@ -0,0 +1,291 @@
...
+	rt := twoSandboxes()
+	removeHost(t, rt)
+	for _, args := range [][]string{{"rm", "faker", "-y"}, {"rm", "-y", "faker"}, {"rm", "--yes", "faker"}} {
+		_, err := run2(t, args)
+		var ue *usageError
+		if !errors.As(err, &ue) {
+			t.Errorf("brig %s: %v, want a usageError", strings.Join(args, " "), err)
+			continue
+		}
+		if !strings.Contains(err.Error(), args[1]) && !strings.Contains(err.Error(), args[2]) {
```
ananos: This passes if the message names either the flag or the ref, since `args[1]` and `args[2]` are those two in some order. A regression that drops the flag and keeps the ref stays green, and naming the flag is what the comment above asks for.

Maybe carry the flag in the table:

```go
for _, tc := range []struct {
    args []string
    flag string
}{
    {[]string{"rm", "faker", "-y"}, "-y"},
    {[]string{"rm", "-y", "faker"}, "-y"},
    {[]string{"rm", "--yes", "faker"}, "--yes"},
} {
```

and assert on `tc.flag`.

cmd/brig/main.go
```diff
@@ -352,15 +356,34 @@ func dispatch(args []string) error {
 		// was spelled as if it were a setting being restored. --all is that
 		// word, on the verb that already removes one.
 		deprecated("brig reset", "brig rm --all")
-		return removeAll("brig reset", rest)
+		others, o, err := takeRemoveFlags(rest)
+		if err != nil {
+			return err
+		}
+		return removeAll("brig reset", others, o)
```
ananos, line 369: `o.all` is read off the reset line and then dropped, so `brig reset --all` works and says nothing about it. Same for a repeated `--all` on rm. No harm done either way, but this is the one path where the PR's own rule -- refuse what you would otherwise read past -- does not apply, so it would be good to have it as a decision rather than a leftover. Up to you.

ananos (commented):
Looks good. The preview, the question, and the refusal when there is no terminal are all the right shape, and reading rm's flags in dispatch rather than through `brigFlags` is the right call -- `brig run opencode --dry-run` has to keep reaching opencode.

I built the branch and ran the suite here: `go build ./...`, `go vet ./...` and `go test ./...` are all green. I also checked that `wrap.Load` writes nothing, so `--dry-run` really does leave the host alone, and the `ForgetSandbox`/`ForgetSlugClaim` pruning and the `PruneNetworks` re-list came through the refactor unchanged.

Three small things below, none of them blocking.

ananos (approved):
Approving. All three notes from the last pass are addressed, and the reset one went further than I asked: `brig reset --all` is refused by name now rather than just documented, with the three spellings under test.

I ran the merged head here. Build, vet and `go test ./...` are green, the thirteen rm/reset tests pass, and `brig reset --all` refuses at exit 2 without touching a sandbox. The docs merge came out clean too, the `brig rm` section survived main's rewrite intact.

Two notes for the record, neither of them blocking.

The branch carries a merge commit. We merge rebase-only, so it gets dropped on replay and nothing of it reaches main. No harm done, but a rebase onto main would keep the branch linear if you prefer.

`docs/migration.md` still says only that `brig reset` needs `-y` without a terminal. It does not mention that `brig reset --all` is refused outright now. The error message teaches the new spelling, so this is minor, maybe add a line when you next touch that file.

Issue #165, opened 2026-09-09
One user mistake returns three different exit codes


pmoust:
## Describe the bug

A stray operand or an unknown verb is one class of mistake. brig answers it with 0, 1 or 2 depending on which verb was typed. The README documents 2 for this class, so a script that branches on the status is wrong for some verbs.

## To reproduce

All on 0.1.0-rc18:

| command | exit |
| --- | --- |
| `brig ls extra` | 2 |
| `brig agent ls extra` | 2 |
| `brig doctor claude extra` | 2 |
| `brig secret ls extra` | 1 |
| `brig telemetry status extra` | 1 |
| `brig policy ls extra` | 0 |
| `brig version extra` | 0 |
| `brig completion bogus` | 2 |
| `brig nosuchverb` | 1 |
| `brig telemetry bogus` | 1 |
| `brig agent` | 1 |
| `brig policy` | 1 |
| `brig secret` | 1 |

## Expected behavior

One code for one class. The README assigns a usage error to 2.

`brig nosuchverb` matters most: an unknown command is the most likely thing a wrapper script hits, and it returns 1, the general-failure code.

## Actual behavior

Eight sites return a plain `errors.New` or `fmt.Errorf` and fall through `exitCode` to 1:

- `main.go:413` unknown command
- `main.go:497` `<verb> needs a profile`
- `main.go:2257` agent needs a subcommand
- `policy.go:50` policy needs a subcommand
- `policy.go:831` policy check needs a profile
- `secret.go:40` secret needs a subcommand
- `secretimport.go:559` import needs a profile
- `telemetry.go:52` unknown telemetry subcommand

Their counterparts call `usagef` and exit 2: `completion.go:83`, `main.go:1532`, `main.go:1105`.

The two zeros are #164.

## Environment

brig 0.1.0-rc18 (d11ba32), macOS 26.5 arm64.

## Additional context

No test pins any of these to 1. `lifecycle_test.go:134-141` asserts the message text only, so nothing blocks the change.

#145 proposes one verb table read by dispatch and the predicates. That table is the natural place to settle which refusals are usage errors.

Issue #145, opened 2026-09-05
One verb table for the CLI, read by dispatch, the predicates and completions


pmoust:
## Problem

`cmd/brig/main.go` dispatches on a bare `switch verb`, and three predicates
beside it each carry their own hand-maintained list of verbs: `forwardsTail`
(which verbs hand a tail to the agent), `honorsRunLine` (which read a
run-line flag), and `verbTakesGlobalJSON` (which take `--json`). Each is
reasonable alone. Together they are a drift surface, and none of them fails
when a new verb forgets to join.

`brigFlags` already shows the shape that works: one row per flag carrying its
properties, read by every predicate, with a test that walks the table.

## Proposed solution

One verb table, one row per verb, carrying what the predicates ask today:
forwards a tail, reads run-line flags, takes `--json`, is deprecated toward
which spelling. The switch, the three predicates, `usage`, and the completion
scripts (#12) read it. A test asserts every verb the switch dispatches has a
row and every row is dispatched.

## Additional context

Raised in review of #133 and #141 while several grammar PRs were open at
once. #12 will need the same list for completions, which is the moment to
build it rather than a fourth copy.

PR #203, opened 2026-09-11
fix(cli): refuse a trailing word on the policy listing and version

### LINKED ISSUE #164: brig policy ls and brig version swallow trailing arguments
## Describe the bug

`brig policy ls` and `brig version` accept any trailing token, ignore it, and exit 0. `brig policy ls --help` prints the policy listing instead of help.

Every other listing verb names the stray word and exits 2. `trailing_test.go:11-16` states the norm: "Every other verb in brig names a stray token rather than swallowing it, and now so does every verb here". That test table at `trailing_test.go:19-30` covers only the ten agent subcommands, so nothing catches these two.

## To reproduce

```
$ brig policy ls extra
no policies yet; your own live in /Users/pmoust/.config/brig/policies
brig policy create <name> writes a starter one
$ echo $?
0

$ brig policy ls --help
no policies yet; your own live in /Users/pmoust/.config/brig/policies
brig policy create <name> writes a starter one
$ echo $?
0

$ brig version extra
brig v0.1.0-rc18
$ echo $?
0
```

Contrast, same release:

```
$ brig ls extra
brig: unexpected argument "extra"; `brig ls` takes no arguments other than -q and --json
$ echo $?
2
```

## Expected behavior

Both name the stray token and exit 2, as `brig ls`, `brig agent ls` and `brig doctor claude extra` do.

## Actual behavior

The token is dropped and the command succeeds. A user who typed `--help` gets output that is not help and a status that says the command worked.

## Cause

`policy.go:57-58` is `case "ls": err = listPolicies()`, and `listPolicies()` at `policy.go:115` takes no parameters, so `args[1:]` is never read. The deprecated `case "list"` at `policy.go:61-63` has the same hole. `main.go:288-290` handles `version` without reading `rest`.

## Environment

brig 0.1.0-rc18 (d11ba32), macOS 26.5 arm64, hull 0.1.0-rc21.

## Additional context

Found while checking the CLI surface against the docs. #145 proposes one verb table read by dispatch and the predicates, which is the structural fix. These two are small enough to correct before that lands, and a row in the `trailing_test.go` table would keep them fixed.

marrgko:
## Summary

`brig policy ls extra` printed the listing, dropped `extra` and exited 0, and
so did `brig version extra`. Every other listing verb names a stray token and
exits 2 — `brig ls extra` and `brig agent ls extra` both do — so these two
answered a mistake as if it were a command, and a script branching on the
status could not tell them apart from success.

`brig policy ls --help` was the worst spelling: it printed the listing, which
is not help, with a status that said the command worked.

Three spellings reach the policy listing, not the one the report reproduces.
The retired `brig policy list` and the retired top-level `brig policies` call
the same `listPolicies`; `brig --version` sits alongside `brig version`. All
five now refuse a stray word.

`--help` is answered rather than refused. `policyCmd` already translates a
verb's `flag.ErrHelp` into the group's usage and exit 0 — `policy.go` says so
where it does it — so `ls` returns `flag.ErrHelp` for a help token and is
answered there, in the same words `brig policy show --help` gets.

## Related issues

Closes #164
Refs #145
Refs #165

## Changes

- `rejectPolicyTail` refuses a stray word for the policy listing, and returns
  `flag.ErrHelp` for a help token so the group's existing translation answers
  it. `listPolicies` stays a printer: it takes no operands and no flags, and
  eleven tests call it as one.
- `brig policies` is routed onto `ls` the way `deprecatedProfileCmd` routes
  onto `agentCmd`, so the retired spelling inherits both answers instead of
  keeping a second copy of either.
- `brig version` and `brig --version` refuse a trailing word, naming the
  spelling that was typed — neither retires the other, so there is no newer
  word to send the reader to.
- The retired spellings refuse in `brig policy ls`'s words rather than their
  own, matching what the retired agent spellings already do: the deprecation
  notice has just sent the reader to that verb.
- Six tests in `cmd/brig/trailing_test.go`, including one asserting that four
  sibling verbs still answer `--help`, so a change that stopped translating
  `flag.ErrHelp` for the whole group fails here too.

Behaviour note for reviewers: `brig version --help` now exits 2 rather than
printing the version. That follows from `version` reading its tail at all,
and matches `brig ls --help` and `brig doctor --help`. See the comment on
this PR for the full comparison.

## Checklist

- [ ] ~~I have run `go test ./... -race`, if the change touches concurrency,
      subprocesses or the daemon~~ — dispatch and argument parsing only.
- [ ] ~~I have exercised the change against a real runtime (`brig run
      <agent>`), if it touches the run, exec or credential path~~ — touches
      neither; `smoke.sh` drives the real binary for these verbs.
- [ ] ~~I have updated the affected docs~~ — no doc states what these verbs do
      with a trailing word, and the exit-code table already gives 2 for a
      stray argument.


marrgko:
One behaviour change in here is worth a reviewer's eye, because it is a
side effect of the fix rather than something #164 asks for.

`brig version --help` used to print the version and exit 0. It now refuses:

```
$ brig version --help
brig: unexpected argument "--help"; `brig version` takes no arguments
$ echo $?
2
```

Same for `-h` and `help`. Nothing in #164 asks for that — the report only
raises `--help` against `brig policy ls`, where the complaint is that the
reader got output that was not help. It follows from `version` now reading
its tail at all, where before it read nothing.

I left it refusing, because that is what its siblings do:

| command | exit | answer |
| --- | --- | --- |
| `brig ls --help` | 2 | `unexpected argument "--help"` |
| `brig doctor --help` | 2 | `unknown flag "--help" for brig doctor` |
| `brig logs --help` | 2 | `unknown flag "--help" for brig logs` |
| `brig version --help` | 2 | `unexpected argument "--help"` (this PR) |

The verb groups go the other way, and this PR keeps them there:
`brig policy ls --help` now prints the group's usage and exits 0, which is
the `flag.ErrHelp` translation `policyCmd` already made for every other
verb in the group.

So the split this PR leaves is: a top-level verb refuses `--help`, a verb
inside a group answers it. That split already existed; `version` simply
stopped being the one verb that silently ignored the question.

Two ways to settle it if the split is not what we want:

1. `brig version --help` prints the top-level usage and exits 0. Consistent
   with the rest of this PR, but then `version` is the only top-level verb
   that answers `--help`, and `brig ls --help` is the odd one instead.
2. Leave it, and treat "top-level verbs refuse `--help`" as the documented
   rule — which would mean saying so in `docs/cli.md`, where it is not
   written down today.

Happy either way; flagging it so it is a decision rather than a side effect.

cmd/brig/policy.go
```diff
@@ -44,6 +44,35 @@ attach binds it to a profile's every run, or -- with -n -- to one session;
 a profile's own inline policy: field does the same without a separate attach.
 `
 
+// rejectPolicyTail refuses a word the policy listing has no place for, rather
```
asapranidis: Proportion and duplication. This comment is nineteen lines on an eight-line function, and most of it is written out again: the "printed the listing and exited 0" story appears here, in the comment on `TestPolicyListingAndVersionRefuseAWordTooMany`, in the comment on `TestPolicyListingAnswersHelpLikeTheRestOfTheGroup`, and in the PR body. Four copies to keep in step.

The two sentences that carry weight are why a help token comes back as `flag.ErrHelp` (the group's translation answers it) and why the check sits at the callers rather than inside `listPolicies` (it stays a printer). Keep those two, and let the tests hold the history.

cmd/brig/policy.go
```diff
@@ -54,13 +83,20 @@ func policyCmd(args []string) error {
 	case "--help", "-h", "help":
 		fmt.Print(policyUsage)
 		return nil
+	// The assignment falls through to the flag.ErrHelp translation below
```
asapranidis: Placement. This comment sits after the `--help` case body and before `case "ls"`, so on first read it looks like a note on the help case, which returns rather than falls through. Either move it inside `case "ls"` next to the `rejectPolicyTail` call, or drop it: the comment on the `errors.Is(err, flag.ErrHelp)` block at the foot of the function already says the same thing for the whole group.

cmd/brig/main.go
```diff
@@ -286,6 +286,12 @@ func dispatch(args []string) error {
 		fmt.Print(usage)
 		return nil
 	case "version", "--version":
+		// Named as the reader spelled it. Both spellings are current -- one is
```
asapranidis, line 289: On the behaviour change you flagged in the PR comment: I checked the siblings on this branch and agree with leaving `brig version --help` refusing. `ls`, `doctor` and `logs` all exit 2 on `--help` today, so `version` joins them rather than becoming the odd one out. Writing "top-level verbs refuse `--help`" into `docs/cli.md` belongs with #165, which already collects this class of inconsistency, not in a fix scoped to #164.

cmd/brig/trailing_test.go
```diff
@@ -152,3 +152,180 @@ func TestAgentAcceptsTheRightNumberOfArguments(t *testing.T) {
 		}
 	}
 }
+
+// The policy listing and brig version read their verb and stopped there, so a
+// trailing word was dropped and the command still exited 0 -- and a reader who
+// typed `brig policy ls --help` got the listing, not help, with a status that
+// said it worked.
+//
+// The table is over every spelling that reaches listPolicies, not just the two
```
asapranidis, line 161: Nit, and about direction rather than this PR: the new comments describe the issue ("not just the two in the report", "the other half of the report's complaint") rather than the behaviour under test. The file already does this in the agent-group tests above, so this matches the local style. The repo's preference elsewhere is for comments that state what must hold and why, with the issue number in the commit trailer. Worth knowing for the next one; no need to churn this PR over it.

Also, "a retired word that quietly drops the tail is still a word that drops the tail" is a second copy of the sentence at line 27.

asapranidis (commented):
Reviewed at 9ca9a19. Read the whole diff, checked each claim in the description against the source, ran the gates in a worktree, and proved the new tests fail without the fix.

**What holds**

- Every spelling that reaches `listPolicies` is covered: `policy ls`, the retired `policy list`, and the top-level `policies`, which is now routed through `policyCmd` the way `deprecatedProfileCmd` routes onto `agentCmd`.
- The `flag.ErrHelp` translation at the foot of `policyCmd` is real and prints `policyUsage` with exit 0, so returning `ErrHelp` from `ls` lands in the same place as `show --help`.
- Class sweep on the branch: `secret ls`, `telemetry`, `completion`, `doctor`, `agent ls` and `ls` all refuse a trailing word already. Only `brig help extra` still ignores its tail, which is defensible for `help`. These two were the last listing verbs with the hole.
- The tests bite. Disabling the tail check fails four tests. Refusing the help token instead of returning `ErrHelp` fails the help test. Removing the `version` check fails both version rows.

**Gates**: gofmt clean, `go vet` clean, `go test -race ./...` all ok, build ok, `script/smoke.sh` exit 0 with 327 checks.

**On `brig version --help` exiting 2**: I agree with leaving it. The split between top-level verbs and grouped verbs predates this PR, and `version` was the one verb hiding it. Writing the rule into `docs/cli.md` belongs with #165, which is already about one mistake getting different answers, rather than in a fix for #164.

**Findings**, none blocking, all inline:

1. The `rejectPolicyTail` doc comment is three paragraphs for an eight-line function, and the same explanation is written out again in two test comments and the PR body.
2. The comment above `case "ls"` in `policyCmd` sits after the `--help` case body, so it reads as attached to the wrong case.
3. Test comments describe the issue's history rather than the behaviour under test. That matches the existing style in the file, so it is a nit about direction, not about this PR alone.

Verdict: correct and complete. Happy to approve once the comments are considered, or as it stands if the author prefers.

pmoust:
@claude-fleet review

asapranidis (approved):

PR #215, opened 2026-09-12
fix(secret): say when a profile has nothing to import

### LINKED ISSUE #178: brig secret import says "importing 0 secrets" and does not say why
## Describe the bug

On a profile that declares no secrets, `brig secret import` reports a count and stops. It never says the profile has nothing to import. Six of the eight shipped profiles are in this state.

## To reproduce

Reproduced on 0.1.0-rc18:

```
$ brig secret import codex
codex: importing 0 secrets
$ echo $?
0
```

`codex` declares no secrets, so there is nothing this command could ever do for it.

## Expected behavior

The message the named form already has, at `secretimport.go:346-347`:

```
%s declares no secrets, so there is nothing to import for %q
```

## Actual behavior

A count of zero, which reads like an import that found nothing rather than a profile with nothing to find.

## Cause

`secretimport.go:114` with `:126-133` and `:437-440` take the count path for the unnamed form. The named form takes the explanatory one.

## Second fault in the same path

`brig secret import` opens the secret store before it checks whether there is anything to import. `secretimport.go:91-99` runs `selectForImport`, then calls `openStore()` unconditionally.

On a host with no keyring, `secret.Open` returns `ErrUnsupported` (`internal/secret/secret.go:24`, `secretservice_linux.go`). So a guaranteed no-op fails there instead of reporting that there was nothing to do.

Moving the open below an `if len(selected) == 0` early return fixes both the ordering and the message.

## Environment

brig 0.1.0-rc18 (d11ba32), macOS 26.5 arm64. The keyring half is read from source: this host has a working keychain, so the `ErrUnsupported` path was not exercised.

pmoust:
## Summary

`brig secret import codex` printed `codex: importing 0 secrets` and exited 0. Six of eight shipped profiles declare no secrets, so this was the normal output for most agents. It also opened the secret store before checking whether there was anything to store, so on a host with no keyring the no-op failed with `no secret store on this platform`.

Now the unnamed form returns before the store opens when nothing was selected. A profile with no secrets prints the sentence the named form already used. A profile whose secrets all lack a source keeps its per-secret notes and drops the count.

## Related issues

Fixes #178

## Changes

- `cmd/brig/secretimport.go`: early return when nothing is selected; hand-created notes moved into `reportHandCreated`
- `cmd/brig/secretimport_test.go`: two tests, one for the message on both forms including `--dry-run`, one that the store is not opened
- `docs/authentication.md`: quotes the new message, drops the no-keyring failure

## Checklist

- ~~I have run `go test ./... -race`~~ no concurrency, subprocess or daemon change
- ~~I have exercised the change against a real runtime~~ import path only, no boot
- [x] I have updated the affected docs

Touches the import path in `cmd/brig` only. No change to what a run reads or forwards. The store is opened later than before, never for a different set of names.

### DIFF
--- a/cmd/brig/secretimport.go
+++ b/cmd/brig/secretimport.go
@@ -92,6 +92,21 @@ func importSecrets(out io.Writer, args []string) error {
 	if err != nil {
 		return err
 	}
+	// Nothing to fill means nothing to read and nothing to store, so say so
+	// and stop before the store is opened: on a host with no keyring, opening
+	// it is what would turn a guaranteed no-op into a failure. A count of zero
+	// reads like an import that found nothing, which is a different thing
+	// from a profile with nothing to find. selectForImport has already made
+	// the named form an error, so this is the unnamed form only.
+	if len(selected) == 0 {
+		if len(p.Secrets) == 0 {
+			fmt.Fprintf(out, "%s declares no secrets, so there is nothing to import\n", p.Name)
+			return nil
+		}
+		fmt.Fprintf(out, "%s declares no secret brig can read from your host\n", p.Name)
+		reportHandCreated(out, p)
+		return nil
+	}
 
 	store, err := openStore()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pmoust:
@claude-fleet review

asapranidis (approved):

PR #228, opened 2026-09-15
feat(publish): Open a guest port on the host

### LINKED ISSUE #18: Publish guest ports to the host
### Problem

An agent that starts a dev server has nowhere to serve it. There is no port
flag and no forwarding, so the usual loop of "run the app, look at it in a
browser" stops at the sandbox edge.

### Proposed solution

Add explicit publishing on `run` and `create`:

```
brig run claude . -p 3000
brig run claude . -p 8080:80
```

Bind to loopback on the host by default. On `hvi`, the forward belongs in the
user-mode gateway brig already owns; on Linux, in the runtime's own
publishing.

Report every published port in the envelope from #10 and in `--json`. A port
is a hole in the boundary, and an unreported hole is the wrong kind.

This lands after the network work in #15. Publishing a port from a sandbox that
shares a network with every other sandbox exports more than the user asked
for.

### Alternatives considered

Publish nothing and tell users to run the server on the host. It keeps the
boundary simple, and it removes a large part of what a coding agent is for.

It widens the boundary, on request and per port. That is why the default is
loopback, and why every port appears in the envelope.

ananos:
## Summary

An agent that starts a dev server had nowhere to serve it. There was no port
flag and no forwarding, so the usual loop of "run the app, look at it in a
browser" stopped at the sandbox edge.

There are two ways to ask for a port now. `--publish` on a run or a create, and
`brig publish` on a sandbox that is already up -- the second matters, because
the interesting case is an agent that has just started a server in a session
you do not want to restart. `brig unpublish` closes one again, and `brig
publish` with no port lists what is open.

The mechanism is the gateway's. brig passes `--api` on every gateway it starts
and reconciles the sandbox's ports against `/forwards`, which brig-sh/hull#74
adds. **This does not work without that PR**, so it should land first.

The other option was `--forward` at gateway startup, which hull has always
taken. It is not usable here: the shared gateway serves every sandbox on the
shared network, so restarting it to add one sandbox's port would drop the
network of every other.

## Related issues

Closes #18. Depends on brig-sh/hull#74.

## Changes

- `--publish PORT` on the run line, repeatable. `brig publish <ref> [PORT...]`
  and `brig unpublish <ref> PORT...|--all` as verbs of their own.
- The grammar is docker's: `3000`, `8080:80`, `127.0.0.1:8080:80`,
  `5353:53/udp`. This is the one part of brig a person arrives already knowing.
- A `PORTS` row in the execution envelope, and `kind: Ports` under `--json`.
- Bound to `127.0.0.1` unless the address says otherwise. `0.0.0.0` says
  "reachable from the network this host is on" on its own envelope row.
- A publication belongs to the sandbox, not to one run. It is recorded beside
  the two address maps in the gateway directory, survives a stop and a
  recreate, and goes with `brig rm`.
- `brig stop` releases the host port and keeps the record, so the next run
  publishes it again without being asked.
- Refused where it cannot be honoured: `vz` takes its network from vmnet, so
  `--publish` there is refused by name with the setting that caused it. On
  Linux nerdctl publishes at creation, which is the only time a container can.
- `docs/cli.md`, `docs/security.md` and the README.

## Checklist

On the fifth box, since that is the one this change most needed. On an hvi
microVM with the hull branch: booted with `--publish 18080:8000`, started a
`python3 -m http.server` in the guest, and fetched the page from the host over
the published port. Then published a second port onto the running sandbox and
fetched that too, withdrew it and confirmed the connection refused while the
first still served. Then stopped the sandbox, confirmed the host port released,
and ran again with no flags to confirm it came back.

`script/smoke.sh` earned its place here. It caught a real bug: the boot
reconciled unconditionally, so a gateway with no API socket at all -- a stub, or
any gateway brig did not start -- failed the run even when the sandbox
published nothing. The guard only covered a 404, not a failure to connect. It
fails only a run that actually wants a port now, and there is a test for it.

This touches the boundary, in the widening direction, and on purpose. It is the
first inbound hole brig has ever opened.

Three things hold it down. It binds to loopback unless you write an address
yourself, so a published port is reachable from this machine and not from the
network this machine is on. Every port is named in the envelope on every run,
including ones this command line did not ask for -- a publication outlives the
run that made it, and a hole nobody was told about is the wrong kind. And an
address other than loopback says so on its own row.

Worth being explicit about one thing: a published port is ingress, so no egress
rule applies to it. A policy decides what the guest may open a connection *to*
and has nothing to say about a connection opened *into* the guest. A sandbox
under a strict egress policy with a port published is still reachable on that
port. That is documented in `docs/security.md` rather than left to be
discovered.

Nothing here touches credential resolution or delivery.

## Two things to argue about

**There is no `-p`.** I know the issue asks for one, and I know it is what
everyone's fingers will type. It is also Claude Code's print flag, so `brig run
claude -p "fix the tests"` would mean a port to brig and a prompt to the agent.
We already have a hygiene test asserting that line stays the agent's, and the
same reasoning is why `--verbose` has no `-v`. So the run line takes
`--publish`, and `brig publish` takes bare words, where there is no agent to
disagree. If we want `-p` anyway, it is a deliberate break and we should take it
knowingly rather than by convenience.

**Should these be a `brig network` group instead?** `publish` and `unpublish`
are flat verbs here, partly because there was no group to put them in. We have
`agent`, `policy`, `secret` and `telemetry`, and network concerns are currently
spread across five places: `--network`, `brig policy`, `logs --gateway`, the
pruning half of `reset`, and now these. A `brig network` group would give them
one home -- `brig network publish`, `brig network ls`, and so on.

I went with the flat verbs because `brig publish claude@demo 3000` is shorter
for the common case and matches docker muscle memory. The cost is real though:
`network` is already a flag meaning "posture", so `brig network` as a verb and
`--network` as a flag would be two subjects sharing one word, which is the kind
of collision we usually refuse.

I do not have a strong preference. This is cheap to reshape now and expensive
after a release, so let us settle it here. @pmoust @asapranidis, what do you
think?


ananos:
@claude-fleet review

ananos:
Both findings were real, thanks -- good catches, and the first one is embarrassing.

**1. Error messages naming `-p`.** Fixed. Those refusals were written before the decision to drop `-p`, and nothing caught them because no test reads the message. They now say `--publish` on the run line, and the conflict examples give the bare port form (`8081:90`) which is what `brig publish` takes. You were right that the old ones were copy-paste-broken: pasting `-p 8081:80` would have run the agent with it and published nothing.

**2. The record running ahead of the gateway.** Fixed, and it was two problems in one.

`recordPublications()` ran before the running-branch, so a live publish that failed left the record naming a port the gateway never had. It now runs on the path that boots; the running path records inside `hull.Publish`, after the gateway has the forward.

And the live path did not implement the move the record and the boot reconcile both do. `hull.Publish` now withdraws before it publishes -- but only a forward belonging to this guest, checked through `publishedOn`, because the gateway keys a forward by protocol and local address alone and withdrawing blind would take another sandbox's port off the shared gateway.

Two tests, both of which fail on the previous code: `TestPublishMovesAHostPortOnALiveSandbox` reproduces your scenario exactly (the old error read `already published by brig-claude-web` -- the sandbox conflicting with itself), and `TestPublishWillNotTakeAnotherSandboxsPort` covers the hazard the fix introduces if done carelessly.

On what you could not check: the gateway wire contract is brig-sh/hull#74, and the two were exercised together against a real gateway on an hvi microVM -- a server in the guest fetched from the host over the published port, a second port published onto the running sandbox, withdrawn again, and the port released and restored across a stop and a run. The 409 body your `conflictGuest` note asks about is the one that hull's endpoint emits; there is a test on each side of it.

The two commits are now one, as requested.

internal/runtime/hull.go
```diff
@@ -749,9 +860,32 @@ func releaseGateway(name string, err error) {
...
+	// again. Best effort, like the rest of this: a forward left behind costs a
+	// port, and there is no answer to "the withdrawal failed" worth
+	// interrupting a stop with.
+	withdrawPublications(name)
 	shutDownGateway(name)
 }
 
+// withdrawPublications takes this sandbox's forwards off the gateway serving
+// it, leaving what the sandbox publishes recorded.
+func withdrawPublications(name string) {
```
asapranidis, line 899: `withdrawPublications` withdraws every entry in the record without checking which guest the forward carries to. `unexposeForward` sends only `protocol` and `local` (gatewayapi.go:136), so the DELETE matches on host address alone. `hull.Unpublish` (hull.go:716) has the same gap; `hull.Publish` guards it with `publishedOn`.

Sequence, both sandboxes on the shared network:

1. `brig run A --publish 8080:3000`, then `brig stop A` — forward withdrawn, record kept by design.
2. `brig publish B 8080` — the host port is free, so it succeeds.
3. `brig rm A`, or `brig run A`: `EnsureRunning` calls `Runtime.Remove` to clear the stale stopped instance (wrap/run.go:218).

Step 3 reaches `hull.Remove` → `releaseGateway` → `withdrawPublications("A")`. A's isolated socket is unreachable and its shared IP record still exists (`releaseGatewayIP` runs after), so `sandboxGateway` returns the shared gateway and `DELETE /forwards?protocol=tcp&local=127.0.0.1:8080` removes B's forward. B loses the port with nothing logged.

Intersect the record with `publishedOn(sock, guestIP)` and withdraw only the forwards that carry to this guest.

internal/runtime/gateway.go
```diff
@@ -317,6 +317,11 @@ func startGateway(bin, sock, subnet, gatewayIP string, policy Egress, spec strin
 	args := []string{"network-gateway",
 		"--socket", sock,
 		"--qemu-socket", qemuGatewaySocket(sock),
+		// The API socket is opened on every gateway, whether or not this run
+		// publishes anything: a port published later has to reach a gateway
+		// that is already serving guests, and a gateway started without this
+		// could only be given one by being restarted. See gatewayapi.go.
+		"--api", gatewayAPISocket(sock),
```
asapranidis, line 324: `--api` is passed to every gateway start with no capability probe. `gatewayEnforces` (gateway.go:381) performs that check for the policy flags.

Against a hull without brig-sh/hull#74, hull rejects the unknown flag and exits. `startGateway` ignores `cmd.Start` and the exit status and polls only for reachability, so every hvi run with a network — including runs publishing nothing — fails after `gatewayReadyTimeout` with "the network gateway did not come up at ... within 30s".

The ENOENT/ECONNREFUSED guard in `reconcilePublications` does not cover this: the gateway never starts, so no boot reaches the guard.

Probe `network-gateway --help` for `--api` and omit the flag when it is absent. That limits the failure to runs that request a port.

cmd/brig/main.go
```diff
@@ -3226,3 +3323,165 @@ func warnf(format string, a ...any) {
...
+	}
+	if len(add) == 0 {
+		return reportPorts(cfg, ref, wantJSON)
+	}
+	running, err := cfg.Runtime.Running(cfg.VMName)
+	if err != nil {
+		return fmt.Errorf("cannot tell whether the sandbox %s is running, so brig cannot "+
+			"publish a port from it: %w", cfg.VMName, err)
+	}
+	if !running {
```
asapranidis, line 3354: `brig publish` records unconditionally with no backend check. The default hypervisor is `vz` (`hypervisorOrDefault`, hull.go:165), and `supports()` refuses any run carrying a publication on a non-hvi backend.

1. `brig publish claude@web 3000` on a stopped sandbox, default settings — recorded, prints "These ports are published when it starts."
2. `brig run claude@web`, no `--publish` on the line. `Load` reads the record into `Publish` (wrap/config.go:397), and `supports()` refuses with "run it on hvi, or drop the --publish flags" (hull.go:403).

There are no flags on the line to drop, and the message does not name `brig unpublish`. The sandbox stays refused until the user finds `published.json`.

`sandboxGateway` already rejects vz by name on the running path. Apply the same refusal before recording, or name `brig unpublish` in the `supports()` message.

internal/runtime/gatewayapi.go
```diff
@@ -0,0 +1,373 @@
...
+	Remote   string `json:"remote"`
+}
+
+// gatewayAPI calls one endpoint on a gateway's API socket.
+//
+// The host in the URL is a placeholder: the dialer ignores it and connects to
+// the socket. Every call goes through here so the timeout and the socket
+// plumbing are written once.
+func gatewayAPI(sock, method, path string, body any) ([]byte, int, error) {
+	client := &http.Client{
```
asapranidis: Each call builds a fresh `http.Transport` and discards it without `CloseIdleConnections`. `brigd` is long-lived and drives `cfg.EnsureRunning` (brigd/main.go:669) → `hull.Run` → `reconcilePublications` → here, at least once per sandbox start.

The body is read and closed, so the connection returns to this transport's idle pool. `IdleConnTimeout` is zero (no limit) and the `persistConn` goroutines keep the transport reachable, so it is never collected: one fd and two goroutines per run, unbounded over the daemon's life.

Use a package-level client, or `defer client.CloseIdleConnections()`.

cmd/brig/main.go
```diff
@@ -3226,3 +3323,165 @@ func warnf(format string, a ...any) {
...
+// and the guest port is not asked for.
+func unpublishPorts(cfg *wrap.Config, ref string, ports []string, all, wantJSON bool) error {
+	if all && len(ports) > 0 {
+		return usagef("--all withdraws every port, so naming %q as well asks for two "+
+			"different things; drop one", ports[0])
+	}
+	var drop []runtime.Publication
+	switch {
+	case all:
+		drop = runtime.Publications(cfg.VMName)
```
asapranidis: `ParsePublications("8080")` leaves `HostAddr` empty, so `Addr()` returns `127.0.0.1` and `Publication.Same` compares that address. A `0.0.0.0:8080:80` record never matches: nothing is dropped, no error is returned, and `reportPorts` prints the port still open.

Copying the printed `HOST` column back also fails. `brig unpublish ref 0.0.0.0:8080` splits into two colon-separated parts, so `0.0.0.0` is read as the host port and refused. The user has to supply the full `ADDR:HOST:GUEST` form, guest port included.

Match on host port alone when no address was typed, or report the ports that were named but not found.

cmd/brig/main.go
```diff
@@ -3226,3 +3323,165 @@ func warnf(format string, a ...any) {
...
+// moment. A port in the record that the gateway does not have is a sandbox
+// that is not running, and the row says so rather than claiming a host port
+// that nothing answers on.
+func reportPorts(cfg *wrap.Config, ref string, wantJSON bool) error {
+	want := runtime.Publications(cfg.VMName)
+	live := map[string]bool{}
+	if publisher, ok := cfg.Runtime.(runtime.Publisher); ok {
+		if have, err := publisher.Published(cfg.VMName); err == nil {
+			for _, p := range have {
+				live[p.Local()] = true
```
asapranidis: `live` is keyed by `Local()`, which drops the protocol, so `5353:53/udp` and `5353:53` share the key `127.0.0.1:5353`. With only the udp forward installed, the tcp row prints `STATE open` — a host port reported as listening when nothing answers on it, which is the misreport the pending/open distinction exists to prevent.

Key the map by `Proto()` and `Local()`.

internal/runtime/published.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	return filepath.Join(dir, "published.json"), nil
+}
+
+// readPublished is every sandbox's publications, keyed by sandbox name.
+//
+// A file that cannot be read is an empty map. The consequence is a sandbox
+// that comes up publishing nothing, which the envelope then reports honestly;
+// the consequence of failing instead is a boot refused over bookkeeping.
+func readPublished(path string) map[string][]Publication {
```
asapranidis: Any read or unmarshal failure yields `{}`, and `RecordPublications` writes that map back. A transient EACCES/EIO, or a truncated file, during `brig publish A 3000` leaves a file containing only A.

B's next boot reconciles against an empty want set, and `reconcilePublications` withdraws B's live forwards. One failed read closes ports on unrelated sandboxes — wider than the "comes up publishing nothing" the comment states.

Distinguish `os.IsNotExist` (empty map is correct) from any other error, and fail the write rather than rewrite from a guess.

ananos:
Thanks @asapranidis -- this is a very good review. Six of the seven were real and are fixed; one I think rests on a premise that does not hold, evidence below.

**1. `withdrawPublications` taking another sandbox's port.** Fixed, and you traced it exactly. Both `withdrawPublications` and `hull.Unpublish` now go through `publishedOn(sock, guestIP)` and withdraw only forwards that carry to this guest, the way `hull.Publish` already did. `withdrawPublications` no longer reads the record at all -- it withdraws what the gateway says this guest holds, which is the only set that can be correct.

Two tests, both failing on the old code: `TestStoppingASandboxLeavesAPortAnotherSandboxTook` is your three-step sequence, and `TestUnpublishLeavesAPortAnotherSandboxTook` covers the explicit path.

**3. `brig publish` recording with no backend check.** Fixed both ways you suggested. `brig publish` now calls `Config.CanPublish` before recording, so the refusal arrives while the port is still on the command line. And the `supports()` message no longer says "drop the --publish flags" -- it names the publication, says a publication outlives the run that made it, and points at `brig unpublish`.

**4. Transport leak.** Fixed: `defer transport.CloseIdleConnections()`. You are right that brigd is the case that matters -- it reconciles on every sandbox start.

**5. `unpublish` missing a 0.0.0.0 port silently.** Fixed. A port written without an address now matches any address on that port, which is also what makes the printed HOST column copy back. Naming a port that is not published is an error rather than silence: `claude@t publishes nothing on 9999. `brig publish claude@t` lists what it does`.

**6. `live` keyed by `Local()`.** Fixed, keyed by `Proto()+"/"+Local()`. Verified: 5353 tcp and 5353 udp now print as separate rows with their own STATE.

**7. Any read failure yielding `{}`.** Fixed, and this was the worst of them. `readPublished` distinguishes `os.ErrNotExist` (an empty map is correct) from every other failure, which is now an error. `Publications` returns that error, `Load` surfaces it, and the write refuses rather than persisting a guess. `TestAnUnreadableRecordIsAnErrorRatherThanNoPublications` asserts the corrupt file is still on disk afterwards. The boot also stopped re-reading the record -- it installs `spec.Publish`, the set the caller already resolved.

---

**2. `--api` with no capability probe.** I do not think this one holds, and I would rather show you than argue it. The premise is that a hull without brig-sh/hull#74 rejects `--api` and exits. `--api` is not new in that PR -- it has been in `network-gateway` since hull's first commit, and every released hull has it:

```
$ hull --version
hull version 0.1.0-rc28
$ hull network-gateway --help | grep -- --api
   --api string     HTTP API socket path (probe endpoint)
```

What #74 adds is the `/forwards` endpoint on that socket, not the flag. So the skew is an older hull that opens the socket and serves no `/forwards`, which is exactly what the guard in `reconcilePublications` covers: a run publishing nothing returns nil whatever the failure, and one publishing a port gets "Upgrade the runtime".

I did write the probe before checking, and then took it out. If you know of a hull older than rc21 in the wild I will put it back, but I would rather not spend a subprocess per gateway start on a case I cannot produce.

All gates green on `def9483`: `make all`, `script/smoke.sh`, `go test ./... -race`, `check-retired-spellings.sh`.

pmoust:
`brig network` makes sense as a group. It can be in another PR (before next release)

asapranidis (approved):
we can deal with the `brig network` as a follow up work

ananos:
Filed the `brig network` follow-up as #312, so this one can go in as it stands. Thanks both -- @pmoust for the group and @asapranidis for the review.

The argument in there for doing it before the next release is that it is free now: `brig publish` has never shipped, so there is no spelling we owe a deprecation window to.

PR #231, opened 2026-09-16
feat(cli): report the build behind the binary in brig version

### LINKED ISSUE #230: Report the build behind the binary in `brig version`
## Problem

`brig version` prints one token: the value stamped into `main.version` at link time (`cmd/brig/main.go:32`, printed at `cmd/brig/main.go:297`). Three build paths stamp three different things, and none of them identifies the build:

| Build path | Stamps | Prints |
|---|---|---|
| goreleaser release (`.goreleaser.yaml:29`) | the tag, without `v` | `brig 0.2.0` |
| `make` (`Makefile:3-4`) | `git describe --tags --always --dirty` | `brig v0.2.0-6-gef4aa8b` |
| `go build` or `go install`, no flags | nothing | `brig dev` |

The binary already carries what is missing. Go embeds build info in every binary built from a module in a git checkout: the module version, the commit, the commit time, whether the tree was modified, the Go version, and the target platform. `brig version` reads none of it. `go version -m` on the plain `go build` binary that printed `brig dev`:

```
mod   github.com/brig-sh/brig  v0.2.1-0.20260915210404-ef4aa8b0efb6
build GOOS=darwin
build GOARCH=arm64
build vcs.revision=ef4aa8b0efb64c7eb76f73a6cdd18d7aedbe7c79
build vcs.time=2026-09-15T21:04:04Z
build vcs.modified=false
```

The same fields in the Homebrew `0.2.0` binary:

```
mod   github.com/brig-sh/brig  v0.2.0
build vcs.revision=131e3bc5615df5ff74e6b5af9a5bcf2ed42b1d57
build vcs.time=2026-09-15T09:36:19Z
build vcs.modified=false
```

Why this matters:

- **Release candidates.** The rc series ships many builds per version. Two rc binaries built from different commits under the same tag, or a tag rebuilt after a force-push, print the same line. A report against `0.3.0-rc2` does not say which `0.3.0-rc2`.
- **Modified trees.** A binary built from a tree with uncommitted changes prints the same version as a clean one. Only the `make` path marks a dirty tree, and only inside the version token.
- **Bug reports.** `CONTRIBUTING.md:140-142` asks reporters for `brig version`. The output is not enough to find the source, so reporters look the commit up by hand. #196 wrote its environment as `brig 0.1.0-rc18 (d11ba32)` for that reason.
- **Unstamped builds.** `go install github.com/brig-sh/brig/cmd/brig@latest` prints `brig dev`, although the build info holds the pseudo-version and the commit.
- **brigd.** The daemon has the same stamped variable (`cmd/brigd/main.go:40`) and returns it over its protocol (`cmd/brigd/main.go:568`). Whatever reads the daemon's version has the same gap.

CI asserts that the VERSION file and the release tag agree (`.github/workflows/release.yml:39-45`). So the version token says what the tag says, which is what the file says. It does not say what was built.

## Proposed solution

Open for discussion. A solution has to meet these constraints:

- `brig version` identifies the build: the commit, the commit time, and whether the tree was modified, next to the release version.
- An unstamped `go build` or `go install` binary reports more than `dev`.
- The first token of the output stays the release version, so a reader who takes the last field of the line still gets a version.
- brigd reports the same fields.

## Additional context

Go 1.24 and later derive the main module version from the git tag, which is the `mod` line in the output above. That is one more source of the release version than the linker flag, and the two can disagree.

asapranidis:
## Summary

`brig version` printed one token, stamped into `main.version` at link time:
the tag under goreleaser, `git describe` under make, and `dev` for a plain
`go build` or `go install`. None of those named the commit, so two binaries
from one tag printed the same line, a modified tree looked like a clean one,
and bug reports asked for the commit by hand.

Read the build from the binary instead. Go 1.24 and later derive the main
module version from the nearest reachable `v*` tag and embed the commit, its
time and the modified flag beside it. Both binaries now report that, the
stamp is gone, and with it the VERSION file: the tag is the version.

```
$ brig version
brig v0.2.0 (131e3bc, 2026-09-15, go1.26.0, darwin/arm64)

$ brig version            # a commit after v0.2.0
brig v0.2.1-0.20260916082002-314cd3a8fed0 (314cd3a, 2026-09-16, go1.26.0, darwin/arm64)
```

Output changes once, from `brig 0.2.0` to `brig v0.2.0`. `docs/cli.md`
already documented the `v` form. The first two words stay `brig <version>`,
so a reader taking the second field still gets a version.

Each commit stands alone: it builds, vets and passes its tests without the
ones after it. Reviewing in order is the intended path.

## Related issues

Closes #230

## Changes

- `internal/buildinfo`: reads version, commit, commit time, modified flag, Go
  version and platform from `debug.ReadBuildInfo`, with `dev` and no commit
  as the fallback for a build without git history.
- `brig version` prints that build on one line. `--json` prints it under the
  standard envelope, kind `Version`, in either flag position.
- brigd's `version` op returns `commit`, `commitTime` and `modified` beside
  `version`; added fields within protocol version 1.
- `brig doctor` opens with a `brig` row and asks a running brigd its build. A
  daemon from a different build is `!!` with a restart as the fix. This is
  the first request brig sends to brigd, and the op is read-only.
- Makefile and `.goreleaser.yaml` drop `-X main.version`. The goreleaser
  comment records the two conditions the embedded version depends on: a full
  clone, and a `go mod tidy` hook that leaves the tree clean.
- `VERSION` and its tag-matches check in the release workflow are deleted.
  `docs/releasing.md` describes the tag-only flow: release, release
  candidate, maintenance branch, and what a build between tags prints.

## Checklist

- [x] `script/smoke.sh` passes (340 ok, 0 failed)
- [ ] ~~I have exercised the change against a real runtime (`brig run <agent>`), if it touches the run, exec or credential path~~ -- nothing in the run, exec or credential path changed

The change was written with an AI assistant; the author reviewed every line.


asapranidis:
@claude-fleet review

pmoust (approved):
Makes sense

pmoust:
Just curious, I am surprised goreleaser doesn't have that functionality OOTB. Did we check whether that flow is supported?
Seems like such a common need.

ananos (commented):
Thanks @asapranidis, this is a nice cleanup. Dropping the `VERSION` file and letting the toolchain answer the question is the right call, and the `docs/releasing.md` rewrite is useful on its own -- the pseudo-version table and the maintenance-branch diagram are both correct against Go's actual rules.

I went through the 7 commits in order and checked the behaviour end to end. Everything the PR body claims holds:

- `go build ./...`, `go vet ./...` and `go test ./...` are green, and each of the 7 commits builds, vets and tests standalone.
- `script/smoke.sh` passes (exit 0), including the new doctor assertion.
- A tagged build prints the bare tag, a build after a tag prints the pseudo-version, and a dirty tree appends `+dirty` and sets `"modified": true`.
- brigd's `version` op answers with the new fields, and doctor marks a mismatched daemon `!!` with the restart fix.

The ordering in `brigdCheck` is good: the 0600 mode check returns before anything dials, so a socket with the wrong mode is never spoken to.

One thing I would fix before merge, plus a few smaller ones.

## A build from a git worktree loses the version and the commit

Same commit, same clone, same toolchain. Only the checkout kind differs:

| built in | main (before) | this PR |
| --- | --- | --- |
| normal clone | `brig v0.2.0-6-gef4aa8b` | `brig v0.2.1-0.20260916082117-13e9bb7b33c8 (13e9bb7, ...)` |
| linked worktree | `brig v0.2.0-6-gef4aa8b` | `brig dev (go1.26.5, darwin/arm64)` |

The reason is in the toolchain. `vcsGit` declares its root as `{filename: ".git", isDir: true}`, and a linked worktree's `.git` is a file. `isVCSRoot` never matches, no repo root is found, and nothing is stamped. It is silent even under `-buildvcs=true`, since there is no repo to complain about.

So in a worktree we land exactly the state this PR sets out to remove: a binary that cannot name its commit. The old `git describe` line handled worktrees fine.

The release path is not affected. `actions/checkout` gives a normal clone, and I confirmed a tagged build prints its tag. This is about the dev workflow.

A line in `docs/releasing.md` would cover it. Another approach would be a Makefile fallback that stamps only when `buildinfo` comes back empty, which keeps the release binaries pure and gives worktree builds their commit back. What do you think?

## The release checklist will not catch a `+dirty` build

The comment in `.goreleaser.yaml` names the risk correctly: the `go mod tidy` hook has to leave the tree clean. goreleaser validates dirtiness before it runs that hook though, so a tidy that touches `go.sum` produces `+dirty` binaries with nothing failing.

The check in `docs/releasing.md` says to look for a pseudo-version. `v0.3.0+dirty` is not a pseudo-version, so it passes. Adding "and no `+dirty`" to that line would close it, and a `git diff --exit-code` after the tidy hook would too.

`.gitignore` already covers `/brig`, `/brigd` and `/dist/`, so repeat `make build` and goreleaser's own output stay clean.

## Nits

- `modified` is always present in `brig version --json` and `omitempty` in brigd's response, so it is absent there when false. `docs/brigd.md` documents this, and the PR body calls them the same fields. Worth making the two agree, or saying why they differ.
- `cmd/brig/main.go` has one 85-column line in the usage block (`after the verb (brig ls --json)...`). The rest of that block is at 80.
- `theirs.Commit != mine.Commit` flags a mismatch when one binary came from `go install` (no VCS data) and the other from a release archive, even at the same version, and the restart it suggests will not help. Comparing commits only when both are non-empty would cover it. Edge case, since the two normally ship together. Up to you.

## On the commits

When you address these, could you fold each fix into the commit it belongs to and force-push, rather than adding fix commits on top? The 7 commits read well in order and it would be nice to keep them that way through merge.

ananos:
@pmoust good question. It does, and more than we were using -- with no `ldflags` block at all, goreleaser stamps four vars by default. On a scratch module tagged `v1.2.3`:

```
version=1.2.3 commit=5411d18f5e2a791534b7b3fb39fcc403e85d4ba6 date=2026-09-16T13:21:50Z builtBy=goreleaser
```

Our `-s -w -X main.version={{.Version}}` was overriding that default, so we were throwing commit and date away.

The catch is that the stamp only exists in goreleaser-built binaries. `make build`, `go build` and `go install` all print `dev`, which is what the toolchain data in this PR fixes.

I would keep both: read `debug.ReadBuildInfo` first and fall back to the stamped vars. Then drop our explicit `ldflags` so goreleaser's default applies again, and let the Makefile stamp `git describe` for dev builds. That also covers the worktree case from my review, where the toolchain stamps nothing.

PR #233, opened 2026-09-16
fix(policy): find a session's policy under its slug, not the name typed

### LINKED ISSUE #172: A session named with --name cannot carry a policy, and says nothing

marrgko:
## Summary

A per-session policy attachment is stored under the session's slug but was
looked up under the name as typed. The two agree for the `<agent>@<label>`
form, whose labels are slug-clean by construction, and differ exactly when
`--name` was given something to sanitise. So `brig run claude --name Foo`
asked for a row under `Foo`, a key `attach -n` will never write, found
nothing, and booted a sandbox that was meant to be under a policy without
one -- and without a word about it.

A session has one identity, its slug: the sandbox name, the workspace and
the session index are all built from it. The policy lookup was the one place
that used something else. This makes it use the slug, like everything else
about the session.

The other remedy the issue offers -- holding `--name` to the strict rule the
`@` form already uses -- was not taken. It would change the behaviour of
every run that passes `--name`, rather than the one lookup that is wrong,
and `--name` is a spelling already on its way out.

## Related issues

Closes #172

## Changes

- `internal/wrap/config.go`: resolve a run's policies under the session slug
  instead of `rawName`, the same key the sandbox name, the workspace and the
  session index already use.
- `cmd/brig/policy.go`: `brig policy check -n` slugs the session on the way
  in too, so what it reports and what a boot applies are the same thing.
  Previously `check -n Foo` printed "no policy applies" for a session that
  boots under `foo`'s rules.
- `cmd/brig/policy.go`: `check -n` stays lenient about the spelling handed to
  it, unlike `attach -n`. A row filed under a non-slug key is still one
  `brig policy ls` prints, so `check` reports it beside the answer and says
  on stderr that no run reaches it, with the `detach` line to remove it. New
  helper `recordedUnder` reads the raw record for this.
- `docs/policies.md`: the page described the bug as open and told the reader
  to scope per-session policies to the `@` form until it was fixed. It now
  describes what the code does, and names one gap this change leaves alone:
  `attach -n` tests a name against every reserved profile rather than the
  ones reserved for its own agent, so the session `claude@desktop` can hold
  no policy of its own.
- Six new tests, three per side of the change.

## Checklist

- [x] I have updated the affected docs (see `docs/README.md` for the map)

This touches `internal/wrap` but moves neither promise. No host directory and
no credential is resolved differently: the change is one map key in
`policy.Resolve`, and the workspace, the mounts and the credential forwarding
all already keyed on the slug. The bug was that the policy lookup did *not*.

It does fix a silent weakening next door to the second promise. A run that
was meant to carry an egress policy booted carrying none, and said nothing,
which is the worst shape for this: the operator who attached `locked-down`
had every reason to believe it applied. `TestSessionPolicyFollowsTheSlug`
fails without the change.

Two of the three new `internal/wrap` tests are negative, to pin the edges the
fix must not move:

- `TestUnboundSessionCarriesNoPolicy` -- a session with nothing attached still
  resolves to no policy. Slugging the key must not make a run pick up rules
  nobody attached to it.
- `TestARunWithNoSessionIsUnaffected` -- a run with no session at all resolves
  exactly as before.

Checked against a real `hull` (0.1.0-rc27, macOS 26.5.1 arm64), with the
attachment record held in a throwaway `BRIG_POLICY_DIR` so nothing was
written to the real config. A policy attached to session `pr172` -- the only
spelling `attach -n` accepts -- and a run started as `--name Pr172`, the
spelling the bug needs. On `main` that session resolves to `NETWORK shared`
and no policy; on this branch to `NETWORK isolated` and the attached policy,
from an identical record. Booted, its own gateway came up carrying the rules:
`egress default deny (1 allow, 0 deny)`, on a `/30` of its own.
`main` puts the same session on the host's shared gateway instead, which is
the shape of the bug: a run meant to be under a policy, given the shared
network, saying nothing.


marrgko:
@claude-fleet review

docs/policies.md
```diff
@@ -332,25 +332,42 @@ naming the slug it must become. This is stricter than the retired
...
-Use `claude@refactor` when you need a policy bound to one session, and
-attach it with `brig policy attach locked-down claude-code -n refactor`.
-Do not rely on `-n` matching a session opened with `--name` unless the name
-you passed to `--name` was already lowercase, slug-clean text.
+The difference is in what you are allowed to type, not in what you get.
+`ParseRef` applies the same slug rule to the `<agent>@<label>` form:
+`claude@refactor` is accepted and `claude@Refactor` is refused, so a ref
+can only ever name a session whose stored name is `refactor`.
+
+A session opened with `brig run claude --name Refactor` <!-- retired-ok -->
```
asapranidis, line 340: what is the following?
```
 <!-- retired-ok -->
```

marrgko: It is an opt-out marker for `script/check-retired-spellings.sh`, the CI step
that refuses a doc teaching a command spelling brig removes in 0.3. `--name`
on the run line is one of those, and the script only looks inside inline code
and fenced blocks, "because that is where a reader copies from".

This page has to write `brig run claude --name Refactor` in backticks: the bug
it documents is specific to `--name`. The marker is the script's own escape
hatch for a line that names a retired spelling to describe it rather than teach
it (`check-retired-spellings.sh:23-28`, which also asks to keep them rare).

Measured on this file:

    $ ./script/check-retired-spellings.sh docs/policies.md
    exit 0

    $ sed 's|||g' docs/policies.md > /tmp/x.md
    $ ./script/check-retired-spellings.sh /tmp/x.md
    /tmp/x.md:332: retired spelling (--name / --workspace / --memory on the run line -> <agent>@<label> / --home / --mem)
        brig run --name
    /tmp/x.md:340: retired spelling (--name / --workspace / --memory on the run line -> <agent>@<label> / --home / --mem)
        brig run claude --name Refactor
    exit 1

Both markers predate this branch -- `git show main:docs/policies.md` has them
at lines 332 and 341.

pmoust (approved):
LGTM - let's please squash in one commit

marrgko:
Squashed into one commit, and rebased onto current `main`

pmoust:
Great - thank you @marrgko !

PR #308, opened 2026-09-20
wrap: keep sh command argv intact under a login shell

### LINKED ISSUE #307: brig sh mangles a command's arguments by re-parsing them through bash -lc
`brig sh <ref> <command>` does not run the command it was given. It joins the trailing words with spaces and re-parses the result through `bash -lc`, so every argument boundary the caller set is thrown away and re-derived by the shell. Anything with a space, a quote, a glob, a `;` or a `$` comes out different from what was typed, and the failure is silent: the command runs, it just runs something else.

We hit this while reproducing another bug from inside a sandbox. `brig sh ubuntu sh -c 'echo FIRST; echo SECOND'` prints a blank line and `SECOND`, never `FIRST`.

## Why

`shellArgv` in `internal/wrap/run.go:635`:

```go
func shellArgv(command []string) []string {
	if len(command) > 0 {
		return []string{"bash", "-lc", strings.Join(command, " ")}
	}
	return []string{"bash", "-l"}
}
```

`Shell` (the `sh` verb) and `ShellAttached` (the `--json` path) both go through it. The three words `sh`, `-c`, `echo FIRST; echo SECOND` are joined into the single string `sh -c echo FIRST; echo SECOND` and handed to `bash -lc`. bash then splits it back into two commands on the `;`: `sh -c echo FIRST`, where the nested `sh -c` takes `echo` as its script and `FIRST` as `$0` and so runs `echo` with no arguments, and `echo SECOND`. The caller's quoting is gone by the time bash sees the string, so there is nothing left to say the words belonged together.

The join collapses whitespace and re-splits every other metacharacter too:

- `brig sh ubuntu echo 'a   b'` prints `a b`, not `a   b`.
- `brig sh ubuntu touch 'a file'` makes two files, `a` and `file`.
- a glob or a `$VAR` in an argument is expanded by the login shell rather than passed through.

`brig exec` does not have the bug. It hands `tail` straight to `Exec` (`cmd/brig/main.go:674`) with the argument boundaries intact, so `brig exec ubuntu -- sh -c 'echo FIRST; echo SECOND'` prints both lines. The two verbs disagree on what a command is, and only `exec` is right.

## What changes

- `sh` runs the command it was given, with the argument vector the caller passed and no re-splitting, while still running it under a login shell for the environment `sh` is meant to provide. The login-shell-plus-argv shape is `bash -lc 'exec "$@"' bash <command...>`: the words arrive as `$@` and are never re-parsed.
- With no trailing words, `sh` still opens an interactive login shell, unchanged.
- `ShellAttached` (the `--json` path) gets the same treatment, since it shares `shellArgv`.

## Acceptance criteria

- `brig sh <ref> sh -c 'echo FIRST; echo SECOND'` prints both lines.
- `brig sh <ref> echo 'a   b'` prints `a   b` with the spaces intact.
- `brig sh <ref> touch 'a file'` creates one file named `a file`.
- `brig sh <ref>` with no command still opens an interactive login shell.
- A test passes a command with spaces, quotes and a `;` through `sh` and asserts the guest received one unmodified argument vector.

## Out of scope

- Any change to `brig exec` or `brig run`, which pass arguments correctly today.
- The interactive-shell path beyond keeping it working.

## Notes

Found at `431d2c8` on macOS with the hull hvi backend, but the code path is backend-independent. Workaround until fixed: use `brig exec <ref> -- <command>`, or lead the `sh` command with a no-op (`brig sh <ref> sh -c ': ; <command>'`) so the token the nested shell eats is one you do not need.

team-humaki:
`brig sh` joined the command with spaces and handed it to `bash -lc`, so quoting, spaces, `;`, globs, and `$` were re-parsed. The login-shell path is now `bash -lc 'exec "$@"' bash <command...>`. Empty `sh` is still an interactive login shell.

Fixes #307.


internal/wrap/run.go
```diff
@@ -628,13 +628,13 @@ func (c *Config) ExecAttached(set creds.Set, argv []string, tty bool) (int, erro
...
-// one per word. Passed individually, bash takes the first as the script and
-// the rest as $0, $1, ...
+// A command is passed as argv after bash -lc 'exec "$@"' bash, so the login
+// shell supplies the environment without re-parsing the caller's words.
+// Joining them into a -c script would drop quoting and split on spaces, ";",
+// globs, and "$". With no command, sh still opens an interactive login shell.
 func shellArgv(command []string) []string {
 	if len(command) > 0 {
-		return []string{"bash", "-lc", strings.Join(command, " ")}
+		return append([]string{"bash", "-lc", `exec "$@"`, "bash"}, command...)
```
pmoust, line 637: This breaks every single-string invocation. If the caller passes one argv element that is a shell snippet, `command` has length 1 and bash execs it as a filename:

```
$ bash -lc 'exec "$@"' bash 'echo A && echo B'
bash: line 0: exec: echo A && echo B: not found
```

That is the documented shape. `docs/guest-image.md:42` states the contract. `docs/manual-tests/secret-files.md:124,168,193,197` are all single quoted scripts, and each one now fails with `not found`.

One fix that keeps both callers: if `len(command) == 1`, keep `bash -lc <script>`. If `len(command) > 1`, use the new argv form.

pmoust, line 637: `exec` also skips builtins, shell functions, and aliases. `brig sh claude cd /work` now stops with `not found`, and so do `source` and anything the login profile defines instead of exports.

This works against the rationale in the new comment above. The profile is read, but nothing that it defines can run.

pmoust (commented):
On the security question: the old code was not a vulnerability. The words come from the user's own command line, and `cmd/brig/main.go:660,666,731,737` pass `tail` straight through. No `brigd` or daemon path reaches `Shell`. The old `strings.Join` was a quoting bug, not an injection. The new argv form is still the correct hardening for a future programmatic caller, so the direction is right.

The change breaks the single-string form, which the repo documents and uses. Two places outside this diff need a fix in the same PR:

1. `script/smoke.sh:431`, `script/smoke.sh:1913`, and `script/smoke.sh:1432` assert the old command line. `.github/workflows/ci.yml:79` runs this script. All three greps return 1 against the new output. Line 1924 (`-- bash -l$`) still passes.
2. `docs/guest-image.md:42` states that `brig sh <agent> '<command>'` runs `bash -lc`. After this change the code does not do that. `docs/cli.md:91-98` can also use one line that shows how to run a snippet with operators: `brig sh claude bash -lc '...'`.

PR #311, opened 2026-09-21
fix(verify): phrase a refusal as one, rather than decorating a warning

### LINKED ISSUE #167: Verification messages: an empty digest on success, and "Booting it unchecked" on a refusal
## Describe the bug

Three sentences the verification path prints are wrong about what happened. Two are on the paths a stock host takes today.

## 1. The tag path prints a sentence ending in "booting "

`Policy.Image` returns `Result{Policy: p, Outcome: Verified, Image: ref}` at `internal/verify/verify.go:279`, with no `Digest`. `Result.Message()` at `verify.go:210` formats the shared template with `r.Digest`:

```
image ghcr.io/brig-sh/claude-code-stock:latest: signature verified, booting
```

The sentence stops after "booting ". Confirmed by calling `Message()` on exactly the `Result` that `Image()` builds, with `DefaultPolicy()`.

This is the success line on any runtime that cannot pin a digest, which includes hull 0.1.0-rc21, the current release.

Both `Message()` tests for `Verified` pass a non-empty `Digest` (`verify_test.go:368`, `:378`), so nothing catches it.

Fix: give the tag path its own sentence, for example "signature verified, booting the tag".

## 2. A require-mode refusal says it is booting

`internal/wrap/verify.go:100` and `:170` wrap `res.Message()` as `"%s (BRIG_VERIFY=require)"`. The `NoTooling` message at `verify.go:216-217` ends "Booting it unchecked". So a refused boot reads:

```
cannot verify image ghcr.io/brig-sh/claude-code-stock:latest: cosign is not installed (`brew install cosign`). Booting it unchecked (BRIG_VERIFY=require)
```

Nothing booted. The same text nests one level deeper in the boot-asset refusal at `internal/wrap/verify.go:357-358`.

Fix: give `Require` its own refusal sentence rather than decorating one written for the warn path.

## 3. A third-party image under require is refused with reassurance

`verify.go:212-214` produces "That is expected for your own image -- just be sure you trust where it came from (BRIG_VERIFY=require)". The run is being refused, and the message never says what to do about it: set `BRIG_VERIFY=warn`, or point `BRIG_VERIFY_REGISTRY` at your own registry.

## Environment

brig 0.1.0-rc18 (d11ba32), macOS 26.5 arm64, hull 0.1.0-rc21, cosign present, `BRIG_VERIFY` unset so the default `warn` applies.

## Additional context

`brew install cosign` is given unconditionally at `verify.go:216-217` and `doctor.go:310`, on a tool that supports Linux. The adjacent oras message does this correctly by naming a URL (`internal/runtime/bootfetch.go:53-55`).

Also at `internal/wrap/verify.go:145-149`, the comment says a `NotOurs` result still pins the digest. `Policy.Verify` returns at `verify.go:311-312` with `Digest` unset, so nothing is pinned. The README and docs/security.md both describe the real behavior; only the comment is wrong.

marrgko:
## Summary

Three sentences the verification path printed were wrong about what had just
happened, and two of them are on the paths a stock host takes.

The success line on the tag path ended in a space: `Policy.Image` resolves
nothing, so `Digest` is empty there, and the shared template printed it
regardless. Under `BRIG_VERIFY=require`, the two "nothing could be checked"
rows refused the boot and then described one, because the callers appended
the mode to a sentence written for the warn path -- so a refused boot read
`Booting it unchecked (BRIG_VERIFY=require)`. A third party's image was
refused with reassurance and named nothing to do about it.

The approach is to stop sharing one sentence between two situations. `Message`
stays exactly as it was and keeps the warn path. A new `Result.Refusal` gives
the two rows their own wording under `require`: it says it refused, and names
the way out. It is one place rather than the three call sites that would
otherwise drift.

A fourth instance turned up while fixing the first: the replaced-policy
spelling of the success line formats the same field and had the same gap.

## Related issues

Closes #167

## Changes

- `internal/verify/verify.go`: `Verified` names the tag when no digest was
  resolved, in both spellings of the line. New `Result.Refusal`, the wording
  `NotOurs` and `NoTooling` take under `require`. It splits on a replaced
  policy: there the registry that put an image in that row is the user's own,
  so the line names neither brig-sh nor the settings they already chose, and
  names the registry alone, which is the only one the row turns on.
- `internal/wrap/verify.go`: the three `require` refusals for those rows read
  `Refusal` instead of decorating `Message`.
- `internal/wrap/verify.go`: the boot-asset branch names its cause rather than
  nesting the image message, whose subject is the image and not the kernel.
  Its warning still says the kernel boots.
- `docs/troubleshooting.md`: the page told the reader to read the exit code
  rather than the last clause of the message. It now quotes what the code says.
- Five new or strengthened tests, including a guard that no refusal says
  something booted.

## Checklist

- [x] I have updated the affected docs

Exercised on macOS 26.5.1 arm64 with hull 0.1.0-rc27 and cosign v2.6.5. Both
refusals were reproduced against the real runtime and both exit `5`:

    $ BRIG_VERIFY=require BRIG_COSIGN_BIN=/nonexistent/cosign brig run claude@t1
    brig: refusing to boot image ghcr.io/brig-sh/claude-code-stock:latest: cosign
    is not installed (`brew install cosign`), so nothing could be checked
    (BRIG_VERIFY=require). Install cosign, or set BRIG_VERIFY=warn to boot it
    unchecked

    $ BRIG_VERIFY=require BRIG_IMAGE=docker.io/library/ubuntu:24.04 brig run claude@t2
    brig: refusing to boot image docker.io/library/ubuntu:24.04: it is not
    published by brig-sh, so there is no signature of ours to check
    (BRIG_VERIFY=require). Set BRIG_VERIFY=warn to boot your own image, or point
    BRIG_VERIFY_REGISTRY, BRIG_VERIFY_IDENTITY and BRIG_VERIFY_ISSUER at the
    publisher you trust

The truncated success line cannot be reproduced on a current host: it belongs
to the tag path, which a runtime takes only when it cannot boot by digest, and
hull has pinned digests since 0.1.0-rc23. `TestVerifiedWithoutADigestNamesTheTag`
covers it instead.

Neither promise changes. The change is confined to the wording of three
refusals and one alert; nothing about what reaches the guest, or which
credentials are resolved, is touched.


marrgko:
@claude-fleet review

ananos (commented):
Thanks @marrgko! The refusals read much better now. A few things I noticed:

- When cosign cannot be found, the refusal says `brew install cosign`, even if cosign is installed and `BRIG_COSIGN_BIN` points at a wrong path. Naming `BRIG_COSIGN_BIN` in that case would send people to the real problem. The same sentence now appears in three places (`verify.go`, `wrap/verify.go`, `doctor.go`), so a shared helper would keep them in step.
- The boot-asset refusal under `require` (for example when the registry cannot be reached) does not say how to get past it, such as `BRIG_VERIFY=warn`. The image refusals now do.
- That message says "the registry could not be reached" for every unresolved case, including a tag that does not exist or one that needs auth.
- With a custom `BRIG_VERIFY_REGISTRY`, the refusal suggests widening it to cover the image. If the identity still points at another publisher, the image then fails the signature check instead, which is harder to get past. Maybe suggest only `BRIG_VERIFY=warn` there?
- `Refusal()` falls back to the warn-style `Message()` for the other outcomes. Nothing reaches that today, but a small test on the fallback would catch it later.

